PR #27, opened 2020-03-31
Update The_Cartridge_Header.md

JL2210:
The checksum is stored in little-endian form, which has the least significant byte first. This was incorrect in the originals, but I have checked this on two games by Nintendo with verified good dumps:

* Super Mario Land v1.1 Japan/US/Europe
* Super Mario Land 2 v1.0 USA/Europe

### DIFF
--- a/content/The_Cartridge_Header.md
+++ b/content/The_Cartridge_Header.md
@@ -244,7 +244,7 @@ entry. The GAME WON'T WORK if this checksum is incorrect.
 
 ### 014E-014F - Global Checksum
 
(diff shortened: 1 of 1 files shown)


nitro2k01:
But that's wrong... The most significant checksum byte _is_ stored in the lower address, ie big-endian, including for the ROMs you listed. Could you please specify in more detail how you came to the conclusion that it's stored as little-endian?

JL2210:
You're right. I was generating the checksum for those files and then printing it (which automatically prints the most significant digit first), and that matched the hexdump from the file.

PR #32, opened 2020-05-13
Add Pixel FIFO information

corybsa:
I documented SameBoy's implementation of the Pixel FIFO. Somebody more knowledgeable than me needs to read over this because I'm still not confident I understand the Pixel FIFO. I tried not to include anything specific to SameBoy, but I'm flying blind on this one since there are no other documents on this topic to reference.


content/pixel_fifo.md
```diff
@@ -0,0 +1,252 @@
...
+
+![](imgs/game-boy-lcd-refresh-diagram.png "imgs/game-boy-lcd-refresh-diagram.png")
+
+## FIFO info
+FIFO stands for First In, First Out. The first pixel to be pushed to the
+FIFO is the first pixel to be popped off. In theory that sounds great,
+in practice there are a lot of intricacies.
+
+There are two pixel FIFOs. One for background pixels and one for OAM
+(sprite) pixels. hese two FIFOs are not shared. They are two independent
```
JL2210, line 17: **T**hese ... They are ~~two~~ independent

avivace:
Thank you @corybsa !
Here's some feedback from @LIJI32 :

- `Palette: on CGB a value between 0 and 7 and on DMG a value between 1 and 0`
On the DMG it only applies to sprites

- `Sprite Priority: on CGB this is the OAM index for the sprite and on DMG this is 0`
On the DMG it simply doesn't exist. This refers to SameBoy's implementation and I don't know if the HW is implemented the same way

`The pixel fetcher has 8 steps.
 The order of the steps are as follows`
 this is also SameBoy specific

In reality it has 5 steps. The first 4 take 2 T-cycles each. The last is attempted every T-cycle until it succeeds

The second push step doesn't exist in reality, it's for save state compatibility

I'll go on later

content/pixel_fifo.md
```diff
@@ -0,0 +1,252 @@
...
+then the value for the tile is read as $FF.
+
+CGB can access both tile index and the attributes in the same clock
+cycle.
+
+### Get Tile Data Low:
+Check LCDC.4 for which tilemap to use. At this step CGB also needs to
+check which VRAM bank to use and check if the tile is flipped vertically.
+Once the tilemap, VRAM and vertical flip is calculated the tile data
+is retrieved from VRAM. However, if the PPU's access to VRAM is
```
leefogg, line 78: Is this possible during Mode 3?

content/pixel_fifo.md
```diff
@@ -0,0 +1,252 @@
...
+NOTE: These conditions are checked only when entering STOP mode and the
+PPU's access to CGB palettes is always restored upon leaving STOP mode.
+
+### Sprite Fetch Abortion
+Sprite fetching may be aborted if LCDC.1 is disabled while the PPU is
+fetching an object from OAM. This abortion lengthens mode 3 by the amount
+of cycles the previous instruction took plus the residual cycles left for
+the PPU to process. When OAM fetching is aborted a pixel is [rendered](#pixel-rendering),
+the [fetcher](#fifo-pixel-fetcher) is advanced one step. This advancement
+lengthens mode 3 by 1 cycle if the X coordinate of the current scanline
```
leefogg, line 250: scanlines don't have x cooridinates, do you mean the current dot?

content/pixel_fifo.md
```diff
@@ -0,0 +1,252 @@
...
+time. The two FIFOs are manipulated only during mode 3 (pixel transfer)
+and they are both cleared at the beginning of mode 3.
+
+Each pixel in the FIFO has four properties:
+- Color: a value between 0 and 3
+- Palette: on CGB a value between 0 and 7 and on DMG a value between 1 and 0
+- Sprite Priority: on CGB this is the OAM index for the sprite and on DMG this is 0
+- Background Priority: holds the value of the [OBJ-to-BG Priority](#vram-sprite-attribute-table-oam) bit
+
+## FIFO Pixel Fetcher
```
leefogg, line 30: How about starting with "Each Pixel FIFO contains a Pixel Fetcher, this fetcher runs in parallel to the FIFO and is responsible for fetching the next tile's data for the FIFO to draw."

content/pixel_fifo.md
```diff
@@ -0,0 +1,252 @@
...
+conditions where either a background pixel or a sprite pixel will have
+display priority.
+
+If there are pixels in the background and OAM FIFOs then a pixel is
+popped off each. If the OAM pixel is not transparent and LCDC.1 is
+enabled then the OAM pixel's background priority property is used if it's
+the same or higher priority as the background pixel's background priority.
+
+Pixels won't be pushed to the LCD if there is nothing in the background
+FIFO or the X coordinate for the current scanline is greater than 159.
```
leefogg, line 205: "or the current pixel is pixel 160 or greater"

content/pixel_fifo.md
```diff
@@ -0,0 +1,252 @@
...
+## FIFO info
+FIFO stands for First In, First Out. The first pixel to be pushed to the
+FIFO is the first pixel to be popped off. In theory that sounds great,
+in practice there are a lot of intricacies.
+
+There are two pixel FIFOs. One for background pixels and one for OAM
+(sprite) pixels. These two FIFOs are not shared. They are independent
+of each other. The two FIFOs are mixed only when popping items. Sprites
+take priority unless they're transparent (color 0) which will be
+explained in detail later. The FIFOs can each hold up to 16 pixels at a
```
leefogg, line 20: "The FIFOs can each hold up to 16 pixels at a time" This doesn't seem relevant here. As this is quite fundamental I'd recommend having a paragraph before this explaining it. e.g "Each FIFO can hold up to 16 pixels. The FIFO and Pixel Fetcher work together to ensure that the FIFO always contains at least 8 pixels at any given time, as 8 pixels are required for the Pixel Rendering operation to take place."

content/pixel_fifo.md
```diff
@@ -0,0 +1,252 @@
...
+time. The two FIFOs are manipulated only during mode 3 (pixel transfer)
+and they are both cleared at the beginning of mode 3.
```
leefogg, line 22: This is confusing. If they are cleared at the start of mode 3 that implies they are empty so Pixel Rendering cannot take place. Surely its cleared and the fetcher starts 8/16 cycles before model 3 so that there are pixels ready at the start of model 3?

content/pixel_fifo.md
```diff
@@ -0,0 +1,252 @@
...
+Before we get started, all references to a cycle are meant as T-cycles
+(4.19 MHz) and cycle counts are doubled on CGB in double speed mode. Also
+when stating that a certain action "lengthens mode 3" this means that
+mode 0 (hblank) is shortened. The extra time has to be made up somewhere
+and the lengthening of mode 3 eats into the mode that comes afterwards,
+which happens to be mode 0 as shown in the following diagram.
+
+![](imgs/game-boy-lcd-refresh-diagram.png "imgs/game-boy-lcd-refresh-diagram.png")
+
+## FIFO info
```
leefogg, line 11: This feels out of place. Instead of this could the info above be made into a yellow area as seen on the "Accessing VRAM and OAM" section from Video Display.md?

avivace:
Thanks @leefogg ! 
@corybsa do you think you can work through these suggestions and corrections?

corybsa:
Yep, I've been trying to find a couple hours to do this. I can probably work through them after midterms next Thursday.

avivace:
@corybsa Great! Let me know if you need anything

leefogg:
Any clue on when this will be updated?

corybsa:
> Any clue on when this will be updated?

Sorry guys, I didn't forget, just been super busy. I just submitted my finals a few minutes ago. Going to work on this right now.

content/pixel_fifo.md
corybsa: I think this might be implementation specific to SameBoy. There is code for clearing both FIFOs, filling them with "junk" and resetting the fetcher at the beginning of mode 3. It might be safe to assume this does not represent the behavior of the real hardware. Reading the code again, I think this might even only be executed at the beginning of each frame.

content/pixel_fifo.md
avivace: I think this is the intention? 
```
::: warning BOXTITLE
```
produces a yellow warning box

avivace:
Thanks @corybsa ! I really appreciate you taking the time to write down all of this. And thanks @leefogg for the precious feedback.

I merged the PR so we can have more exposure / feedback on this, feel free to keep the feedback and the comments coming

Issue #28, opened 2020-04-05
Adding current version to exported files?


nitro2k01:
Would it be possible to add some kind of version stamp to exported files? I don't know what would be easiest or best, but for example date+commit hash or an auto-incrementing version number.

avivace:
I second this.
Maybe a commit hash with a date?

avivace:
18fc351fa11fe2b253f74ad91679771e8ed49feb includes a commit hash and a link to that commit.
In the future, we can exploit the git tag functionality to produce tagged and name releases of the document.

avivace:
> [18fc351](https://github.com/gbdev/pandocs/commit/18fc351fa11fe2b253f74ad91679771e8ed49feb) includes a commit hash and a link to that commit.
> In the future, we can exploit the git tag functionality to produce tagged and named releases of the document.

@nitro2k01 what do you think about this?

PR #36, opened 2020-05-23
Use vuepress-plugin-git-log to embed commit metadata.

bfabio:
This simplifies the workflow (and maybe it can also be the solution for #28: `vuepress-plugin-git-log` injects each page with the [`$page.git.updated` and `$page.git.commits` variables](https://github.com/vuepress/vuepress-plugin-git-log#api)).

### DIFF
--- a/content/About.md
+++ b/content/About.md
@@ -2,3 +2,9 @@ This is a new, experimental version of Pan Docs, mantained in the Markdown forma
 
 To learn more about the history and the mission of the project, check the [README](https://github.com/gbdev/pandocs#history).
 
+This document version was produced from git commit
+<a :href="`https://github.com/gbdev/pandocs/tree/${$page.git.HEAD.hash}`"
+    target="_blank">
+    {{ $page.git.HEAD.shorthash }}<OutboundLink />
+</a>
+({{ $page.git.HEAD.timestamp }})
--- a/render/.vuepress/config.js
+++ b/render/.vuepress/config.js
@@ -2,6 +2,25 @@ module.exports = {
   "title": "Pan Docs",
   // We're serving this on gbdev.github.io/pandocs
   base: "/pandocs/",
+  plugins: [
+    ['git-log', {
+      onlyFirstAndLastCommit: true,
+      extendGitLog: (git) => {
+        const spawn = require("cross-spawn").sync
+
+        const res = spawn("git", ["log", "-1", "--format=%H/%h/%ci", "HEAD"])
+        if (res.stderr.toString().trim()) return
+
+        const [hash, shorthash, timestamp ] = res.stdout.toString().trim().split("/")
+
+        git.HEAD = {
+          hash,
+          shorthash,
+          timestamp,
+        }
+      },
+    }],
+  ],
   themeConfig: {
     navbar: false
     /*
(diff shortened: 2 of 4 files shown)


avivace:
Thanks a lot for the contribution!

Unfortunately, this is a solution I tried and then excluded because it's tied to Vuepress.
One of the principles (see the RFC) of this project is keeping Pandocs in a agnostic format (markdown) and use tools able to consume this format to produce the final documents. Currently we only have "2 tools":

- Vuepress produces an HTML version,
- the `Pandoc` toolchain produces a PDF build

`vuepress-plugin-git-log` would mean to adopt an additional solution to add the same data but on the other tool (because pandoc takes the plain MD file), thus duplicating the effort.

Currently, the metadata gets added directly in the MD, so it's available to whatever tool makes use of the markdown file.

However, this could be useful when/if we decide to render every single page also in a different HTML page (so an additional version to the "merged" one).

Let me know if I explained myself, @bfabio

bfabio:
@avivace Thanks for the project!

> One of the principles (see the RFC) of this project is keeping Pandocs in a agnostic format (markdown) and use tools able to consume this format to produce the final documents.

That makes a lot of sense.

My reasoning behind this PR was to use just one environment (vuepress) and absorb in it the features handled manually (merge.sh and markdown-pp). After all, a documentation system like vuepress should cover all our bases.

I felt we were kinda fighting vuepress and in the long run I think it can end up biting us (especially if we can't use its advanced features), negating all the goodies we can get from it and forcing us into a limbo of "almost standard vuepress, but not quite".

Also, merge.sh is kinda awkward and does that unexpected trick with `About.md` which makes me uncomfortable.

So, I think ideally a single tool supporting what we need out of the box from the get go would be great, but that tool wouldn't be vuepress.

Have you considered [readthedocs](https://github.com/readthedocs/readthedocs.org)?
It's powerful, themeable, extensible (templating the index shouldn't be a problem), supports reStructuredText (which can be a even finer format for documentation) and a ton of conversion formats via [Sphinx](https://www.sphinx-doc.org/en/master/).
I think it can check all our boxes.

avivace:
@bfabio Sorry for the late reply, busy time.

> Also, merge.sh is kinda awkward and does that unexpected trick with About.md which makes me uncomfortable.

What are you referring to? The way I'm injecting the revision data?

I think everything you mentioned makes a lot of sense. My main concern here is burying the project (if we change the infrastructure) under more content-unrelated work that could slow things down now that we are *able* to accept contributions and actually push the research forward (one of the main priorities).

I don't actually have time to take a look at the alternative (but I know readthedocs and I really like it), although it seems promising, especially considering that the only requisite I want from the frontend build is being able to use our CSS.

Feel free to work on proof of concepts and ask for feedback here (or on our [Discord server](https://discord.gg/gpBxq85) where you can get more instantenous feedback).

bfabio:
> What are you referring to? The way I'm injecting the revision data?

Yes, in particular changing a source file, even for a brief amount of time, was unexpected to me. It's even more visible if markdown-pp fails for some reason and you are left wondering who changed it. 

It's also awkward if we wanted to Dockerize the whole thing, because it forces us to keep track of the file UID to restore it after the change.

It would be cleaner to just add
`
This document version was produced from git commit ["$shortcommithash"](https://github.com/gbdev/pandocs/tree/"$commithash") ("$timestamp")."
`
to `index.mdpp` and commit the change, as  if`$shortcommithash` and the others were magic variables the doc system provides. WDYT?

> Feel free to work on proof of concepts and ask for feedback here (or on our Discord server where you can get more instantaneous feedback).

I'll be happy to give it a go, I just joined the Discord server :)

dcasella (approved):

PR #40, opened 2020-06-08
Correcting the mode select and adding information about MBC1 multi-ga…

TheThief:
…me carts

### DIFF
--- a/content/MBC1.md
+++ b/content/MBC1.md
@@ -8,16 +8,18 @@ types of MBCs.
 Note that the memory in range 0000-7FFF is used for both reading from
 ROM, and for writing to the MBCs Control Registers.
 
-### 0000-3FFF - ROM Bank 00 (Read Only)
+### 0000-3FFF - ROM Bank 00/20/40/60 (Read Only)
 
-This area always contains the first 16KBytes of the cartridge ROM.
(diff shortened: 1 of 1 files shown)


mattrberry:
I would clarify that there are 2 banking registers that are being updated independently. That wasn't immediately obvious to me. I'd also specify that the first register defaults to 1. Still, this was useful to me in implementing proper MBC1 functionality!

avivace:
@TheThief do you think you can incorporate @mattrberry suggestion somewhere?

TheThief:
I've had a go at clarifying a few things

avivace:
Looks pretty good to me!

mattrberry:
@TheThief @avivace I know this is already in develop, but I still think it'd be useful to specify that the rom banking number defaults to 1. I feel like the standard assumption would be that they'd both default to 0, but the rom banking number breaks that assumption. That caused a bug in _my_ emulator when I was implementing MBC1, and the only way I figured it out was by asking @TheThief on discord.

All of the other changes look great!

TheThief:
That is actually mentioned now!

mattrberry:
Ah, I read past that! Looks good!

avivace:
:tada:

PR #41, opened 2020-06-29
Remove extraneous '^'; introduced by wiki to MD conversion?

RetroEdit:


avivace:
Thanks @RetroEdit ! Is this happening somewhere else?

RetroEdit:
I did a grep and didn't find any '^'; there could be similar artifacts from tags in the wiki, potentially (haven't downloaded and grepped wiki yet). Essentially the problem appears to come from `2<sup>22</sup>` being converted to `2^22^`.

What converter was used to turn the wikitext into Markdown?

avivace:
Oh okay, makes sense.
The export from mediawiki and conversion to pandoc is handled by the scripts here: https://github.com/gbdev/pandocs/tree/develop/mediawiki-exporter

The XML -> Markdown conversion is done via pandoc

RetroEdit:
Ah, [pandoc](https://pandoc.org/), not this Pan Docs. I was a bit confused until I looked at the Python code for processing the MediaWiki XML export.

avivace:
@RetroEdit markdown-pp is only doing the "merge", the actual rendering (for the web version) is handled by vuepress

PR #47, opened 2020-07-31
Make it clear that OPRI is not actually a register that is usable from a normal cartridge game/rom

daid:
The current description of OPRI does not make it clear that changing it outside of the boot rom has no effect.

While maybe there will be a way found in the future to effect the mode, as long as we don't know, it's better not to word this so as if this register actually does something.

### DIFF
--- a/content/CGB_Registers.md
+++ b/content/CGB_Registers.md
@@ -94,8 +94,11 @@ does not include an infra-red port.
 This register serves as a flag for which object priority mode to use. While
 the DMG prioritizes objects by x-coordinate, the CGB prioritizes them by
 location in OAM. This flag is set by the CGB bios after checking the game's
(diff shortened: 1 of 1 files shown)


avivace:
Thanks @daid ! I've just put the warning in a "to be verified" box.

LIJI32:
As mentioned in another commit, writing to OPRI has an effect if a PGB value (0xX8, 0xXC) is written to KEY0 but `STOP` hasn't been executed yet, and the write takes effect instantly. It does not have an effect, at least not an instant effect, if written to during CGB or DMG mode after the boot ROM has been unmapped. It is not known if triggering a PSM NMI, which remaps the boot ROM, has an effect on this register's behavior.

avivace:
Thanks @LIJI32 . I'll add this information to the box

PR #52, opened 2020-09-05
Correct MBC2 information to reflect gbctr

mattrberry:
The changes here align the Pan Docs with gekkio's gbctr docs. Gekkio has tested this behavior with the mooneye-gb MBC2 tests. All of gekkio's testing is done on hardware using real MBCs.


avivace:
Great, thanks @mattrberry !

PR #62, opened 2020-09-30
Fix many malformed local links

RetroEdit:
In the process, I removed a few links specific to the original wiki:

BGB, which originally linked to https://gbdev.gg8.se/wiki/articles/BGB

Mobile Adapter, Popslide, VBlank, and VBlank handler, which were all red links on the wiki.

I opted to retain gameboy-camera but transform it to a proper local link, even though it isn't currently rendered, because Gameboy_Camera.md exists and could be added to the document later.

The links in Gameboy_Camera.md and Pan_Docs.md were left untouched because they are not currently rendered.


avivace:
LGTM. Thanks!

PR #63, opened 2020-10-01
Updated Spelling And Language

wawesomeNOGUI:
I fixed some spelling errors and added words to make the text easier to understand.

### DIFF
--- a/content/CGB_Registers.md
+++ b/content/CGB_Registers.md
@@ -48,9 +48,9 @@ pseudo code would be:
 ```
 
 The CGB is operating in Normal Speed Mode when it is turned on. Note
-that using the Double Speed Mode increases the power consumption, it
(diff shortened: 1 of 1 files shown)


content/CGB_Registers.md
```diff
@@ -48,9 +48,9 @@ pseudo code would be:
 ```
 
 The CGB is operating in Normal Speed Mode when it is turned on. Note
-that using the Double Speed Mode increases the power consumption, it
+that using the Double Speed Mode increases the power consumption, therefore it
```
aaaaaa123456789, line 51: This is still a comma splice; the comma must be changed to a semicolon. The connector in that case becomes optional; however, if one is used, it must be followed by a comma (like "however" in this sentence).

Issue #35, opened 2020-05-20
Game Boy Printer conversion ignored rowspan


pinobatch:
Three fields in the Game Boy Printer packet format are two bytes in length: "Magic bytes", "Length of data", and "Checksum". The table in the old version used `colspan="2"` to expand "Description" and "Size" to span two columns of "GB to Printer" and "Printer to GB". The conversion was transposed to look better on narrow screens of handheld computers: columns became rows, and rows became columns. It could have used `rowspan="2"` to expand them, or it could have taken both of the byte values in one cell. But instead, they are misaligned, with the second byte of each field pushing the following fields down a row.

- "Magic bytes" should be "$88, $33" and "Command" should be "See below", but the conversion has "Command" set to "$33" and "Compression flag" set to "See below".
- "Length of data" should be "LSB, MSB", but the conversion has "Command-specific data" set to "LSB" and "Checksum" set to "MSB".
- "Checksum" should be "LSB, MSB", but the conversion has "Status" set to "LSB" and the second half of the field cut off the end of the table.

Issue #9, opened 2020-02-07
Use of single or double quotes


avivace:
E.g. in CGB Registers

PR #72, opened 2020-10-17
Replace ' with " and fix GB Printer Table

Hpmason:
Replaced instances of single quotes with double quotes (issue #9 ). 
Also fixed the GB Printer table (issue #35 ). I'm pretty sure I've formatted the table correctly now according to the issue. I added "See below" for the Status since it goes more in-depth below, which wasn't mentioned in the issue.


Hpmason:
This resolves #35 and resolves #9

avivace:
That's great, thank you @Hpmason ! :tada:

PR #77, opened 2020-11-10
clarify the effect of double speed mode and sgb1 v 2 for the Timer

CasualPokePlayer:


ISSOtm:
Nice! While we're at this, could you check if there are other clock measurements that would benefit from a similar update?

CasualPokePlayer:
Looking at other measurements, the most that could be changed is clarifying SGB1 timings for the video and sound timings at their respective points.

ISSOtm (approved):

avivace (approved):

PR #79, opened 2020-11-13
A trivial addition on which bits enable/disable window/bg of LCDC.

adembudak:


content/Video_Display.md
```diff
@@ -566,8 +568,7 @@ coordinates of the top left corner of a window are WX-7,WY. The tiles
 for the window are stored in the Tile Data Table. Both the Background
 and the window share the same Tile Data Table.
 
-Both background and window can be disabled or enabled separately via
-bits in the LCDC register.
+Window can be disabled(0)/enabled(1) with 5th bits of LCDC register.
```
avivace, line 571: I don't like the  *disabled(0)/enabled(1)*

content/Video_Display.md
```diff
@@ -556,6 +556,8 @@ to select the origin of the visible 160x144 pixel area within the total
 256x256 pixel background map. Background wraps around the screen (i.e.
 when part of it goes off the screen, it appears on the opposite side.)
 
+Background can be disabled(0)/enabled(1) with 0th bit of LCDC register.
```
avivace, line 559: I don't think "with 0th" sounds good. Maybe *with the bit 0*?

avivace (changes requested):

content/Video_Display.md
adembudak: I'll look for better alternatives, maybe *LCDC.bit0* would be better.

content/Video_Display.md
aaaaaa123456789: The grammatically correct form would be "The window can be disabled or enabled with bit 5 of the LCDC register." It's not the 5th bit (because we count from 0).

content/Video_Display.md
aaaaaa123456789: Just like above, the grammatically correct form would be "The background can be disabled or enabled with bit 0 of the LCDC register."

PR #98, opened 2020-12-25
Polish header info

ISSOtm:
- Fix hex formatting
- Improve pseudocode formatting

### DIFF
--- a/content/The_Cartridge_Header.md
+++ b/content/The_Cartridge_Header.md
@@ -1,12 +1,12 @@
-An internal information area is located at 0100-014F in each cartridge.
+An internal information area is located at $0100-014F in each cartridge.
 It contains the following values:
 
 ### 0100-0103 - Entry Point
 
 After displaying the Nintendo Logo, the built-in boot procedure jumps to
-this address (100h), which should then jump to the actual main program
-in the cartridge. Usually this 4 byte area contains a NOP instruction,
-followed by a JP 0150h instruction. But not always.
+this address ($0100), which should then jump to the actual main program
+in the cartridge. Usually this 4 byte area contains a `nop` instruction,
+followed by a `jp $0150` instruction. But not always.
 
 ### 0104-0133 - Nintendo Logo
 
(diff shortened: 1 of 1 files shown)


content/The_Cartridge_Header.md
```diff
@@ -214,33 +213,36 @@ Specifies if this version of the game is supposed to be sold in Japan,
...
-x=0:FOR i=0134h TO 014Ch:x=x-MEM[i]-1:NEXT
+x = 0
+i = $0134
+while i <= $014C
+	x = x - [i] - 1
 ```
 
-The lower 8 bits of the result must be the same than the value in this
-entry. The GAME WON'T WORK if this checksum is incorrect.
+If the byte at $014D does not match the low 8 bits of `x`, the boot ROM will lock up,
```
avivace, line 244:
```suggestion
If the byte at $014D does not match the lower 8 bits of `x`, the boot ROM will lock up,
```

content/The_Cartridge_Header.md
```diff
@@ -214,33 +213,36 @@ Specifies if this version of the game is supposed to be sold in Japan,
...
+and the cartridge program **will not be run**.
```
avivace, line 245:
```suggestion
and the cartridge program **won't run**.
```

content/The_Cartridge_Header.md
```diff
@@ -43,21 +43,20 @@ CGB cartridges the upper bit is used to enable CGB functions. This is
 required, otherwise the CGB switches itself into Non-CGB-Mode. Typical
 values are:
 ```
- 80h - Game supports CGB functions, but works on old gameboys also.
- C0h - Game works on CGB only (physically the same as 80h).
+ 80h - Game supports CGB functions, but works on old Game Boys also.
```
avivace, line 46:
```suggestion
 80h - Game supports CGB functions, but also works on old Game Boys.
```

avivace (changes requested):
Looks good to me. Suggested a couple of changes

PR #105, opened 2021-01-24
Update Timer_Obscure_Behaviour.md

wawesomeNOGUI:
Just a spelling fix. Also, how is the timer affected when writing to DIV? Is it reset too?

### DIFF
--- a/content/Timer_Obscure_Behaviour.md
+++ b/content/Timer_Obscure_Behaviour.md
@@ -14,7 +14,7 @@ Notice how the values that are connected to the inputs of the
 multiplexer are the values of those bits, not the carry of those bits.
 This is the reason of a few things:
 
-- When writing to DIV, the whole counter is reseted, so the timer is
+- When writing to DIV, the whole counter is reset, so the timer is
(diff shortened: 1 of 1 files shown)


CasualPokePlayer:
For all intents and purposes, rTIMA is essentially a separate counter which ticks according to a set frequency described in rTAC. This frequency is tied to rDIV and its internal bits. All writing rDIV does is reset rDIV and its internal bits; it does not reset rTIMA. But since rDIV and its internal bits are used to source the frequency of rTIMA's ticks, it will be affected by the rDIV write.

ISSOtm:
Well, it *may* be affected: `TIMA` is updated when a falling edge is detected on a certain bit of `DIV`; which bit this is is selected by `TAC`. Since writing to `DIV` resets it, it may cause the observed bit to be reset, thus incrementing `TIMA`.

ISSOtm (approved):

wawesomeNOGUI:
Cool! Thanks for the information! @CasualPokePlayer and @ISSOtm

Issue #66, opened 2020-10-01
Use test ROMs as citations to stated facts


avivace:
Credit @LIJI32

ISSOtm:
Since I'm about to make a PR that details additional behavior, what style do we want? What should the link point to, the specific test ROM's source, the ROM collection? What if multiple ROMs need to be referenced?

avivace:
#74

PR #106, opened 2021-01-26
Document address wrapping behaviour of MBC2 RAM, and note questionable source of $52-$54 ROM size tags

TheThief:
Also note that 2KB RAM size tag is erroneously used by a lot of "PD" ROMs (no cart was ever made with this size).

### DIFF
--- a/content/MBC2.md
+++ b/content/MBC2.md
@@ -16,15 +16,22 @@ The MBC2 doesn't support external RAM, instead it includes 512x4 bits
 of built-in RAM (in the MBC2 chip itself). It still requires an external
 battery to save data during power-off though. As the data consists of
 4bit values, only the lower 4 bits of the "bytes" in this memory area
-are used.
+are used. The upper 4 bits of each byte are undefined and should not be
+relied upon.
+
+### A200-BFFF - 15 "echoes" of A000-A1FF
+
+Only the bottom 9 bits of the address are used to index into the internal
+RAM, so RAM access repeats.
 
 ## Control Registers
 
 ### 0000-3FFF - RAM Enable and ROM Bank Number (Write Only)
 
 This address range is responsible for both enabling/disabling the RAM
-and for controlling the ROM bank number. Bit 8 of the address determines
-whether to control the RAM-enable flag or the ROM bank number.
+and for controlling the ROM bank number. Bit 8 of the address (the least
+significant bit of the upper address byte) determines whether to control
+the RAM enable flag or the ROM bank number.
 
 #### When Bit 8 is Clear
 
--- a/content/MBC2.md
+++ b/content/MBC2.md
@@ -22,7 +22,9 @@ relied upon.
 ### A200-BFFF - 15 "echoes" of A000-A1FF
 
 Only the bottom 9 bits of the address are used to index into the internal
-RAM, so RAM access repeats.
+RAM, so RAM access repeats. As with the A000-A1FF region, only the lower
+4 bits of the "bytes" are used, and the upper 4 bits of each byte are
+undefined and should not be relied upon.
 
 ## Control Registers
 
--- a/content/The_Cartridge_Header.md
+++ b/content/The_Cartridge_Header.md
@@ -192,7 +192,9 @@ Specifies the ROM Size of the cartridge. Typically calculated as "N such that 32
 | $53 | 1.2 MByte | 80 *
 | $54 | 1.5 MByte | 96 *
 
-\* Only listed in unofficial docs
+\* Only listed in unofficial docs. No cartridges or ROM files using these sizes are known.
+As the other ROM sizes are all powers of 2, these are likely inaccurate. The source for these
+values is unknown.
 
 ### 0149 - RAM Size
 
@@ -200,15 +202,23 @@ Specifies the size of the external RAM in the cartridge (if any).
 
 | Code | Size   | Comment
 |------|--------|---------
-| $00  | 0      | No RAM
-| $01  | -      | Unused (but unofficial docs list it as 2 KB)
+| $00  | 0      | No RAM *
+| $01  | -      | Unused **
 | $02  | 8 KB   | 1 bank
 | $03  | 32 KB  | 4 banks of 8 KB each
 | $04  | 128 KB | 16 banks of 8 KB each
 | $05  | 64 KB  | 8 banks of 8 KB each
 
-When using a MBC2 chip, $00 must be specified in this entry, even though
+\* When using a MBC2 chip, $00 must be specified as the RAM Size, even though
 the MBC2 includes a built-in RAM of 512 x 4 bits.
+
+\** Listed in various unofficial docs as 2KB. However, a 2KB RAM chip was never used in a cartridge.
+The source for this value is unknown.
+
+Various "PD" ROMs ("Public Domain" homebrew ROMs generally tagged "(PD)"
+in the filename) are known to use the $01 RAM Size tag, but this is believed
+to have been a mistake with early homebrew tools and the PD ROMs often don't use
+cartridge RAM at all.
 
 ### 014A - Destination Code
 


content/The_Cartridge_Header.md
```diff
@@ -187,6 +187,11 @@ Specifies the ROM Size of the cartridge. Typically calculated as "N such that 32
 |`$06`|   2 MByte |128 banks|
 |`$07`|   4 MByte |256 banks|
 |`$08`|   8 MByte |512 banks|
+
+The following rom types are listed on older pandocs but the source is unknown, and no carts using them are known. You probably don't need to support these.
```
ISSOtm, line 191: Aren't they used by MMM01?

content/The_Cartridge_Header.md
```diff
@@ -208,6 +213,8 @@ Specifies the size of the external RAM in the cartridge (if any).
 When using a MBC2 chip $00 must be specified in this entry, even though
 the MBC2 includes a built-in RAM of 512 x 4 bits.
 
+Note: Ram sizes below 8KB will _repeat_ in the "cart ram" address range.
```
ISSOtm, line 216: Not necessarily, that depends on the mapper.

ISSOtm (changes requested):

content/The_Cartridge_Header.md
TheThief: The mapper isn't aware of the ram size outside the bank number - the address pins are wired directly to the ram chip. The chips only have as many pins as they need for their size - a 2KB ram has 11 address pins. Accessing at offset 0x0800 (2048) of the ram region will set pin 12 to 1 and the 11 to zero - only the 11 are connected, so the ram will dutifully respond with the contents of address 0, wrapping itself into the space.

In the case of MBC2, the MBC2 chip only has 9 address pins for the ram - 512 possible addresses. Higher addresses would set higher pins that aren't connected to the MBC so it just returns as if the address was masked to 9 bits, which results in it repeating, exactly the same as a separate ram chip.

content/The_Cartridge_Header.md
TheThief: The only known cart with that mapper is Momotarou Collection 2 - which is a 1MB ROM. Not 1.1, 1.2, or 1.5.

TheThief: Regarding other odd MBCs:
MBC6 is used in one cart, and has 1MB of rom and 1MB flash - not 1.1, 1.2, or 1.5
MBC7 - a handful of related carts, all also 1 MB

content/The_Cartridge_Header.md
CasualPokePlayer: No commercial game actually used the 2KB RAM configuration so arguing that is as fruitful as arguing how ROM_RAM worked. The MBC2 is the only place this applies for commercial games (which is different anyways since it's not a separate RAM, but probably identical anyways). In fact the MBC7 would be a somewhat a counterexample of this claim (invalid address ranges of it just read out FF instead of looping back). MBCs don't necessarily have to have nothing past RAM too, they could potentially stuff it with registers of extra hardware.

TheThief: >No commercial game actually used the 2KB RAM configuration
I had a look, and it seems you're right - it's a common claim of PD roms but probably wasn't accurate if they were ever put on actual carts (most likely a flash cart with 8KB or more ram was used if any).

It seems it's only relevant for MBC2, so perhaps it would be better to put it on that page.

> In fact the MBC7 would be a somewhat a counterexample of this claim (invalid address ranges of it just read out FF instead of looping back).
MBC7 doesn't have or support RAM. The claim specifically mentions ram.

CasualPokePlayer: Hence somewhat, there isn't RAM per se (the closest being the EEPROM) but what is there does not loop around as the claim would say.

TheThief: Ah but it does - the addresses are listed as like Ax5x - because they repeat every 0x100!

aaaaaa123456789:
ROM and RAM are acronyms and they should be spelled that way, not "rom" or "ram".

TheThief:
> ROM and RAM are acronyms and they should be spelled that way, not "rom" or "ram".

Too true! I'm normally a stickler for that...

content/The_Cartridge_Header.md
daid: > No commercial game actually used the 2KB RAM configuration

I think it might be good to mark each header option that was never seen in the wild as so.

avivace:
Given #108 , this looks good to me. @ISSOtm can you mark the conversations you opened as resolved if you're okay with the clarifications?

content/The_Cartridge_Header.md
TheThief: Anyway, I moved the note to MBC2, and added a note about the 2KB size and "PD" ROMs.

avivace (approved):

TheThief:
When you merge it back, might want to flatten it... I added a couple more commits fixing formatting etc.

avivace:
> When you merge it back, might want to flatten it... I added a couple more commits fixing formatting etc.

Don't worry, we're gonna rebase/squash in any case

ISSOtm:
Gonna re-review from scratch

content/MBC2.md
```diff
@@ -18,13 +18,18 @@ battery to save data during power-off though. As the data consists of
 4bit values, only the lower 4 bits of the "bytes" in this memory area
 are used.
 
+### A200-BFFF - 15 "echoes" of A000-A1FF
```
ISSOtm, line 21: Markdown expects a newline below titles

content/MBC2.md
```diff
@@ -18,13 +18,18 @@ battery to save data during power-off though. As the data consists of
...
+Only the bottom 9 bits of the address are used to index into the internal
+RAM, so RAM access repeats.
```
ISSOtm, line 23: As per #66, are there sources for this claim?

content/MBC2.md
```diff
@@ -18,13 +18,18 @@ battery to save data during power-off though. As the data consists of
...
 ## Control Registers
 
 ### 0000-3FFF - RAM Enable and ROM Bank Number (Write Only)
 
 This address range is responsible for both enabling/disabling the RAM
-and for controlling the ROM bank number. Bit 8 of the address determines
-whether to control the RAM-enable flag or the ROM bank number.
+and for controlling the ROM bank number. Bit 8 of the address (the least
+significant bit of the upper address byte) determines whether to control
+the RAM-enable flag or the ROM bank number.
```
ISSOtm, line 32: I don't like the inconsistency between "RAM-enable" and "ROM bank number", maybe remove the hyphen?

content/The_Cartridge_Header.md
```diff
@@ -187,6 +187,12 @@ Specifies the ROM Size of the cartridge. Typically calculated as "N such that 32
 |`$06`|   2 MByte |128 banks|
 |`$07`|   4 MByte |256 banks|
 |`$08`|   8 MByte |512 banks|
+
+The following ROM types are listed on older pandocs but the source is unknown,
```
ISSOtm, line 191: "older versions of this document"?

content/The_Cartridge_Header.md
```diff
@@ -208,6 +214,11 @@ Specifies the size of the external RAM in the cartridge (if any).
 When using a MBC2 chip $00 must be specified in this entry, even though
 the MBC2 includes a built-in RAM of 512 x 4 bits.
 
+Note: The 2 KB RAM size was not used in any production cartridges.
```
ISSOtm, line 217: Nintendo's docs actually mention that $01 shouldn't be used (see page 299 of [this archived manual](https://ia803208.us.archive.org/9/items/GameBoyProgManVer1.1/GameBoyProgManVer1.1.pdf)), so I'm not entirely sure where the "2 KB" number came from.

ISSOtm (changes requested):

content/MBC2.md
TheThief: Yes - the mooneye docs for one:

https://gekkio.fi/files/gb-docs/gbctr.pdf
https://github.com/Gekkio/mooneye-gb/blob/master/tests/emulator-only/mbc2/ram.s

I also found a pin diagram of the MBC2 chip - it only has address pins for the bottom 9 bits and the top 2 - the top two are used to select between ROM 0/1 areas and the cart ram, the bottom 9 index the ram. Addresses with bits 10+ set are simply invisible to the chip as it doesn't have pins for them: https://fms.komkon.org/GameBoy/Tech/Cart2.gif

CasualPokePlayer: mooneye also has test roms that specifically test for this and they have been verified to pass with a real MBC2.

https://github.com/Gekkio/mooneye-gb/blob/master/tests/emulator-only/mbc2/ram.s

content/The_Cartridge_Header.md
TheThief: That's a good point - it does follow the x4 progression of the other values, but maybe it should just be noted as invalid?

content/MBC2.md
ISSOtm: Alright, though I would also copy the note that the upper 4 bits of reads are undefined *and should not be relied upon*. That's a fairly important thing to know (if for some reason one were to use MBC2).

content/The_Cartridge_Header.md
ISSOtm: As noted in https://github.com/gbdev/pandocs/pull/108#discussion_r567311144, it's probably an early mistake. That said, since this line is being taken care of by the other PR, I would rather drop it from this one, to avoid conflicts.

ISSOtm:
Ping @TheThief

TheThief:
I'm waiting for the final version of the other PR to be merged before I make changes to this one relating to 2k ram etc.

ISSOtm:
#108 has just been merged.

TheThief:
@ISSOtm would you re-review?

ISSOtm (changes requested):
Just the last unresolved request to go, otherwise LGTM.

TheThief:
> Just the last unresolved request to go, otherwise LGTM.

Actually, I added this text already: "The upper 4 bits of each byte are undefined and should not be
relied upon."

ISSOtm:
The unresolved item talks about *copying* that notice to the wrapping behavior of the MBC2's SRAM.

TheThief:
> The unresolved item talks about _copying_ that notice to the wrapping behavior of the MBC2's SRAM.

Ah I get you now. I thought I'd only added the first notice after you made the comment, but yes it makes sense to copy it into the echo section.

ISSOtm (approved):

avivace:
Looks good to me, thanks @TheThief for working through this! :partying_face:

PR #107, opened 2021-01-29
Improve and expand STOP description in Reducing Power Consumption

daid:
Change the details of the STOP instruction to be more accurate.

### DIFF
--- a/content/Reducing_Power_Consumption.md
+++ b/content/Reducing_Power_Consumption.md
@@ -59,7 +59,7 @@ power standby mode. For example, a program may use this feature when it
 hasn't sensed keyboard input for a longer period (for example, when
 somebody forgot to turn off the Game Boy).
 
-No licenced rom makes use of STOP outside of CGB speed switching.
+No licensed rom makes use of STOP outside of CGB speed switching.
 Special care need to be taken if you want to make use of the STOP
 instruction.
 
@@ -71,7 +71,7 @@ black screen. Except if the LCD is in Mode3, where it will keep drawing
 the current screen.
 
 STOP is terminated by one of the P10 to P13 lines going low. Which means
-before entering STOP P14 and/or P15 should be asserted by writting $00,
+before entering STOP P14 and/or P15 should be asserted by writing $00,
 $10 or $20 to the P1 register, depending on which buttons you want
 to terminate the STOP on.
 
--- a/content/Reducing_Power_Consumption.md
+++ b/content/Reducing_Power_Consumption.md
@@ -71,8 +71,9 @@ black screen. Except if the LCD is in Mode3, where it will keep drawing
 the current screen.
 
 STOP is terminated by one of the P10 to P13 lines going low. Which means
-before entering STOP P14 and/or P15 should be active by writting $30 to the
-P1 register.
+before entering STOP P14 and/or P15 should be asserted by writting $00,
+$10 or $20 to the P1 register, depending on which buttons you want
+to terminate the STOP on.
 
 If STOP is invoked while a button is helt down, it might enter a bugged
 STOP mode. Which functions like a HALT instead of STOP, so requires
--- a/content/Reducing_Power_Consumption.md
+++ b/content/Reducing_Power_Consumption.md
@@ -59,14 +59,24 @@ power standby mode. For example, a program may use this feature when it
 hasn't sensed keyboard input for a longer period (for example, when
 somebody forgot to turn off the Game Boy).
 
-Before invoking STOP, it might be required to disable Sound and Video
-manually (as well as IR-link port in CGB). Much like HALT, the STOP
-state is terminated by interrupt events. STOP is commonly terminated
-with a joypad interrupt.  A bug may keep STOP from terminating if STOP
-was invoked while a button was held.
+No licenced rom makes use of STOP outside of CGB speed switching.
+Special care need to be taken if you want to make use of the STOP
+instruction.
 
-During STOP mode, the display will turn white, so avoid using it in your
-game's main loop.
+On a DMG, disable the LCD before invoking STOP, leaving the LCD enabled
+will draw a horizontal black line on the screen and damage the hardware.
+
+On CGB, leaving the LCD enabled when invoking STOP will return in a
+black screen. Except if the LCD is in Mode3, where it will keep drawing
+the current screen.
+
+STOP is terminated by one of the P10 to P13 lines going low. Which means
+before entering STOP P14 and/or P15 should be active by writting $30 to the
+P1 register.
+
+If STOP is invoked while a button is helt down, it might enter a bugged
+STOP mode. Which functions like a HALT instead of STOP, so requires
+interrupts to be enabled.
 
 # Disabling the Sound Controller
 
--- a/content/Reducing_Power_Consumption.md
+++ b/content/Reducing_Power_Consumption.md
@@ -63,21 +63,17 @@ No licensed rom makes use of STOP outside of CGB speed switching.
 Special care need to be taken if you want to make use of the STOP
 instruction.
 
-On a DMG, disable the LCD before invoking STOP, leaving the LCD enabled
-will draw a horizontal black line on the screen and damage the hardware.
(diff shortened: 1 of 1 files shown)


content/Reducing_Power_Consumption.md
```diff
@@ -59,14 +59,25 @@ power standby mode. For example, a program may use this feature when it
...
 
-Before invoking STOP, it might be required to disable Sound and Video
-manually (as well as IR-link port in CGB). Much like HALT, the STOP
-state is terminated by interrupt events. STOP is commonly terminated
-with a joypad interrupt.  A bug may keep STOP from terminating if STOP
-was invoked while a button was held.
-
-During STOP mode, the display will turn white, so avoid using it in your
-game's main loop.
+No licenced rom makes use of STOP outside of CGB speed switching.
```
avivace, line 62: We are currently using the `license` spelling, can you keep that?

In case you want to switch to the British variant we you can open an Issue about that.

content/Reducing_Power_Consumption.md
```diff
@@ -59,14 +59,25 @@ power standby mode. For example, a program may use this feature when it
...
+
+On a DMG, disable the LCD before invoking STOP, leaving the LCD enabled
+will draw a horizontal black line on the screen and damage the hardware.
+
+On CGB, leaving the LCD enabled when invoking STOP will return in a
+black screen. Except if the LCD is in Mode3, where it will keep drawing
+the current screen.
+
+STOP is terminated by one of the P10 to P13 lines going low. Which means
+before entering STOP P14 and/or P15 should be asserted by writting $00,
```
avivace, line 74: `writting` typo?

avivace (changes requested):

content/Reducing_Power_Consumption.md
daid: only 1 typo? ... that's below my average...

content/Reducing_Power_Consumption.md
daid: I didn't even see that there are two ways to spell license, so no hidden meaning.

avivace (approved):

content/Reducing_Power_Consumption.md
```diff
@@ -59,14 +59,21 @@ power standby mode. For example, a program may use this feature when it
...
-Before invoking STOP, it might be required to disable Sound and Video
-manually (as well as IR-link port in CGB). Much like HALT, the STOP
-state is terminated by interrupt events. STOP is commonly terminated
-with a joypad interrupt.  A bug may keep STOP from terminating if STOP
-was invoked while a button was held.
-
-During STOP mode, the display will turn white, so avoid using it in your
-game's main loop.
+No licensed rom makes use of STOP outside of CGB speed switching.
+Special care need to be taken if you want to make use of the STOP
```
ISSOtm, line 63: Actually, "need" should be "needs".

content/Reducing_Power_Consumption.md
```diff
@@ -59,14 +59,21 @@ power standby mode. For example, a program may use this feature when it
...
+instruction.
+
+On a DMG, disabling the LCD before invoking STOP, leaves the LCD enabled
```
ISSOtm, line 66: The second comma should be at the end of this line, not here.

content/Reducing_Power_Consumption.md
```diff
@@ -59,14 +59,21 @@ power standby mode. For example, a program may use this feature when it
...
+drawing a horizontal black line on the screen and damaging the hardware.
```
ISSOtm, line 67: Is it proven that it damages the hardware? AFAIK, it still hasn't been proven nor disproven

content/Reducing_Power_Consumption.md
```diff
@@ -59,14 +59,21 @@ power standby mode. For example, a program may use this feature when it
...
+
+On CGB, leaving the LCD enabled when invoking STOP will return in a
```
ISSOtm, line 69: Not sure about the "return in" there?

content/Reducing_Power_Consumption.md
```diff
@@ -59,14 +59,21 @@ power standby mode. For example, a program may use this feature when it
...
+black screen. Except if the LCD is in Mode3, where it will keep drawing
```
ISSOtm, line 70: There should be a space between "Mode" and "3".

content/Reducing_Power_Consumption.md
```diff
@@ -59,14 +59,21 @@ power standby mode. For example, a program may use this feature when it
...
+the current screen.
+
+STOP is terminated by one of the P10 to P13 lines going low. For this
```
ISSOtm, line 73: Referring to joypad input as "lines" is probably confusing if you don't know the terminology prior, imo they should be changed to more common terms.

Further, behavior resulting from writing $00 should probably made explicit.

Finally, wasn't there an edge case, where the state would only be terminated on a falling *edge*, so any buttons already pressed would not exit STOP immediately? I also remember @LIJI32 talking about STOP potentially locking up in a certain configuration, but I may be misremembering.

ISSOtm (commented):
There are some problems with this PR, some being debatable, so I'd like to collect opinions before submitting a fixup.

content/Reducing_Power_Consumption.md
daid: It gives the same result as LCD off outside VBlank, which the nintendo manual mentioned as causing damage. Also, someone on discord did mention having semi-permanent damage after LCD off outside vblank. And it is quite clearly driving the LCD outside of specs.

content/Reducing_Power_Consumption.md
daid: my head remembers this as "results", so:
> ... will result in a ...

content/Reducing_Power_Consumption.md
daid: I initially had something about bugged stop in there, but because I wasn't a 100% sure, it was removed an #110 was opened. So we can use that issue to add that information.

daid: About the "lines" that's my EE background getting in there. The P1 part calls it inputs, which would work better and be more consistent indeed.

content/Reducing_Power_Consumption.md
ISSOtm: Nintendo's manual ([archive](https://ia803208.us.archive.org/9/items/GameBoyProgManVer1.1/GameBoyProgManVer1.1.pdf)) actually only mentions, on page 255, that:

> In early DMGs, a black horizontal line appears on the screen if the LCDC is stopped (LCDC register bit 7 ← 0) at any time other than during vertical blanking.

It also mentions, oddly enough, that LCDC should only be turned off when reading $9*1* or greater from `rLY`—not $90 as software typically does. Finally, this is only listed as *recommended*, whereas I would have expected it to be listed as "required" if it could damage hardware at all.

The same "programming cautions" section has an entry (page 257) about `stop`, but only notes that `rIE` should be set to 0 beforehand.

content/Reducing_Power_Consumption.md
ISSOtm: To reduce the cognitive load, and since the relevant input lines directly map to `rP1` bits, I would talk about those, unless that assumption is wrong.

content/Reducing_Power_Consumption.md
daid: Well, if you try it, it ignores the contrast setting, which is a pretty big red flag in my book. And even if the manual doesn't mention it, check on discord who tested this. (or find the old conversation)

I also rather be on the side of caution here.

ISSOtm: Me too, hence why I'd prefer if it was hinted as highly likely, but not guaranteed.

content/Reducing_Power_Consumption.md
avivace: This was my bad, I agree

content/Reducing_Power_Consumption.md
avivace: Agree

PR #108, opened 2021-01-29
Note which options are not used by licensed games

daid:
So both ROM developers know that they shouldn't use those options, and Emulator developers know that there isn't a "correct" way to implement these options.

### DIFF
--- a/content/The_Cartridge_Header.md
+++ b/content/The_Cartridge_Header.md
@@ -170,7 +170,7 @@ cartridge, and if further external hardware exists in the cartridge.
 |`$FE`|  HuC3|
 |`$FF`|  HuC1+RAM+BATTERY|
 
-*) No licensed game makes use of this option. Exact behaviour is unknown.
+*) No licensed cartridge makes use of this option. Exact behaviour is unknown.
 
 ### 0148 - ROM Size
 
--- a/content/The_Cartridge_Header.md
+++ b/content/The_Cartridge_Header.md
@@ -170,7 +170,7 @@ cartridge, and if further external hardware exists in the cartridge.
 |`$FE`|  HuC3|
 |`$FF`|  HuC1+RAM+BATTERY|
 
-*) No licenced game makes use of this option. Exact behavour is unknown.
+*) No licensed game makes use of this option. Exact behaviour is unknown.
 
 ### 0148 - ROM Size
 
--- a/content/The_Cartridge_Header.md
+++ b/content/The_Cartridge_Header.md
@@ -147,8 +147,8 @@ cartridge, and if further external hardware exists in the cartridge.
 | $03 | MBC1+RAM+BATTERY
 | $05 | MBC2
 | $06 | MBC2+BATTERY
-| $08 | ROM+RAM (*)
-| $09 | ROM+RAM+BATTERY (*)
+| $08 | ROM+RAM *
+| $09 | ROM+RAM+BATTERY *
 | $0B | MMM01
 | $0C | MMM01+RAM
 | $0D | MMM01+RAM+BATTERY
@@ -170,7 +170,7 @@ cartridge, and if further external hardware exists in the cartridge.
 | $FE | HuC3
 | $FF | HuC1+RAM+BATTERY
 
-*) No licensed cartridge makes use of this option. Exact behaviour is unknown.
+\* No licensed cartridge makes use of this option. Exact behaviour is unknown.
 
 ### 0148 - ROM Size
 
@@ -188,11 +188,11 @@ Specifies the ROM Size of the cartridge. Typically calculated as "N such that 32
 | $06 |   2 MByte | 128
 | $07 |   4 MByte | 256
 | $08 |   8 MByte | 512
-| $52 | 1.1 MByte | 72 *)
-| $53 | 1.2 MByte | 80 *)
-| $54 | 1.5 MByte | 96 *)
+| $52 | 1.1 MByte | 72 *
+| $53 | 1.2 MByte | 80 *
+| $54 | 1.5 MByte | 96 *
 
-*) Only listed in unofficial docs
+\* Only listed in unofficial docs
 
 ### 0149 - RAM Size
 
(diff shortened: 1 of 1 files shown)


content/The_Cartridge_Header.md
```diff
@@ -170,6 +170,7 @@ cartridge, and if further external hardware exists in the cartridge.
 |`$FE`|  HuC3|
 |`$FF`|  HuC1+RAM+BATTERY|
 
+*) No licenced game makes use of this option. Exact behavour is unknown.
```
avivace, line 173: Use `license` spelling

content/The_Cartridge_Header.md
```diff
@@ -208,6 +209,8 @@ Specifies the size of the external RAM in the cartridge (if any).
 When using a MBC2 chip $00 must be specified in this entry, even though
 the MBC2 includes a built-in RAM of 512 x 4 bits.
 
+*) No licenced game makes use of this option. Exact behavour is unknown.
```
avivace, line 212: Use `license` spelling

avivace (changes requested):
Looks pretty good to me, just the `licence` -> `license` spelling

content/The_Cartridge_Header.md
avivace: `behavour` is a typo

TheThief:
I prefer the term "production carts" (or cartridges). Various third-party cartridges were released that weren't officially licensed games.

daid:
> I prefer the term "production cartridges" (or just "cartridges"). Various third-party cartridges were released that weren't officially licensed games.
> 
> Also, not every cartridge was a game.

Small issue with that, the cartridge list says there is 1 with the 2kb sram option, but it is marked as unlicensed. I haven't checked if this is actually a cart that was produced.

CasualPokePlayer:
> > I prefer the term "production cartridges" (or just "cartridges"). Various third-party cartridges were released that weren't officially licensed games.
> > Also, not every cartridge was a game.
> 
> Small issue with that, the cartridge list says there is 1 with the 2kb sram option, but it is marked as unlicensed. I haven't checked if this is actually a cart that was produced.

The cartridge list also has various ROM hacks, and it's probably unlikely all of them actually produce cartridges of these hacks.

content/The_Cartridge_Header.md
```diff
@@ -208,6 +209,8 @@ Specifies the size of the external RAM in the cartridge (if any).
 When using a MBC2 chip $00 must be specified in this entry, even though
 the MBC2 includes a built-in RAM of 512 x 4 bits.
 
+*) No licensed cartridge makes use of this option. Exact behaviour is unknown.
```
ISSOtm, line 212: Actually, where does the "2 KB" number come from? See [this comment](https://github.com/gbdev/pandocs/pull/106#discussion_r567310926).

daid: No idea. Maybe an older version the manual, pre-CGB?

I also have a "source" for "MBC4" values: http://www.dotmatrixgame.com/documentation/gb-cartridge-header.html

In general, there seems a lot of noise in the details of the gameboy information. But the 2KB is listed everywhere, so it could have been an early mistake that was copied by everyone. In any case, we cannot just "hide" it from the list now, that would just cause people to wonder why this list is incomplete.

ISSOtm: Agreed, though I think it'd be better to mention that the source is unknown, then. Or, change the entry to "Unused, *) but unofficial docs list it as 2 KB"?

nitro2k01: @daid That's not a "source" for MBC4 or anything else. It even plainly states (or at least implies) that the information is copied from the wiki. Looking back through the historical documents in this repo, MBC4 was first mentioned in Martin Korth's 2001 version of Pandocs. I'd say either Martin interpolated the values himself, or find the details in some other doc.

daid: I was not serious on the MBC4, it is an example of just bad info out there. That site is pretty much the only location that mentions it now in 2021

daid:
Of the 9438 roms I could find, only 102 use the 2KB ram option. All of those are (PD) or (Unl), and a few are marked as bad dumps. And 15 roms use values > 0x05 for the RAM field.

None of the 9438 roms use ROM+RAM or ROM+RAM+BATTERY.

None of the 9438 roms use ROM size $52, $53 or $54. And there are only 5 roms which have out of range values which are all (Unl)

And, I found zero roms that have the CGB flag sets for "PGB mode" which is currently mentioned in the pandocs as "TODO + Research"

ISSOtm:
PGB mode was confirmed by @LIJI32, so it's definitely real.

If 2K cannot be confirmed, then I suggest again "Unused, *) but unofficial docs list it as 2 KB".

Once that is done, unless @avivace has other opposition, this should be merged.

LIJI32:
PGB mode *probably* can't even be used in plain ROMs because it seems to require very dedicated hardware in the cartridge instead of a plain ROM+MBC.

avivace:
LGTM, thanks everybody. 

Can we add the *unconfirmed*/*unused* note on the 2 KB thingy?

ISSOtm:
Ping @daid

daid:
Sorry, it slipped my notification list. So thanks for the ping.

I've also marked the rom sizes > $50 as unofficial. As zero roms in my giant rom set use this option, and there is no documentation to backup this claim, and it doesn't really make sense to have those sizes of rom.

My guess it that it was used by someone in an attempt to "fix" certain bad dumps or something, where they didn't understand the MBC and thus got a strange number of banks.

TheThief:
I would remove the >$50 sizes from the table and put a note underneath that other docs mention them but the source is unknown and no known ROMs use them.

I had this in my other pull request:

```
The following ROM types are listed on some older unofficial documentation but the source is unknown, and no cartridges or ROMs using them are known.

|code | Size      | Banks |
|-----|-----------|--------------|
| $52 | 1.1 MByte | 72 banks |
| $53 | 1.2 MByte | 80 banks |
| $54 | 1.5 MByte | 96 banks |
```

ISSOtm (approved):

avivace (approved):

avivace:
This looks good to me, but @TheThief solution also makes sense and maybe this could set a precedent (keep unconfirmed stuff separate)

daid:
While TheThiefs solution works really well for tge rom size, I would preferrer keeping a table sorted over 2 tables.

PR #113, opened 2021-02-05
Update SOU_TRN info

ISSOtm:
From my experience with it.
Also clean up the SOUND ID tables while at it.


akumanatt:
Also, since we are at SGB functions, can you add tile data format in CHR_TRN? It's the same as SNES's 4bpp tile format.

PR #124, opened 2021-02-12
Clarification

AlvaroBurnett:
When WX is 7, the Window X position is 0. Therefore WX is "Window X Position plus 7"


ISSOtm (changes requested):
The formatting for this title is because it combines two register descriptions, so its formatting is currently consistent with all others, just repeated twice.

If you think this formatting is better for register description titles, it should be applied consistently.

Arguably, splitting this in two anchors might be better, but I'm not sure how to do that

ISSOtm (approved):

PR #126, opened 2021-02-13
Fix explanation

AlvaroBurnett:
To display a sprite in the upper left corner of the screen, we set sprite X=8, Y=16. Therefore:

- Byte 0 specifies the sprites vertical position on the screen (plus 16)
- Byte 1 specifies the sprites horizontal position on the screen (plus 8)

### DIFF
--- a/content/Video_Display.md
+++ b/content/Video_Display.md
@@ -620,8 +620,8 @@ Y=160 hides a sprite.
 
 ### Byte1 - X Position
 
-X = Sprite's horizontal position on the screen + 8. This works like the examples
-above but simpler, because the width of a sprite is always 8. An
+X = Sprite's horizontal position on the screen + 8. This works similarly
+to the examples above, except that the width of a sprite is always 8. An
 off-screen value (X=0 or X\>=168) hides the sprite, but the sprite still
 affects the priority ordering, thus other sprites with lower priority may be
 left out due to the ten sprites limit per scan-line.
--- a/content/Video_Display.md
+++ b/content/Video_Display.md
@@ -620,7 +620,8 @@ Y=160 hides a sprite.
 
 ### Byte1 - X Position
 
-X = Sprite's horizontal position on the screen + 8. An
+X = Sprite's horizontal position on the screen + 8. This works like the examples
+above but simpler, because the width of a sprite is always 8. An
 off-screen value (X=0 or X\>=168) hides the sprite, but the sprite still
 affects the priority ordering, thus other sprites with lower priority may be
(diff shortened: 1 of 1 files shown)


content/Video_Display.md
```diff
@@ -609,12 +609,12 @@ four bytes with the following meanings:
 
 ### Byte0 - Y Position
 
-Specifies the sprites vertical position on the screen (minus 16). An
+Specifies the sprites vertical position on the screen (plus 16). An
```
ISSOtm, line 612: Since both ways around make sense in their own way ("write the position plus 16" / "the value there is the position when 16 is subtracted"), I'd rather write e.g. "Specifies the sprite's vertical position on the screen, with an offset of 16 pixels. For example, writing 17 will have the sprite's topmost row on the third screen line."

While we're at this, please correct "sprites" to "sprite**'**s".

Finally, we may want to further clarify the behavior for partially off-screen values (e.g. 2, or 154; careful with 8x16 mode)

ISSOtm (changes requested):

AlvaroBurnett:
I made some changes. Please check if it's correct.

content/Video_Display.md
```diff
@@ -609,15 +609,22 @@ four bytes with the following meanings:
...
+Y=16 displays a sprite at the top of the screen,
+Y=144 displays a 8x16 sprite aligned with the bottom of the screen,
+Y=152 displays a 8x8 sprite aligned with the bottom of the screen,
+Y=154 displays the first six rows of a sprite at the bottom of the screen,
+Y=160 hides a sprite.
 
 ### Byte1 - X Position
 
-Specifies the sprites horizontal position on the screen (minus 8). An
+X = Sprite's horizontal position on the screen + 8. An
```
ISSOtm, line 623: It's probably a good idea to avoid repeating examples like above, but I'd just point out that it's "like the above", in case someone skips over the section above and gets confused.

ISSOtm (changes requested):

content/Video_Display.md
```diff
@@ -620,7 +620,8 @@ Y=160 hides a sprite.
 
 ### Byte1 - X Position
 
-X = Sprite's horizontal position on the screen + 8. An
+X = Sprite's horizontal position on the screen + 8. This works like the examples
```
avivace, line 623: "This works like the examples above but simpler" doesn't sound clear enough for me.
Maybe "This works similarly to the examples above"

avivace (changes requested):

avivace (approved):

PR #127, opened 2021-02-13
Change proposed

AlvaroBurnett:
If we could display unlimited sprites on a single scan line we wouldn't care about hidden sprites and their priorities (we would just display them all). However we can only display the ten sprites with the highest priorities. Thus the real problem is not their priorities, but the fact that these hidden sprites count towards the limit of ten.


ISSOtm (approved):

AlvaroBurnett:
Not needed any more because I just included this change here: #126

PR #129, opened 2021-02-13
Improve grammar

AlvaroBurnett:

PR #133, opened 2021-02-13
Improve grammar

AlvaroBurnett:
This was fixed in a previous patch but was accidentally undone when a later patch was applied.


ISSOtm:
...as noted in https://github.com/gbdev/pandocs/pull/129#issuecomment-778705393 :p

PR #136, opened 2021-02-16
Exclude discussion about sprites and metasprites

AlvaroBurnett:
Sprites and metasprites are software concepts. Since this document is about Game Boy hardware features (not game programming in general), I don't think that sprites and metasprites should be mentioned.

### DIFF
--- a/content/Rendering.md
+++ b/content/Rendering.md
@@ -71,9 +71,10 @@ on the screen.
 *Objects* are made of 1 or 2 stacked tiles (8x8 or 8x16 pixels).
 
 ::: tip NOTE
-Several objects can be combined (in that case they are called *metasprites*) to draw
-a larger graphical element, usually called "sprite". A sprite is a collection of tiles
-used to display one frame of a game character's animation.
+Several objects can be combined (they can be called *metasprites*) to draw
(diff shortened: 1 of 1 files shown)


ISSOtm (changes requested):
The point of this discussion was to clarify terminology commonly used; it's not strictly relevant to GB hardware, but the term "sprite" is largely misused in the more general gaming/programming community, including some meanings out of hardware scope.

So, while they're in theory out of scope for this document, as a GB programming introduction, I believe those *are* relevant.

avivace:
I agree with both of you. This kind of information is not related to GB and can add more confusion trying to explain those concepts. However they are important and commonly used and should be mentioned if we aim to provide something besides the bare hardware fundamentals.

Maybe we can move this information in a "info" box so it's separate from the rest but can still give some context?

ISSOtm:
Agreed—it might be useful, too, to mention in that box that the term "sprite" sees a lot of misuse in the more common language? (Don't we have a "disambiguation" box template, or am I misremembering?)

ISSOtm (approved):

avivace (approved):

PR #139, opened 2021-02-17
Mystery I/O register

AlvaroBurnett:
Does the $FF50 register exist? I don't see it mentioned anywhere else.


CasualPokePlayer:
This does in fact exist ([rBANK](https://github.com/ISSOtm/gb-bootroms/blob/443d7f057ae06e8d1d76fa8083650cf0be2cd0ae/src/cgb.asm#L18)), in fact you can literally see at the end of any boot ROM that is it set to a non-zero value (in reality, it is a 1 bit register where 0 has the boot rom accessible and 1 has the boot rom inaccessible).

AlvaroBurnett:
Cool. Thanks.

PR #140, opened 2021-02-17
Explain STAT interrupt in a different way

AlvaroBurnett:


ISSOtm (changes requested):
As-is, this can't be accepted. The mechanism is fully understood, and the new explanation inserts some probability terms ("may not trigger", "usually goes right after").

If you didn't find the original description clear enough, we can work towards improving it.

On that topic, paging @LIJI32, as I know there are some subtleties that are not covered by this. (Mode 2...)

And finally, while we're editing this, can we rename "LCDC status interrupt" to either "LCD status interrupt" or "STAT interrupt"? `LCDC` has absolutely *nothing* to do with it, so the name's more confusing than anything else.

AlvaroBurnett:
I said "may not trigger" and "usually goes right after" because sometimes there is a Mode 1 between the Mode 0 and Mode 2. If we made the example for Mode 2 and Mode 3 instead, then there wouldn't be any doubt that one goes always right after the other.

content/Video_Display.md
```diff
@@ -105,11 +105,11 @@ hidden by the text box.
...
 As we saw in the description of the STAT register, the LCD Controller cycles
 through the different modes in a fixed order. If we set the STAT bits
-in such a way that they would interrupt the LCD Controller at two
+in a way that they would interrupt the LCD Controller at two
 consecutive modes, then the second interrupt will not trigger. So for example,
-if the Mode 0 and LY=LYC interrupts are enabled, the LY=LYC interrupt may
-not trigger. This is because the LY=LYC interrupt triggers during Mode 2
-(which usually goes right after Mode 0).
+if we enable the interrupts for Mode 0 and LY=LYC (which triggers during Mode 2),
+the LY=LYC interrupt will not trigger unless we are lucky and there is a Mode 1
```
avivace, line 111: `unless we are lucky` seems very out of context here. Maybe `unless a Mode 1 happens` but it'll need some rephrasing specifying the once per frame thing.

avivace (changes requested):

ISSOtm (changes requested):
I see that you're trying to turn technical descriptions into something more "noob-friendly".

I'm neither for nor against it *per se*, but this begs the more general question of whether we want Pan Docs to be "user-friendly" descriptions, or more technical ones. I guess e.g. [gb-ctr](https://github.com/Gekkio/gb-ctr) could serve the latter purpose, and Pan Docs the former.

I just think that then, we should make sure that any technical information reworded out of Pan Docs is present elsewhere.

content/Video_Display.md
```diff
@@ -103,11 +103,12 @@ the handler disable sprites. This can be used if you use the window for
 a text box (at the bottom of the screen), and you want sprites to be
 hidden by the text box.
 
-The interrupt is triggered when transitioning from "No conditions met"
-to "Any condition met", which can cause the interrupt to not fire.
-Example : the Mode 0 and LY=LYC interrupts are enabled ; since the
-latter triggers during Mode 2 (right after Mode 0), the interrupt will
-trigger for Mode 0 but fail to for LY=LYC.
+As we saw in the description of the STAT register, the LCD Controller cycles
```
ISSOtm, line 106: Maybe this should be turned into a warning box? It's a behavior that may surprise people, after all.

content/Video_Display.md
```diff
@@ -4,7 +4,7 @@
 A *dot* is the shortest period over which the PPU can output one pixel: is it equivalent to 1 T-state on DMG or on CGB single-speed mode or 2 T-states on CGB double-speed mode. On each dot during mode 3, either the PPU outputs a pixel or the fetcher is stalling the [FIFOs](#pixel-fifo).
 :::
 
-### FF41 - STAT - LCDC Status (R/W)
+### FF41 - STAT (R/W)
```
ISSOtm, line 7: "LCD status" or "PPU status" should be a fitting description.

ISSOtm (approved):

ISSOtm:
(Oops, misclick, I meant to "Changes requested".)

ISSOtm (approved):

avivace (approved):

PR #143, opened 2021-02-18
Complete the paragraph that explains tile addressing

AlvaroBurnett:


ISSOtm (commented):
Isn't this redundant with the line just below?

AlvaroBurnett:
Yes, I think that we can safely delete the part that says:
"To put it differently, "8000 addressing" takes tiles 0-127 from block 0 and tiles 128-255 from block 1, whereas "8800 addressing" takes tiles 0-127 from block 2 and tiles 128-255 from block 1."

ISSOtm:
Why delete that one instead of the proposed one? The current description is indeed somewhat redundant, but given how much trouble newcomers have proven to have with all these concepts, I still think it's good.

avivace:
Looks better to me now, what do you think @ISSOtm ?

avivace (approved):

ISSOtm (approved):
I'm not entirely sure, but not opposing either, so let's roll with this.

PR #146, opened 2021-02-18
"Normal background" is just the background

AlvaroBurnett:


ISSOtm (commented):
This was written as "normal" BG because the window is essentially an override on the BG mechanism, and so behaves like it. If it doesn't cause confusion regarding that, I'd agree to this change.

PR #147, opened 2021-02-18
Sprite priorities - Condense two sentences and explain when X is off-screen

AlvaroBurnett:


ISSOtm (commented):
I don't find this edit particularly better, I'll let someone else say what they think of it.

avivace (approved):

avivace:
Both changed parts sound better and clearer to me.

PR #148, opened 2021-02-18
Sprite priorities - OBJ-to-BG priority

AlvaroBurnett:


ISSOtm (changes requested):
I find this more confusing than the original, since it made the distinction between OBJ-to-OBJ priority and OBJ-to-BG priority clear.

Maybe it could help to instead mention at the beginning that OBJ-to-OBJ priority is computed before OBJ-to-BG?

avivace (approved):

avivace:
@ISSOtm I think it sounds better now. What do you think?

content/Video_Display.md
```diff
@@ -670,10 +670,11 @@ sprite in OAM (\$FE00-\$FE03) has the highest priority, and so on. In
 Non-CGB mode, the smaller the X coordinate, the higher the priority. The
 tie breaker (same X coordinates) is the same priority as in CGB mode.
 
-The priority calculation between sprites disregards OBJ-to-BG Priority
-(attribute bit 7). Only the highest-priority nonzero sprite pixel at any
+After sprites priorities have been determined, in those areas where they overlap
+the OBJ-to-BG Priority (attribute bit 7) of the lower priority sprites is ignored.
```
ISSOtm, line 674: I think this is confusing, and would put it after the following sentence: the OBJ-to-BG priority doesn't get ignored if there's overlap at all, it just only gets considered *afterwards*—as the next sentence explains. I believe that swapping the order would remove any ambiguity.

ISSOtm (changes requested):

content/Video_Display.md
AlvaroBurnett: How is it considered afterwards? Sorry I don't understand.

ISSOtm: As you may know if you've read the docs about the FIFO, the GB actually has a FIFO for BG pixels and another one for OBJ pixels.

OBJ-to-OBJ priority is considered when inserting pixels into the OBJ FIFO (to decide whether an incoming pixel should overwrite the current one or not).

OBJ-to-BG priority is considered each time the FIFOs are shifted, to select of which FIFO to keep the pixel.

Therefore, OBJ-to-OBJ priority is considered at an earlier stage than the OBJ-to-BG priority, and the OBJ-to-OBJ priority process does not consider the OBJ-to-BG bit. The combination of these is what causes the quirk.

content/Video_Display.md
```diff
@@ -670,10 +670,11 @@ sprite in OAM (\$FE00-\$FE03) has the highest priority, and so on. In
 Non-CGB mode, the smaller the X coordinate, the higher the priority. The
 tie breaker (same X coordinates) is the same priority as in CGB mode.
 
-The priority calculation between sprites disregards OBJ-to-BG Priority
-(attribute bit 7). Only the highest-priority nonzero sprite pixel at any
-given point is compared against the background. Thus if a sprite with a
-higher priority (based on OAM index) but with OBJ-to-BG Priority turned
+Priority among pixels that overlap is determined using the rules explained
```
ISSOtm, line 673: Priority is only considered among *opaque* pixels, though. An opaque pixel will always trump a transparent one, regardless of priority.

content/Video_Display.md
```diff
@@ -670,10 +670,11 @@ sprite in OAM (\$FE00-\$FE03) has the highest priority, and so on. In
...
+above. After the pixel with the highest priority has been determined,
+the OBJ-to-BG priority of only that pixel is honored (or disregarded if
```
ISSOtm, line 675: I think the "only" should be emphasized.

content/Video_Display.md
```diff
@@ -670,10 +670,11 @@ sprite in OAM (\$FE00-\$FE03) has the highest priority, and so on. In
...
+this is a zero pixel). Thus if a sprite with a
```
ISSOtm, line 676: I see what you mean, but I think it may help to mention that these are transparent pixels. Pointing out that these are pixels with index 0 (hence, regardless of palette) is probably a good idea indeed.

ISSOtm (changes requested):
Also, do we want to put all of this paragraph in a box? I'd suggest the "tip" type, then.

PR #149, opened 2021-02-18
Fix grammar

AlvaroBurnett:


ISSOtm (approved):
Shouldn't the "still **proof**" line 728 be changed also? The intention seems to be "It's still *safe*".

avivace (approved):

PR #157, opened 2021-02-18
Interrupts: Improve language

AlvaroBurnett:


content/Interrupts.md
```diff
@@ -57,44 +57,44 @@ interrupt "waits" until both IME and IE allow its execution.
...
-register becomes automatically reset by the CPU, and the IME flag
-becomes cleared (disabling any further interrupts until the program
+When an interrupt is executed, the corresponding bit in the IF
+register is automatically reset by the CPU, and the IME flag
+is also reset (disabling any further interrupts until the program
 re-enables the interrupts, typically by using the RETI instruction), and
-the corresponding Interrupt Vector (that are the addresses in range
-0040h-0060h, as shown in IE and IF register decriptions above) becomes
+the corresponding Interrupt Vector (which is one of the addresses in the range
+0040h-0060h, as shown in the IE and IF register decriptions above) is
```
ISSOtm, line 65: We use prefix `$` for hex.

ISSOtm, line 65: Maybe we could use an anchor link?

content/Interrupts.md
```diff
@@ -57,44 +57,44 @@ interrupt "waits" until both IME and IE allow its execution.
...
 
 1. More than one interrupt signal changed from Low to High at the same time.
 2. Several interrupts have been requested during a time in which IME/IE didn't allow these interrupts to be executed directly.
 3. The user has written a value with several "1" bits (for example 1Fh) to the IF register. 
 
-Provided that IME and IE allow the execution of more than one of the
-requested interrupts, then the interrupt with the highest priority
-becomes executed first. The priorities are ordered as the bits in the IE
-and IF registers, Bit 0 (V-Blank) having the highest priority, and Bit 4
+IME and IE allow the execution of more than one of the
```
ISSOtm, line 84: No, this *is* only if both interrupts are able to be serviced at the same time.

ISSOtm (changes requested):

content/Interrupts.md
```diff
@@ -57,44 +57,44 @@ interrupt "waits" until both IME and IE allow its execution.
...
+The CPU automatically disables all the other interrupts by setting IME=0
 when it executes an interrupt. Usually IME remains zero until the
-interrupt procedure returns (and sets IME=1 by the RETI instruction).
-However, if you want any other interrupts of lower or higher (or same)
-priority to be allowed to be executed from inside of the interrupt
-procedure, then you can place an EI instruction into the interrupt
+interrupt procedure returns (and sets IME=1 by means of the RETI instruction).
+However, if you want any other interrupts (of any priority)
+to be allowed to be executed from inside the interrupt
+procedure, then you can use the EI instruction in the interrupt
```
ISSOtm, line 97: Given that we have "procedure" twice, maybe replace this instance with "handler"?

ISSOtm (changes requested):

content/Interrupts.md
AlvaroBurnett: The problem with using different words for the same thing is that people may think that we are talking about different things. I would use "handler" everywhere.

ISSOtm: Agreed!

avivace: Agree

content/Interrupts.md
```diff
@@ -57,45 +57,45 @@ interrupt "waits" until both IME and IE allow its execution.
...
 unless/until IME and IE allow its execution.
 
 ### Interrupt Priorities
 
-In the following three situations it might happen that more than 1 bit in the IF register are set, requesting more than one interrupt at once:
+In the following three situations it might happen that more than one bit in the IF register is set, requesting more than one interrupt at once:
 
 1. More than one interrupt signal changed from Low to High at the same time.
 2. Several interrupts have been requested during a time in which IME/IE didn't allow these interrupts to be executed directly.
 3. The user has written a value with several "1" bits (for example 1Fh) to the IF register.
```
ISSOtm, line 82: While we're here, `1Fh` → `$1F`

ISSOtm (changes requested):

avivace (approved):

ISSOtm (approved):

PR #160, opened 2021-02-19
CPU instruction set: Improve language

AlvaroBurnett:


content/CPU_Instruction_Set.md
```diff
@@ -1,5 +1,5 @@
-Tables below specify the mnemonic, opcode bytes, clock cycles, affected
-flags (ordered as znhc), and explanation. The timings assume a CPU
+Tables below specify the mnemonic, encoding, cycles, affected
```
ISSOtm, line 1: Those are actually clocks (4 MiHz), not cycles (1 MiHz). The distinction, as explained below, is very important.

ISSOtm (changes requested):

content/CPU_Instruction_Set.md
AlvaroBurnett: Then should the table headings be modified also to say "Clock cycles" instead of "cycles"?

ISSOtm: Please do!

ISSOtm (approved):

PR #168, opened 2021-02-20
Modify description of LCDC bit 0

AlvaroBurnett:


CasualPokePlayer:
This isn't really accurate (and is inconsistent with a later part of the docs). The meaning is more BG Display or BG and Window Priority. https://gbdev.io/pandocs/#lcdc-0-bg-window-display-priority

ISSOtm:
When that bit is clear on DMG, the BG FIFO is disabled, but that also causes the window not to be displayed. Hence, the later part of the docs is actually wrong.

By the way, wouldn't it be nicer if the rows linked to the relevant anchors?

AlvaroBurnett:
Is it ok if we replace "Non-CGB Mode" with "DMG Mode"?

ISSOtm:
The problem is that on one hand there's CGB (and AGB) in Color mode, and OTOH there's everything else. How is that "everything else" best referred to?

AlvaroBurnett:
Is there any difference between the Non-CGB mode of the CGB and the Non-CGB mode of the AGB?

ISSOtm:
None significant, if any.

ISSOtm (approved):

Issue #60, opened 2020-09-19
ToC depth


avivace:
We need to revision and possibily rethink how deep is the ToC, since it provides the main way to navigate the document.

From a comment on Discord:
> Yeah, I think that is a key issue in finding info quicky. The video section has only 1 main header but contains a lot of information in subheaders.

ISSOtm:
Couldn't we auto-generate it? Even a `sed` script should be enough to auto-generate one, so it shouldn't be too difficult.

I don't see why we'd want to maintain it manually, either: all good documents the size of Pan Docs begin with a ToC at least 2, maybe 3 levels deep.

ISSOtm:
Basically, we need to auto-gen the ToC.

Here are some questions:
- Should it be entirely auto-generated?
- What depth should it be generated at?
- Should it be all at the beginning, or should sections have their individual ToCs, like GBATEK?

ISSOtm:
Closed by #292

PR #170, opened 2021-02-20
memory mapped I/O

AlvaroBurnett:
This is bigger than the old excerpt but I think it's worth it.


ISSOtm:
The table is only there as a summary of the various I/O ranges. I'm against it because it shouldn't be a substitute for a proper table of contents (#60), but *this* is really a ToC in table form. Rejecting.

Issue #167, opened 2021-02-19
Check anchor links


avivace:

avivace:
@pinobatch are you working on this?

pinobatch:
I've started. Here's an example of what I have so far. Is it at all useful?
```
Dangling internal links (whose fragment matches no id= value)
    <a href="#ff68-bcps-bgpi-cgb-mode-only-background-color-palette-specification-or-background-palette-index">BCPS/BGPI</a>,
    <a href="#ff69-bcpd-bgpd-cgb-mode-only-background-color-palette-data-or-background-palette-data">BCPD/BGPD</a>,
    <a href="#ff4a-wy-window-y-position-r-w-ff4b-wx-window-x-position-minus-7-r-w">above</a> for the definition of "when the window is visible".)
    <a href="#gameboy-camera">Game Boy Camera</a>).

Empty sections
    # FF51 - HDMA1 (New DMA Source, High) - CGB Mode Only 
    # FF53 - HDMA3 (New DMA Destination, High) - CGB Mode Only 
    # FF72 - Bits 0-7 (Read/Write) 
    # Comparison with Z80 
    # A000-BFFF - RAM Bank 00-03, if any (Read/Write) 
```
And if so, would I make it a script that parses an HTML file (or HTML on standard input) using [html5lib](https://pypi.org/project/html5lib/) and produces a list of errors like this in a text file (or on standard output)?

avivace:
Check also some remarks about the implementation here: https://github.com/gbdev/pandocs/wiki

PR #172, opened 2021-02-21
Add HTML internal link checker

pinobatch:
I wrote a Python script to check an HTML document for these errors:
- Same `id=` value on multiple elements
- Internal links whose fragment matches no `id=` value in the document
- Empty sections (e.g. two consecutive `<h4>` or an `<h4>` followed by an `<h2>`)

See if it's useful for #167 at all. An empty output means no errors. I have no experience with GitHub Actions, so I'll be watching edits to this PR to see what happens.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,7 +4,7 @@ on:
   pull_request
 
 jobs:
-  build:
+  run-tests:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout gbdev/pandocs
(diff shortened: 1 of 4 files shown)


ISSOtm:
So, this should be run as `render/proposed/checkanchors.py index.html`, and output nothing, right?

pinobatch:
Ideally, yes: an absence of output indicates no warnings. I don't know what the expectation is for a linter's exit code.

Also I put it in the `proposed` directory to stay out of the way of your CI. Feel free to move it once you integrate it.

avivace:
Thanks @pinobatch, this works as expected. I'll add this into the GitHub Actions pipeline as soon as I have some time

PR #176, opened 2021-02-21
similar -> similarly

AlvaroBurnett:


ISSOtm (changes requested):
s/opcodes/instructions/gi

content/Memory_Map.md
```diff
@@ -27,7 +27,7 @@ The following addresses are supposed to be used as jump vectors:
 
 However, the memory may be used for any other purpose in case that your
 program doesn't use any (or only some) RST commands or interrupts. RST
-commands are 1-byte opcodes that work similar to CALL opcodes, except
+commands are 1-byte instructions that work similarly to CALL instructions, except
```
ISSOtm, line 30: I'd rather say that "`rst` is a 1-byte instruction that works similarly to the 3-byte `call` instruction, except that the destination address is restricted.". What do you think?

ISSOtm (changes requested):

content/Memory_Map.md
AlvaroBurnett: I agree. Fixed.

ISSOtm (approved):
Will wait on others' feedback to be sure about readability.

avivace (approved):

PR #182, opened 2021-02-21
DMA - fix language and hex notation

AlvaroBurnett:


content/Video_Display.md
```diff
@@ -290,46 +290,46 @@ of this brightness correction.
 
 Writing to this register launches a DMA transfer from ROM or RAM to OAM
 memory (sprite attribute table). The written value specifies the
-transfer source address divided by 100h, that is, source & destination are:
+transfer source address divided by $100, that is, source and destination are:
 
 ```
-Source:      XX00-XX9F   ;XX in range from 00-F1h
-Destination: FE00-FE9F
+Source:      $XX00-$XX9F   ;XX = $00 to $F1
```
ISSOtm, line 296: Wait... why up to $F1? It shouldn't go higher than $DF... and strictly speaking, all values do *something*, it's just "weird" when it gets higher than $FE, and $E0+ access echo RAM...

content/Video_Display.md
```diff
@@ -290,46 +290,46 @@ of this brightness correction.
...
 this procedure to start the transfer from inside HRAM, and wait until
 the transfer has finished:
 
 ```
  run_dma:
-  ld a, start address / 100h
-  ldh  (FF46h),a ;start DMA transfer (starts right after instruction)
-  ld  a,28h      ;delay...
+  ld a, start address / $100
+  ldh  ($FF46),a ;start DMA transfer (starts right after instruction)
```
ISSOtm, line 312: RGBDS syntax uses brackets instead of parens.

content/Video_Display.md
```diff
@@ -290,46 +290,46 @@ of this brightness correction.
...
 reason, the programmer must copy a short procedure into HRAM, and use
 this procedure to start the transfer from inside HRAM, and wait until
 the transfer has finished:
 
 ```
  run_dma:
-  ld a, start address / 100h
-  ldh  (FF46h),a ;start DMA transfer (starts right after instruction)
-  ld  a,28h      ;delay...
+  ld a, start address / $100
```
ISSOtm, line 311: Is this really clear? Alternatives include `>> 8` and `HIGH()`. (Note the 2 more instances below.)

content/Video_Display.md
```diff
@@ -290,46 +290,46 @@ of this brightness correction.
...
   ret
 ```
 
 Because sprites are not displayed while OAM DMA is in progress, most
-programs execute this procedure from inside of their VBlank
-procedure. But it is also possible to execute it during display redraw
-also, allowing to display more than 40 sprites on the screen (that is, for
-example 40 sprites in upper half, and other 40 sprites in lower half of
+programs execute this procedure from inside of their V-Blank
+handler. But it is also possible to execute it during display redraw,
```
ISSOtm, line 322: I think it should be clarified that "display redraw" means Mode 2 and 3; further, while it's in progress, the PPU reads all of OAM as $FF, possibly causing graphical errors if started during Mode 3 as a sprite is about to be drawn.

ISSOtm (changes requested):

content/Video_Display.md
AlvaroBurnett: I like $100 because when you see it you instantly recognize that it means "remove the last two digits of that address". But the decision is yours.

content/Video_Display.md

content/Video_Display.md
AlvaroBurnett: Are the graphical errors different if the DMA transfer starts while in Mode 2 instead of in Mode 3?

ISSOtm: If it starts during Mode 3 as a sprite is about to be drawn, the tile and/or attr may be read as $FF, potentially producing a small graphical error. Apparently, the PPU re-reads Y and X (as $FF), and so normally doesn't display OBJ anymore. I kinda wonder why, but it seems that it's how it is.

AlvaroBurnett: Updated

content/Video_Display.md
```diff
@@ -290,58 +290,59 @@ of this brightness correction.
...
-procedure. But it is also possible to execute it during display redraw
-also, allowing to display more than 40 sprites on the screen (that is, for
-example 40 sprites in upper half, and other 40 sprites in lower half of
-the screen), at the cost of a couple lines that lack sprites.
+programs execute this procedure from inside their V-Blank
+handler. But it is also possible to execute it during display redraw (Modes 2 and 3),
+allowing to display more than 40 sprites on the screen (that is, for
+example 40 sprites in the top half, and other 40 sprites in the bottom half of
+the screen), at the cost of a couple lines that lack sprites due to the fact that
+during those couple lines the PPU reads OAM as $FF.
```
ISSOtm, line 326: Apparently, glitches are possible if starting OAM DMA during Mode 3? (cc @hacktix) If we can't get conclusive evidence of the timing, we should mention that they are at least possible.

ISSOtm (changes requested):

content/Video_Display.md
Hacktix: Mode 2 and 3 specifically, confirmed again just today when I was adding a test ROM for exactly this to GBEmulatorShootout.

ISSOtm: Great! If the behavior is understood, could it be documented on Pan Docs, please?

AlvaroBurnett: Done

Hacktix: Well, it's not exactly understood. What's certain is that OAM reads don't always return $FF, as modifying the data that's transferred during either mode 2 or 3 can change attributes of the visual glitches. (I only managed to change the tile ID in my experiments through effectively brute-force so far, hoping to find a pattern to it eventually)

ISSOtm: Okay. @AlvaroBurnett's changes should be good enough for now, we'll just open a separate issue for this.

content/Video_Display.md

ISSOtm (approved):

PR #193, opened 2021-02-23
Revert "replace "first ROM bank" with "ROM bank 0""

avivace:
Reverts gbdev/pandocs#186


avivace:
The bank at 0000-3FFF is able to be swapped around by some MBCs (most notably the MBC1). It should rather be called either "ROM0", "ROM0 space", or "$0000-3FFF".

Thanks @ISSOtm , @CasualPokePlayer

PR #202, opened 2021-03-04
LCDC7: describe PPU

AlvaroBurnett:


ISSOtm (changes requested):
Imo the PPU *should* be described somewhere in the docs, but rather in a dedicated "system architecture" section at the top, not here. This is because it's a term, along with e.g. "APU", which is references all around the document.

AlvaroBurnett:
I understand. Should we place this description in one of the existing files, or maybe create a new one named "System_Architecture.md"?

ISSOtm:
Good question...

avivace:
> Imo the PPU _should_ be described somewhere in the docs, but rather in a dedicated "system architecture" section at the top, not here. This is because it's a term, along with e.g. "APU", which is references all around the document.

I agree

> I understand. Should we place this description in one of the existing files, or maybe create a new one named "System_Architecture.md"?

New file is better, so we can import it somewhere in the introductory part and give some context

avivace (changes requested):
Move to another file

AlvaroBurnett:
Replaced with a proper PR.

PR #209, opened 2021-03-04
added A,B register startup values for different models

jkotlinski:


content/Power_Up_Sequence.md
```diff
@@ -11,8 +11,10 @@ If the least significant byte of the result is a not a zero, then the Game Boy w
 If it is a zero, then the internal ROM is disabled and cartridge program execution begins at location $100 with the following register values:
 
 ```
-  AF=$01B0
-  BC=$0013
+  A =$01-GB/SGB, $FF-GBP, $11-GBC
```
ISSOtm, line 14: We use the console codenames, so DMG, SGB, MGB, SGB2, CGB, and AGB.

ISSOtm (changes requested):
Aren't the register values different on CGB in DMG mode if a palette is selected? It should probably be mentioned that a CGB will still return $11 even in compatibility mode.

jkotlinski:
Sorry, I do not know if the register values different on CGB in DMG mode if a palette is selected. I will humbly retract my PR.

ISSOtm:
Oh, please don't close it; that should be able to be determined through either static analysis from [gb-bootroms](https://github.com/ISSOtm/gb-bootroms/blob/master/src/cgb.asm), or by dynamically checking in an emulator.

Someone else can build upon the PR, keeping it open reminds everyone that this *is* an issue with the docs ^^

jkotlinski:
OK. Added #211 as a reminder for further investigation

PR #222, opened 2021-03-07
add PPU overview

AlvaroBurnett:
I think that this is the right place to give an overview of the PPU because it goes just after the talk about tiles, palettes and layers, which are the main ingredients of the PPU.


ISSOtm (changes requested):
I think that this should be part of a more general description of the hardware architecture (see #202)

Issue #194, opened 2021-02-23
Add Glossary


avivace:
I think a preliminary section mentioning some jargon and commonly used acronyms (e.g. LSB, MSB) could be helpful. It could also introduce some of the standards and rules we adopt in the document, adding some clarifications (e.g. units, bases)

ISSOtm:
As per #287, "dots" is introduced twice.

brainard52:
@avivace requested that I make a new issue, but I'll just hijack this one.

I think a glossary with links to each section the terms appear in would be great. It could additionally include definitions that appear on mouse-over of the term (maybe with a delay).

ISSOtm:
I can try adding markup for that as part of #292.

brainard52:
> I can try adding markup for that as part of #292.

Nice. If you have a plaintext release of pandoc, I can get a list of words that aren’t typically in an english dictionary for you

ISSOtm:
Does [this Markdown render](https://github.com/gbdev/pandocs/files/6511640/index.md) do the job?

brainard52:
That’s great. If this works, we might even be able to integrate the glossary generation into the build.

avivace:
To add: "CGB mode"

ISSOtm:
Markup possibilities:
https://github.com/commonmark/commonmark-spec/wiki/Deployed-Extensions#abbreviation
https://github.com/commonmark/commonmark-spec/wiki/Deployed-Extensions#definition-list

avivace:
Needs to be added to the Glossary: https://github.com/gbdev/pandocs/pull/350#discussion_r979296228

avivace:
Needs to be added to the Glossary: https://github.com/gbdev/pandocs/pull/534#pullrequestreview-1819981650

PR #223, opened 2021-03-07
create new System Architecture document

AlvaroBurnett:
And start populating it with a description of the PPU.

### DIFF
--- a/content/System_Architecture.md
+++ b/content/System_Architecture.md
@@ -1,0 +1,4 @@
+# PPU
+
+The PPU is the subsystem that reads graphics data from video RAM,
+translates that data into pixels and transfers them to the LCD controller.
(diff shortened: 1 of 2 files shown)


ISSOtm:
Alright, this looks like a good start! Should it go before the memory map? I think terms introduced in the architecture could then be used to better explain the memory map.

AlvaroBurnett:
> I think terms introduced in the architecture could then be used to better explain the memory map.

Yep, I agree. Fixed.

avivace:
LGTM

avivace (approved):

avivace:
Maybe it's related to https://github.com/gbdev/pandocs/issues/194 ?

ISSOtm:
Yeah, I think this should instead be superseded / expanded on by a more global fix to #194.

PR #224, opened 2021-03-07
tile index

AlvaroBurnett:


content/Video_Display.md
```diff
@@ -633,13 +633,17 @@ affects the priority ordering, thus other sprites with lower priority may be
...
+### Byte 2 - Tile Index
+
+In 8x8 mode (LCDC bit 2 = 0), this byte specifies the sprite's only tile index ($00-$FF).
+This unsigned value selects a tile from the memory area at $8000-$8FFF.
+In CGB Mode this could be either in
+VRAM bank 0 or 1, depending on bit 3 of the following byte.
+In 8x16 mode (LCDC bit 2 = 1), the memory area at $8000-$8FFF is still interpreted
+as a series of 8x8 tiles, where every 2 tiles form a sprite. In this mode, this byte
+specifies the index of the first (top) tile of the sprite. This is enforced by the
+hardware: the least significant bit of the tile index is ignored. IE: the top 8x8
```
ISSOtm, line 645: Please replace "IE" with "I.e." or "That is", so as not to confuse it for the register `IE`.

content/Video_Display.md
```diff
@@ -633,13 +633,17 @@ affects the priority ordering, thus other sprites with lower priority may be
...
+tile is "NN AND $FE", and the bottom 8x8 tile is "NN OR $01".
```
ISSOtm, line 646: Please use binary operators `&` and `|` instead of words, to make it clearer.

ISSOtm (changes requested):

content/Video_Display.md
avivace: If using the binary operators, the string should not be "A AND B" but `A & B`

ISSOtm (approved):

PR #230, opened 2021-03-07
STAT

AlvaroBurnett:


content/Video_Display.md
```diff
@@ -7,11 +7,11 @@ A *dot* is the shortest period over which the PPU can output one pixel: is it eq
 ### FF41 - STAT (LCD Status) (R/W)
 
 ```
-Bit 6 - LYC=LY Coincidence Interrupt (1=Enable) (Read/Write)
+Bit 6 - LYC=LY Interrupt             (1=Enable) (Read/Write)
 Bit 5 - Mode 2 OAM Interrupt         (1=Enable) (Read/Write)
 Bit 4 - Mode 1 V-Blank Interrupt     (1=Enable) (Read/Write)
 Bit 3 - Mode 0 H-Blank Interrupt     (1=Enable) (Read/Write)
-Bit 2 - Coincidence Flag  (0:LYC<>LY, 1:LYC=LY) (Read Only)
+Bit 2 - LYC=LY Flag           (0=False, 1=True) (Read Only)
```
ISSOtm, line 14: I think that "0=false, 1=true" is pretty self-explanatory, whereas e.g. "0=different, 1=equal" is clearer.

ISSOtm (changes requested):

content/Video_Display.md

ISSOtm (approved):

PR #240, opened 2021-03-07
IME

AlvaroBurnett:


content/Interrupts.md
```diff
@@ -1,27 +1,27 @@
...
+1 - Enable all interrupts that are enabled in the IE register (FFFF)
 ```
 
 The IME flag is used to disable all interrupts, overriding any enabled
-bits in the IE Register. It isn't possible to access the IME flag by
-using a I/O address, instead IME is manipulated by
-the following instructions/operations:
+bits in the IE register. It isn't possible to access the IME flag by
+using a I/O address. IME can be modified by
+the following instructions/operations only:
```
ISSOtm, line 11: I think "operations" should be changed here.

content/Interrupts.md
```diff
@@ -1,27 +1,27 @@
...
 
 ```
-EI     ;Enable Interrupts  (that is, IME=1)
-DI     ;Disable Interrupts (that is, IME=0)
-RETI   ;Enable Ints & Return (same as the instructions combination EI, RET)
-<INT>  ;Disable Ints & Call to Interrupt Vector
+EI     ;Enable interrupts  (that is, IME=1)
```
ISSOtm, line 14: I present tense should be used instead of imperative, so the `<INT>` entry makes more sense.

ISSOtm (changes requested):

content/Interrupts.md
AlvaroBurnett: Fixed

content/Interrupts.md
AlvaroBurnett: Not sure if this is what you wanted but I have tried to fix it.

ISSOtm: I meant e.g. "Enables interrupts and returns"

ISSOtm (approved):

PR #241, opened 2021-03-07
interrupt priorities

AlvaroBurnett:


avivace:
Why `%`?

Please check and comply to https://github.com/gbdev/pandocs/wiki/Document-Style. If you don't agree with any of the points, we must first change the guidelines then go on in applying them

AlvaroBurnett:
> Why `%`?
> 
> Please check and comply to https://github.com/gbdev/pandocs/wiki/Document-Style. If you don't agree with any of the points, we must first change the guidelines then go on in applying them

Oh ok, I used '%' because I believed that it was the standard here. Personally, I would vote for the C/C++ notation.
Fixed.

content/Interrupts.md
```diff
@@ -78,7 +78,7 @@ In the following three situations it might happen that more than one bit in the
 
 1. More than one interrupt signal changed from Low to High at the same time.
 2. Several interrupts have been requested during a time in which IME/IE didn't allow these interrupts to be executed directly.
-3. The user has written a value with several "1" bits (for example $1F) to the IF register. 
+3. The user has written a value with several "1" bits (for example 00011111 binary) to the IF register.
```
ISSOtm, line 81: Wouldn't "binary 00011111" be better?

AlvaroBurnett: Fixed

ISSOtm (approved):

PR #252, opened 2021-03-11
MBC5

AlvaroBurnett:


content/MBC5.md
```diff
@@ -24,12 +24,12 @@ Same as for MBC1, except that RAM sizes are 8 KiB, 32 KiB and 128 KiB.
 
 ### 0000-1FFF - RAM Enable (Write Only)
 
-Mostly the same as for MBC1, a value of $0A will enable reading and
-writing to external RAM. A value of $00 will disable it.
+Mostly the same as for MBC1. Writing $0A will enable reading and
+writing to external RAM. Writing $00 will disable it.
```
ISSOtm, line 28:
```suggestion
writing to external RAM. Writing $00 will disable it.

Actual MBCs actually enable RAM when writing any value whose bottom 4 bits equal $A (so $0A, $1A, and so on), and disable it when writing anything else.
Relying on this behavior is not recommended for compatibility reasons.
```

ISSOtm (changes requested):

AlvaroBurnett:
Applied the requested change, but GitHub complains about a bunch of things that seem unrelated to this change.

ISSOtm (approved):
They are unrelated indeed, they come from the checks introduced by #172; since this change doesn't introduce any new errors, it's fine.

avivace (approved):

PR #263, opened 2021-03-13
TIMA

AlvaroBurnett:


content/Timer_and_Divider_Registers.md
```diff
@@ -20,9 +20,9 @@ increment at 32768Hz in double speed.
 ### FF05 - TIMA - Timer counter (R/W)
 
 This timer is incremented at the clock frequency specified by the TAC
-register (\$FF07). When the value overflows (exceeds $FF)
+register (\$FF07). When the register overflows (exceeds $FF)
 it is reset to the value specified in TMA (FF06) and an interrupt
-is requested, as described below.
+may be requested, as described below.
```
ISSOtm, line 25: Under which circumstances does the timer overflow without an interrupt being requested?

AlvaroBurnett: Fixed

CasualPokePlayer: I think you're misunderstanding the term "requested" here. An Interrupt Request (ie IRQ) simply requests (ie set the specific bit in rIF) an Interrupt. The actual execution of the interrupt happens when the interrupt is enabled and an IRQ is pending, along with IME being 1.

AlvaroBurnett: Yep, sorry. Don't know what I was thinking lol...

PR #267, opened 2021-03-15
TAC

AlvaroBurnett:


content/Timer_and_Divider_Registers.md
```diff
@@ -40,6 +40,7 @@ Bits 1-0 - Input Clock Select
 ```
 
 ::: tip NOTE
+Select the clock frequency first, then enable the Timer.
```
ISSOtm, line 43: Do you mean that it should be done by a separate write? If so, why should that be done?

ISSOtm (changes requested):

content/Timer_and_Divider_Registers.md
AlvaroBurnett: Explained

content/Timer_and_Divider_Registers.md
```diff
@@ -40,6 +40,8 @@ Bits 1-0 - Input Clock Select
 ```
 
 ::: tip NOTE
+Select the clock frequency first, then enable the Timer. Starting the Timer before
+or at the same time as the clock frequency is selected may result in excessive count up operation.
```
ISSOtm, line 44: That sounds copy-pasted from the official manual; is it? If so, then research as to what the actual behavior is may be needed. (Looking to eliminate that "may".)

ISSOtm (changes requested):

content/Timer_and_Divider_Registers.md
AlvaroBurnett: We could ask @Gekkio, if he shows up in a near future. In the meantime, following the recommendation doesn't hurt.

ISSOtm: There's [already a document detailing the circuits](https://gbdev.io/pandocs/#timer-obscure-behaviour), so we should be able to figure this out without bothering him ^^

avivace: I agree. No point in reporting an apparently non-deterministic behaviour if we don't have enough proof

avivace (changes requested):

Issue #211, opened 2021-03-05
Add A,B register startup values for different models


jkotlinski:
Add A,B register startup values for different models to Power_Up_Sequence.md

ISSOtm:
Note that there may be other differences, notably given the CGB boot ROM's more complex behavior.

- [ ] Different regs in compatibility mode?
- [ ] Different regs if a palette is manually chosen?
- [ ] Different regs if the logo tilemap is written?

We probably also should document the CGB boot ROM's palette & tilemap selection behavior.

CasualPokePlayer:
Taking a quick look, it looks like the only time differences occur is with compatibility mode (and even then, if it is in DMG compatibility mode, then B is dependent on the title in the header and whether you have the right licensee bytes for built in palettes). Manually choosing palettes has no effect, and I'm not sure what you mean by logo tilemap being written?

In DMG Compatibility Mode (& Licensee)
`AF: 1180 ; 1100 for AGB`
`B: Sum of bytes in 0134-0143 inclusive & 0xFF ; +1 & 0xFF if AGB`
`C: 00`
`DE: 0008`
`HL: 007C`
In DMG Compatibility Mode (& Not Licensee)
`AF: 1180 ; 1100 for AGB`
`BC: 0000 ; 0100 for AGB`
`DE: 0008`
`HL: 007C`
In CGB Mode
`AF: 1180 ; 1100 for AGB`
`BC: 0000 ; 0100 for AGB`
`DE: FF56`
`HL: 000D`

Again this was a quick look so I might have missed something.

PR #269, opened 2021-03-18
Missing register values for different GB models

nefenit:
I found those values in historical versions of pandocs (gbspec.txt) and GB CPU Manual v 1.01.
"Detecting CGB (and GBA) functions " has both A = $11 in GBC/GBA and B = $01 in GBA  mentioned but is missing here.
This error probably occured because someone copy-pasted "Power Up Sequence" page from gbdev.gg8.se wiki which is incomplete.

### DIFF
--- a/src/Power_Up_Sequence.md
+++ b/src/Power_Up_Sequence.md
@@ -13,10 +13,9 @@ If the least significant byte of the result is a not a zero, then the Game Boy w
 If it is a zero, then the internal ROM is disabled and cartridge program execution begins at location $100 with the following register values:
 
 ```
-  A = $01-GB/SGB, $FF-GBP, $11-GBC/GBA
+  A = 0x01-GB/SGB, 0xFF-GBP/GBL/SGB2, 0x11-CGB
   F = $B0
-  B = $00-GB/SGB/GBP/GBC, $01-GBA
-  C = $13
+  BC = $0013
   DE = $00D8
   HL = $014D
   SP = $FFFE
--- a/src/Power_Up_Sequence.md
+++ b/src/Power_Up_Sequence.md
@@ -13,11 +13,14 @@ If the least significant byte of the result is a not a zero, then the Game Boy w
 If it is a zero, then the internal ROM is disabled and cartridge program execution begins at location $100 with the following register values:
 
 ```
-  AF=$01B0
-  BC=$0013
-  DE=$00D8
-  HL=$014D
-  Stack Pointer=$FFFE
+  A = $01-GB/SGB, $FF-GBP, $11-GBC/GBA
+  F = $B0
+  B = $00-GB/SGB/GBP/GBC, $01-GBA
+  C = $13
+  DE = $00D8
+  HL = $014D
+  SP = $FFFE
+  PC = $0100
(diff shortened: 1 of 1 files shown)


CasualPokePlayer:
see https://github.com/gbdev/pandocs/issues/211

nefenit:
dmg_bootrom.gb (DMG-01) last 2 lines of disassembly:
  LD A, $01
  LD $($FF00+$50), A
mgb_boot.bin (MGB-001) 2 last lines of disassembly:
  LD A,$FF
  LD ($FF00+$50),A

A = $01-GB, $FF-GBP
This behaviour is consistent and should be added.
In issue #211 it seems that it affects CGB in compatibility mode someone should check it.
A = $11 in GBC/GBA and B = $01 adding that to "Power Up Sequence" makes sense because this information should be in one place and not scared around.

ISSOtm:
Are you sure about all these values when selecting a CGB palette for DMG ROMs?

nefenit:
no, i don't know how they would behave

CasualPokePlayer:
A=$11 is constant, although the B=$01 is not. B can wildy change in DMG compatibility mode w/ Licensee (the title checksum meant for the predefined palettes for DMG games is left onto B, causing in theory any possible value for B). Also, manually changing palettes does not affect any registers, this can easily be confirmed with a bit of manual testing. The old logo tilemap being written could have an effect but I have not tested it.

nefenit:
I didn't knew that @CasualPokePlayer thanks for explaining
In GB Prog manual 1.0 i found on page 120 that A values are consistent across models
they also match up with what i found using boot ROMs and in previously found files
```
A = 0x01-GB/SGB, 0xFF-GBP/GBL/SGB2, 0x11-CGB
```
Would adding only that part be ok?

avivace (approved):

CasualPokePlayer:
> 
> 
> A=$11 is constant, although the B=$01 is not. B can wildy change in DMG compatibility mode w/ Licensee (the title checksum meant for the predefined palettes for DMG games is left onto B, causing in theory any possible value for B). Also, manually changing palettes does not affect any registers, this can easily be confirmed with a bit of manual testing. The old logo tilemap being written could have an effect but I have not tested it so I can't say for certain.

Doing some more tests, it does appear the old logo tilemap being written does change registers. Keep in mind you can only get the old logo tilemap written if you have the Nintendo Licensee bytes in the header along with DMG mode there too, so AF, BC, and DE will be identical to that situation (although, even then, there's only 2 versions of B possible here). However, HL will be different:
`HL: 991A`

The A value is an absolute constant so it's probably fine to include that in there, although you should use MGB instead of GBP, as GBP can also refer to the Game Boy Player (which the only usage of GBP in the Pandocs refers to the Game Boy Player in that context, with the GameBoy Pocket being referred to as MGB everywhere else).

avivace:
@ISSOtm what's the status on this?

ISSOtm:
Still needs some work, and I currently don't have time to do it myself.

ISSOtm:
I'll try to make a separate PR adding this info and more initial reg state (partial overlap with #309, so that it could be used as a reference for that), and improving the section's formatting.

PR #272, opened 2021-03-25
Correct HDMA transfer timing

mattcurrie:
In 1x mode HDMA transfers 16 bytes in 8 *M-cycles*, not 8 *T-states* (and it copies 16 bytes in 16 M-cycles in 2x mode).

See below capture from cartridge bus when running in 1x mode for reference.

hdma-1x


ISSOtm (approved):
LGTM. Could we host the image you attached, and *link*—not embed—to it as reference?

avivace (approved):

avivace:
Good for me too. Are you the author of the image? We can add it in the references

mattcurrie:
> Good for me too. Are you the author of the image? We can add it in the references

Yep, I can do a capture in 2x mode also if you think it'd be helpful?

ISSOtm:
Yes please!

mattcurrie:
HDMA when running in 2x mode:

hdma-2x

PR #276, opened 2021-04-10
Add rumble (MBC5) documentation with plots

Hacktix:
Added documentation on the MBC5 Rumble feature based on [Jeff Frohwein's Schematic](http://www.devrs.com/gb/files/rumble.gif) as well as results from self-conducted testing which can be viewed [here.](https://hacktix.github.io/GBEDG/mbcs/mbc5/#rumble)


GreenAndEievui:
I think you should mention oscillating the register at different frequencies to control the intensity.

content/MBC5.md
```diff
@@ -47,3 +47,9 @@ maps the corresponding external RAM bank (if any) into the memory area at
 A000-BFFF.
 
 ![](imgs/MBC5.png "imgs/MBC5.png")
+
+### Rumble
+
+On cartridges which feature a rumble motor, bit 3 of the RAM Bank register
+is connected to the Rumble circuitry. Setting the bit to 1 enables the rumble
```
ISSOtm, line 54:
```suggestion
is connected to the Rumble circuitry instead of the RAM chip. Setting the bit to 1 enables the rumble
```

content/MBC5.md
```diff
@@ -47,3 +47,9 @@ maps the corresponding external RAM bank (if any) into the memory area at
...
+motor and keeps it enabled until the bit is reset again.
```
ISSOtm, line 55:
```suggestion
motor and keeps it enabled until the bit is reset again.
To control the rumble's intensity, it should be turned on and off repeatedly, as seen with these two examples from Pokémon Pinball.

"Light":
![](<image>)

"Intense":
![](<image>)
```

ISSOtm (changes requested):

ISSOtm (approved):

content/MBC5.md
```diff
@@ -47,3 +47,20 @@ maps the corresponding external RAM bank (if any) into the memory area at
...
 
 ![](imgs/MBC5.png "imgs/MBC5.png")
+
+### Rumble
+
+On cartridges which feature a rumble motor, bit 3 of the RAM Bank register
+is connected to the Rumble circuitry instead of the RAM chip. Setting the
+bit to 1 enables the rumble motor and keeps it enabled until the bit is reset again.
+
+To control the rumble's intensity, it should be turned on and off repeatedly,
```
avivace, line 57: Is this some kind of PWM technique? Do we know more details on how fast the switch could happen (duty-cycle)?

avivace (changes requested):

avivace:
Great, thank you for the contribution, @Hacktix . Looks great to me.

My only issue is with the images: those diagrams are low resolution and have a watermark. Maybe we can find a way to generate actual plots out of the data? Having SVGs would be much more appreciated

Hacktix:
I could up the resolution a bit and edit out the watermark, I scaled the images down a little from the original because they took up a lot of space vertically, but if that's not an issue...

I could also try to convert the graphs to SVGs (or maybe the PicoScope software actually has SVG export features? Not exactly sure, but I'll check it out, I do still have the original capture files)

avivace:
> I could up the resolution a bit and edit out the watermark, I scaled the images down a little from the original because they took up a lot of space vertically, but if that's not an issue...
> 
> I could also try to convert the graphs to SVGs (or maybe the PicoScope software actually has SVG export features? Not exactly sure, but I'll check it out, I do still have the original capture files)

Maybe the software you are using can export the timeseries as raw data? That way we can generate a figure with any plotting library we want

content/MBC5.md
Hacktix: The rumble motor rotates quite intensely when turned on, so the frequent on-and-offs are just needed to not make it shake too much. The Pokémon Pinball examples show the "Intense" Rumble Setting flipping between on and off approximately every 17.5ms, so pretty close to a simple on/off switch per frame, which seems to work pretty well on hardware.

Hacktix:
> Maybe the software you are using can export the timeseries as raw data? That way we can generate a figure with any plotting library we want 

That's probably a much better idea than what I was planning on doing, it definitely can, so I'll get to making a new graph with that as soon as possible. (Probably within the next 24 hours :p)

avivace:
> > Maybe the software you are using can export the timeseries as raw data? That way we can generate a figure with any plotting library we want
> 
> That's probably a much better idea than what I was planning on doing, it definitely can, so I'll get to making a new graph with that as soon as possible. (Probably within the next 24 hours :p)

That's great! I think we can add an "src" folder inside "imgs" to put the raw file + the scripts to generate the plots from this raw data.

Let me know if you need help in any of the steps.

Hacktix:
I hope modifying the requirements.txt and merge.sh isn't an issue, I tried to make a little script that generates SVGs in the `imgs` folder while keeping it as modular as possible for potential re-use with other graphs.

avivace (changes requested):

avivace:
> I hope modifying the requirements.txt and merge.sh isn't an issue, I tried to make a little script that generates SVGs in the `imgs` folder while keeping it as modular as possible for potential re-use with other graphs.

Nice work, thanks. Modifying requirements.txt and merge.sh was the right thing to do to touch the pipeline 😄 .
I'm not super sure about keeping the name of the graph/plot as first line of the csv (making it not a proper/parsable csv) so I'd move that information to the script

content/imgs/src/Rumble_Medium.csv
```diff
@@ -0,0 +1,4889 @@
+"Light Rumble"
```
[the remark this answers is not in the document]
Hacktix: Oh boy, yeah, a bit of a naming inconsistency, I used "Medium" and "Light" interchangeably, there's only two settings in the actual game. And there it's called "Strong" and "Mild"... might wanna go through all references to these and make them consistent.

Hacktix:
> > I hope modifying the requirements.txt and merge.sh isn't an issue, I tried to make a little script that generates SVGs in the `imgs` folder while keeping it as modular as possible for potential re-use with other graphs.
> 
> Nice work, thanks. Modifying requirements.txt and merge.sh was the right thing to do to touch the pipeline 😄 .
> I'm not super sure about keeping the name of the graph/plot as first line of the csv (making it not a proper/parsable csv) so I'd move that information to the script

Newest commit contains changes to the graph_render script allowing for graph titles to be configured there, cause keeping the csvs parse-able does seem like a more clean solution. :p

I also renamed the Pokémon Pinball example files and graph titles to stay consistent with the ingame names of the rumble intensity settings.

avivace:
@Hacktix Great, thanks for sticking to this :partying_face: 

I added a bit of styling and upped a it the sizes to make the plots more fitting to the document. Set the font as the sans-serif face we use (Inter) and added a print to show in the logs the current job. The size should also a bit better now.

The only thing I really didn't manage to fix is the `x_label_rotation` which I want to set to 0 but when I do so it prints `...` as labels (that's why it's set at 0.01 now :laughing: )

PR #279, opened 2021-04-25
Include information about the window's internal line counter

Lycoder:
I'm not exactly sure on where to put this information in, so I just dropped it here, It would be good to describe an example too, maybe also a little diagram explaining it? I can do that if you want me to, but I decided to just keep it like this until I get some more info about PRs on this doc (such as keeping the design language of diagrams consistent).

### DIFF
--- a/content/Video_Display.md
+++ b/content/Video_Display.md
@@ -592,9 +592,9 @@ for the definition of "Window visibility".)
 
 ::: tip Window Internal Line Counter
 
-The window keeps an internal line counter that's functionally similar to LY, and increments alongside it. However, it only gets incremented when the window is visible, as described [here](https://github.com/gbdev/pandocs/blob/develop/content/Video_Display.md#ff4a---wy-window-y-position-rw-ff4b---wx-window-x-position--7-rw). That is whenever:
+The window keeps an internal line counter that's functionally similar to `LY`, and increments alongside it. However, it only gets incremented when the window is visible, as described [here](#ff4a---wy-window-y-position-rw-ff4b---wx-window-x-position--7-rw). That is whenever:
 
-1. The window is enabled
+1. The window is enabled (LCDC bits 5 and 0 (DMG Mode) set)
 2. `LY >= WY`
 3. `WX` has an on-screen value.
 
--- a/content/Video_Display.md
+++ b/content/Video_Display.md
@@ -594,7 +594,7 @@ for the definition of "Window visibility".)
 
 The window keeps an internal line counter that's functionally similar to `LY`, and increments alongside it. However, it only gets incremented when the window is visible, as described [here](#ff4a-wy-window-y-position-r-w-ff4b-wx-window-x-position-7-r-w). That is whenever:
 
-1. The window is enabled (LCDC bits 5 and 0 (DMG Mode) set)
+1. `LCDC.5` is set (window enabled)
 2. `LY >= WY`
 3. `WX` has an on-screen value.
 
--- a/content/Video_Display.md
+++ b/content/Video_Display.md
@@ -592,11 +592,7 @@ for the definition of "Window visibility".)
 
 ::: tip Window Internal Line Counter
 
-The window keeps an internal line counter that's functionally similar to `LY`, and increments alongside it. However, it only gets incremented when the window is visible, as described [here](#ff4a-wy-window-y-position-r-w-ff4b-wx-window-x-position-7-r-w). That is whenever:
-
-1. `LCDC.5` is set (window enabled)
-2. `LY >= WY`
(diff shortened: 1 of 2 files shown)


ISSOtm:
This section is more of a beginner-friendly overview, and not really the right place to get into nitty-gritty details. The other place where weird window behavior is described is the WX/WY registers, so maybe add an info bubble there?

Lycoder:
Yep, I just moved it to the bottom of WY's description, but I still feel this is not its place, as it is not directly related with WY anyways. I'm thinking [here](https://github.com/Lycoder/pandocs/blob/patch-1/content/Video_Display.md#window) might be better?

ISSOtm:
Right, yeah, I think that makes sense.

Lycoder:
There you go, I moved it and made it into an info bubble

content/Video_Display.md
```diff
@@ -590,6 +590,18 @@ Whether the Window is displayed can be toggled using
 (See [above](#ff4a-wy-window-y-position-r-w-ff4b-wx-window-x-position-7-r-w)
 for the definition of "Window visibility".)
 
+::: tip Window Internal Line Counter
+
+The window keeps an internal line counter that's functionally similar to LY, and increments alongside it. However, it only gets incremented when the window is visible, as described [here](https://github.com/gbdev/pandocs/blob/develop/content/Video_Display.md#ff4a---wy-window-y-position-rw-ff4b---wx-window-x-position--7-rw). That is whenever:
```
ISSOtm, line 595:
```suggestion
The window keeps an internal line counter that's functionally similar to LY, and increments alongside it. However, it only gets incremented when the window is visible, as described [here](#ff4a---wy-window-y-position-rw-ff4b---wx-window-x-position--7-rw). That is whenever:
```
This should only point to the anchor, not the GitHub render; otherwise it'll break in the HTML version.

content/Video_Display.md
```diff
@@ -590,6 +590,18 @@ Whether the Window is displayed can be toggled using
...
+
+1. The window is enabled
```
ISSOtm, line 597: This should point to LCDC bit 6, and perhaps LCDC bit 0 in non-CGB mode?

ISSOtm (changes requested):

Lycoder:
Done!

ISSOtm:
Are there test ROMs that check bit 6 and bit 0's effect on the window's "scanline counter"?

Lycoder:
You mean bit 5? Matt's dmg-acid2 tests this behavior [here](https://github.com/mattcurrie/dmg-acid2#chin) however he doesn't really go into detail on why there should be an internal window line counter, that's why I thought this should be documented here.
And I'm not sure about bit 0 stopping the counter, I'll ask him right now

ISSOtm (changes requested):

Lycoder:
Fixed?

Lycoder:
Matt confirmed to me that the window is still activated when LCDC.0 is reset, bit 5 is the one to care about. Got rid of that bit 0 info

ISSOtm:
Do you have a source for that?

content/Video_Display.md
```diff
@@ -590,6 +590,18 @@ Whether the Window is displayed can be toggled using
 (See [above](#ff4a-wy-window-y-position-r-w-ff4b-wx-window-x-position-7-r-w)
 for the definition of "Window visibility".)
 
+::: tip Window Internal Line Counter
+
+The window keeps an internal line counter that's functionally similar to `LY`, and increments alongside it. However, it only gets incremented when the window is visible, as described [here](#ff4a-wy-window-y-position-r-w-ff4b-wx-window-x-position-7-r-w). That is whenever:
+
+1. `LCDC.5` is set (window enabled)
+2. `LY >= WY`
```
mattcurrie, line 598: This condition is not completely correct, WY just had to have been equal to LY at the start of mode 2 at some point in the frame.

Perhaps it better to not repeat the window activation conditions here and instead just have the link to the WX/WY section?

mattcurrie:
> Do you have a source for that? I mean something that would be quoted/linked to within the docs, as per #66.

You may be familiar with this test, which checks the duration of mode 3 with LCDC.5 set and LCDC.0 not set 😉

https://github.com/mattcurrie/mealybug-tearoom-tests/blob/master/src/ppu/win_without_bg.asm

mattcurrie:
But yes, a specific test for whether the window's internal line counter is actually incremented when it is activated with LCDC.0 not set would be good too.

ISSOtm:
This would be good enough for the time being, imo.

avivace (approved):
Looks good to me! Great use of the tip box

PR #288, opened 2021-05-14
Clarify condition for triggering OAM corruption

ISSOtm:
Regarding references, Liji said he has yet to publish his OAM corruption test suite, but that blargg was the best we have in the meantime.

I think it's better to wait until we can point at a specific ROM rather than linking to blargg and overshooting the scope.


avivace:
We should at least open an Issue mentioning the hanging reference

avivace (approved):

Issue #57, opened 2020-09-12
Convert Vuepress custom containers 


avivace:
We need a script that pattern matches Vuepress custom containers, the syntax generating the info/error/warning boxes and converts them in plain text/latex/... to keep the markdown files agnostic and not tied not one tool

See #7

ISSOtm:
The way it's implemented is a hack, but the functionality has been added via our custom mdBook "renderer".

Issue #18, opened 2020-02-19
Headings are hard to discern


mid-kid:
I'm finding it hard to tell whenever I'm moving into a new section, both because the difference in size between them isn't very clear, and because the different types of headings seem to be used inconsistently. This makes it hard for me to skim the document and find the right section.

ISSOtm:
The heading consistency is something that we inherited from the OG document, I believe.

As for improving the feeling of switching sections, horizontal bars and/or a different background per heading (akin to the style used by nocash's docs) would probably help a lot.

avivace:
Yes, the problem in headings and subheadings is that they were used consistently in the wiki and the export added another layer of confusion, so you may see a register in h2 in one page and one in h4 in another.

avivace:
https://github.com/gbdev/pandocs/commit/f17d85b89227bd8c08d1e8319df1db1971d31515

mid-kid:
This seems to break under firefox ESR 68.4.2 ![](https://transfer.sh/IqP88/screenshot.png)

nitro2k01:
Imo, the current tentative fix is kind of a mess. One issue compared to the original version of Pan Docs, is that the original version has a ToC and sub-ToCs whereas we currently only have a single level of ToC. *All* sections linked from a ToC (whether from the main or a sub-ToC) are highlighted with the grey background.

My suggestions are:

1. Recreate the multi-level ToC from nocash's Pandocs. (Whether that ends up being a tree structure at the top, local ToC's in each section that needs one, or both.)
2. Like nocash's version, highlight all main headings in some way. Unlike nocash's version, probably highlight subsections differently, for example with a lighter background.

ISSOtm:
Can you check if [this preview](https://eldred.fr/pandocs) of #292 fixes this issue?

mid-kid:
That sidesteps the issue by just putting everything on a separate page, and having no way to view everything as a single page, so I'm not sure how it'd look like that. If the intention is to keep it all on separate pages I'd say it's solved, since this works even when printed to PDF (every section has a pagebreak), though the horizontal gray bars in for example [3.3 OAM](https://eldred.fr/pandocs/OAM.html) and [3.7 Palettes](https://eldred.fr/pandocs/Palettes.html) look out of place.

ISSOtm:
As far as I remember, the push for a single-page document was the ability to easily search it via Ctrl+F, but mdBook provides this functionality via the search bar. Were there any other reasons?
[...]

PR #292, opened 2021-05-18
Migrate to mdBook

### LINKED ISSUE #7: Usage of Vuepress' Custom Containers
Example:

image

Vuepress plugin: https://vuepress.github.io/en/plugins/container/#render

### LINKED ISSUE #19: Gray background headers
As suggested by @ISSOtm to potentially fix/improve the situation described in #18 , consider gray banded brackground heading/subheadings, as in nocash docs:

image

This requires (and thus it's blocked by) #18 consistency in the usage of headings (h1-h4)

### LINKED ISSUE #58: Empty sidebar?
Remove the sidebar if we don't plan on using it

### LINKED ISSUE #60: ToC depth

### LINKED ISSUE #289: Easier navigation of Pan Docs
I often have multiple tabs open that each have https://gbdev.io/pandocs/ loaded. Right now, the page title is 
Screen Shot 2021-05-16 at 10 55 49 AM

Being able to dynamically update the window/tab title to include the anchored section would be great. For example, I think something akin to 
Screen Shot 2021-05-16 at 10 57 05 AM would be perfect for https://gbdev.io/pandocs/#lcd-monochrome-palettes 

Alternatively, being able to track several sections all at once and easily jump between them via hotkeys would be amazing as well. I don't have the ability to mock it up, but I'm imagining a + next to each anchor link which would add an anchor link to a little floating side bar on the left. That way, I can just quickly navigate between sections I want to remember.

Some sort of sane hotkey system to navigate between "bookmarked" sections would be a great addition to that.

### LINKED ISSUE #290: Multiple page version
Pandocs is currently being served as a one long page, following the spirit of the original document.

Some users (as #289 ) expressed the desire to be able to navigate the document split in different pages (one per chapter?). It should be trivially possiblle to add another HTML export target having one chapter / subchapter per page, while also keeping the current version.

avivace:
Why:
- single and multi page layout
- themes, palettes, better theming
- custom backends
- custom language highlight.js 

What's missing:
- [x] info/warning boxes renderer
- [x] Styling those boxes
- [x] Re-wiring CI
- [x] Fixing links
- [x] Custom highlight.js
- [x] Making heading levels consistent
- [x] Make info/warning boxes render on print

Fixes #289
Fixes #290
Fixes #7, since we're not using Vuepress anymore
Fixes #19 by virtue of using custom themes, though a custom "Pan Docs"/"Classic" theme could be added.
Fixes #58 due to an entirely different layout
Fixes #60


docs/.nojekyll
```diff
@@ -0,0 +1 @@
+This file makes sure that Github Pages doesn't process mdBook's output.
```
avivace, line 1: This file is not needed, it's sufficient to not enable github pages with jekyll in the repository setting and github pages will just serve the selected folder / branch

docs/About
```diff
@@ -0,0 +1 @@
+# About
```
avivace, line 1: why is the output in this branch?

ISSOtm: Missing `.gitignore`...

docs/.nojekyll
ISSOtm: It's automatically generated, and I'm not seeing a way to disable it.

ISSOtm:
@avivace How are we going to handle the custom styling of the boxes? Also, should we offer the current Pan Docs color scheme as a theme? (If yes, there are a lot of corollary questions.)

ISSOtm:
Misc. changes to be made:

- LCDC.md, line 57 says "further below", but this is not *below* anymore.
- Video_Display.md is huge, and generates a beeg chapter without a ToC, so it should be split into multiple sub-chapters. (We can use the navbar's collapsing feature to avoid making it huge as well.)
- SGB_Functions.md is huge as well.
- CPU_Comparison_with_Z80.md compares with Z80 *and* 8080, so a slightly more generic name would be more appropriate.
- "MID-FRAME QUIRKS" box isn't marked up in the Video Display chapter.

ISSOtm:
How should glossary entries be marked up?

ISSOtm:
[A preview is now live](https://eldred.fr/pandocs)! Feedback is welcome.

ISSOtm:
Note: to test CI, as well as to update the preview, I am now pushing to <https://github.com/iSSOtm/pandocs/tree/master> instead.

ISSOtm:
I also consolidated the build and deploy CI workflows into a single one, so that the build's results can be immediately used. Deployment is skipped when not pushing, pushing to a branch that isn't `master`, or if a SSH deploy key hasn't been configured.

ISSOtm:
Since this has a custom PDF export (via the browser), can we consider #57 fixed by this?

ISSOtm:
Everything is done besides the glossary markup, but that still has to be added separately, so there's no point in blocking the rest of the improvements. I'm waiting on @mid-kid's discussion on #18, and any other feedback that may surface.

If no opposition is brought up, then I will push [the changes](https://github.com/iSSOtm/pandocs/) to this branch and merge them.

avivace:
Great work @ISSOtm . This is gorgeous.

The Glossary markup is not a priority for the migration PR. It wasn't there before, it's not necessary to have it now before migrating, it could be secondary task after we migrate to the new rust-based pipeline.

avivace:
> Since this has a custom PDF export (via the browser), can we consider #57 fixed by this?

Nope, I think we already brought that up but we could've done that before too and the PDF exported by a print from the browser does not provide a decent table of contents, etc so that issue still stands.

ISSOtm:
> > Since this has a custom PDF export (via the browser), can we consider #57 fixed by this?
> 
> Nope, I think we already brought that up but we could've done that before too and the PDF exported by a print from the browser does not provide a decent table of contents, etc so that issue still stands.

Well, it's a hack, but since we have control over the renderer, so I can either edit `print.html` in-place, or generate an alternative version with a ToC.
Alternatively, I'll see if there is a PDF renderer for mdBook, or one to a format that can be easily converted to PDF. I've seen ePub, for example, but I don't know.

By the way, since we have a ToC for navigating to a page, and the pages themselves are small enough (except "SGB Functions", which probably needs to be split), should we consider #60 fixed?

ISSOtm:
[mdbook-latex](https://crates.io/crates/mdbook-latex) is worth looking into; if we can't get it to work, its README links to other projects that look potentially useful.

avivace:
> > > Since this has a custom PDF export (via the browser), can we consider #57 fixed by this?
> > 
> > 
> > Nope, I think we already brought that up but we could've done that before too and the PDF exported by a print from the browser does not provide a decent table of contents, etc so that issue still stands.
> 
> Well, it's a hack, but since we have control over the renderer, so I can either edit `print.html` in-place, or generate an alternative version with a ToC.
> Alternatively, I'll see if there is a PDF renderer for mdBook, or one to a format that can be easily converted to PDF. I've seen ePub, for example, but I don't know.
> 
> By the way, since we have a ToC for navigating to a page, and the pages themselves are small enough (except "SGB Functions", which probably needs to be split), should we consider #60 fixed?

I agree on closing #60. For the PDF export, I still want to consider merging everything into a single MD/latex file and the render an actual PDF (and HTML, for the single page export) from there.

ISSOtm:
For that, I think the renderer I just linked could do a good job. We should look into it after merging this PR to avoid scope creep, though.

PR #298, opened 2021-05-27
Fix typo about CAM Registers bank enable bit and specify usual bank value in hex

Lycoder:
Changed "bit 5" to "bit 4" (seems someone had mismatched bit counting?), also specify the usual bank number in hex too, for consistency with the paragraph above


avivace (approved):

avivace:
Upstream source is @AntonioND

PR #303, opened 2021-06-21
Fix example instructions for building highlight.js

drglove:
highlight.js looks in the extra directory for 3rd party languages, not extras.


ISSOtm (approved):
Confirmed on my local copy that this is indeed a typo.

avivace:
Thanks @drglove !

PR #309, opened 2021-07-06
Add description of SGB detection using initial register values

martendo:
The values of $13 and $14 in the C register come from the [Mooney GB tests](https://github.com/Gekkio/mooneye-gb/tree/master/tests/acceptance) (boot_regs-*.s), double-checked with the boot ROMs at [gbdev.gg8.se](https://gbdev.gg8.se/files/roms/bootroms/).

PR #316, opened 2021-07-13
Split the SGB documentation into chapters

daid:
I've split the SGB single "massive page" into multiple chapters. Should make it easier to browse and look up things.

I did not change a single letter in the contents itself, only split it. Chapters are now setup as:
image

Note that this will conflict with #309


ISSOtm (approved):
At a glance, the split appears reasonable. May need tweaking later on, but this is a very good step in the right direction :)

daid:
> At a glance, the split appears reasonable. May need tweaking later on, but this is a very good step in the right direction :)

There is a lot more room for improvement. Like many of the "tables" in the SGB section are just code blocks would be a simple cleanup for someone to do as well.

ISSOtm:
We already have #87 for that one, but it can (should) be tackled in a separate PR.

avivace (approved):

PR #321, opened 2021-07-15
Update Gameboy_Printer.md

avivace:


ISSOtm:
Yeah, I had tried that, but it doesn't work; the server redirects HTTP to HTTPS, the latter causing the issue. :/

avivace:
We can just drop the link and put the full name of the company, tbf

ISSOtm:
Agreed.

Issue #318, opened 2021-07-15
Styling for "bit explanation" tables/diagrams


TheThief:
In this discussion: https://github.com/gbdev/pandocs/pull/310#discussion_r670237566

It came up that the current standard for explaining the function of individual bits in a control register is less than readable and/or is ugly.

Lets decide on something better.

Using this section as an example: https://gbdev.io/pandocs/STAT.html#ff41---stat-lcd-status-rw

**Old Pandocs style**:
(ugly)

```
Bit 6 - LYC=LY STAT Interrupt source         (1=Enable) (Read/Write)
Bit 5 - Mode 2 OAM STAT Interrupt source     (1=Enable) (Read/Write)
Bit 4 - Mode 1 VBlank STAT Interrupt source  (1=Enable) (Read/Write)
Bit 3 - Mode 0 HBlank STAT Interrupt source  (1=Enable) (Read/Write)
Bit 2 - LYC=LY Flag                          (0=Different, 1=Equal) (Read Only)
Bit 1-0 - Mode Flag                          (Mode 0-3, see below) (Read Only)
          0: In HBlank
          1: In VBlank
          2: Searching OAM
          3: Transferring Data to LCD Controller
```

**Current style guide (same thing but as a table)**:
(wall of text)

Bits | Name/Function | Value | Read/Write
-----|---------------|-------|-----------
Bit 6 | LYC=LY STAT Interrupt source | 1=Enable | Read/Write
Bit 5 | Mode 2 OAM STAT Interrupt source | 1=Enable | Read/Write
Bit 4 | Mode 1 VBlank STAT Interrupt source | 1=Enable | Read/Write
Bit 3 | Mode 0 HBlank STAT Interrupt source | 1=Enable | Read/Write
Bit 2 | LYC=LY Flag                          | 0=Different, 1=Equal | Read Only
Bit 1-0 | Mode Flag                         | Mode 0-3, see below | Read Only

Mode | Name/Function
-----|--------------
0 | In HBlank
1 | In VBlank
2 | Searching OAM
3 | Transferring Data to LCD Controller

**"Diagram" style (based on tauwasser's MMM01 docs)**:
https://wiki.tauwasser.eu/view/MMM01#RAM_Enable_Register

```
Bits: X 6 5 4 3 2 1 0
        | | | | | \_/
        | | | | |  \----- Bits 0-1: Current Mode                        (Read Only)
        | | | | \-------- Bit 2:    LYC=LY Flag                         (Read Only)
[...]

PR #327, opened 2021-07-20
Separate CGB Infrared section; clarify IR Receiver Adaptation

asiekierka:
This is based on notes from shonumi's Dan Docs, though a lot is omitted due to not being necessarily of use from a programmer's perspective or documenting the behaviour of specific games.

I've decided not to include the full detail on using RP as rudimentary light detection, but I'm fine with this being discussed further - I did include more detailed information on IR light adaptation as something to keep in mind while using the port for its intended purpose.

### DIFF
--- a/src/CGB_Infrared.md
+++ b/src/CGB_Infrared.md
@@ -1,0 +1,49 @@
+# CGB Infrared
+
+The GBC comes with an infrared port on the very top of the handheld,
+which can be used to send and receive IR signals via its two separate
+diodes. This functionality is not limited to other GBCs - it can be
+used to communicate with external devices as well.
+
+Note that the GBA does not include an infrared port and, as such, this
+method of communication is unavailable in its CGB compatibility mode.
+
+## FF56 - RP - Infrared Communications Port
+
+This register allows data to be transmitted and received through the
+infrared port.
+
+Bit | Name             | Usage notes                     | R/W?
+----|------------------|---------------------------------|-------------
+6-7 | Data read enable | 0=Disable, 3=Enable             | Read/Write
+ 1  | Read data        | 0=Receiving IR Signal, 1=Normal | Read Only
+ 0  | Write data       | 0=LED Off, 1=Led On             | Read/Write
+
+In order to read data, both Bit 6 and 7 must be set. Note that if Bit 0
+is not cleared, the CGB will end up reading its own IR signal.
+
+Both sending and receiving IR data increase power consumption - when not
+used, RP should be reset to 00h.
+
+## IR Receiver Adaptation
+
+The IR sensor in the GBC adapts to the current level of infrared light
+as a way to handle potential light pollution. In other words, if the
+GBC receives a sustained IR signal beyond a certain amount of time,
+eventually the sensor treats this as a new "normal" level of light,
+which causes Bit 1 of RP to go back to 1. This is called the signal
+"fade" because it may appear as if the signal disappears.
+
+The following should be kept in mind while working with the infrared
+port:
+
+* IR communication protocols should be designed with this effect in mind.
+  For example, a TV remote control may send a series of 32 LED ON/OFF
+  pulses (length 10us ON, 17.5us OFF each) instead of a permanent 880us
+  LED ON signal.
+* When enabling data read in the RP register under brighter light
+  conditions, one may initially read 0 from Bit 1, implying that an IR
+  signal is being received. As such, a short amount of time should pass
+  between writing 0xC0 to RP and attempting to read an IR signal. (TODO:
+  How much? The upper bound is known to be 3-4 ms; the lower bound
+  may be as little as 0.05-0.1 ms.)
--- a/src/CGB_Registers.md
+++ b/src/CGB_Registers.md
@@ -191,24 +191,7 @@ TODO: confirm whether interrupts can occur (just the joypad one?) during the pau
 ### FF56 - RP - CGB Mode Only - Infrared Communications Port
 
 This register allows to input and output data through the CGBs built-in
-Infrared Port. When reading data, bit 6 and 7 must be set (and obviously
-Bit 0 must be cleared - if you don't want to receive your own Game Boy's
-IR signal). After sending or receiving data you should reset the
-register to 00h to reduce battery power consumption again.
-
-```
- Bit 0:   Write Data   (0=LED Off, 1=LED On)             (Read/Write)
- Bit 1:   Read Data    (0=Receiving IR Signal, 1=Normal) (Read Only)
- Bit 6-7: Data Read Enable (0=Disable, 3=Enable)         (Read/Write)
-```
-
-Note that the receiver will adapt itself to the normal level of IR
-pollution in the air, so if you would send a LED ON signal for a longer
-period, then the receiver would treat that as normal (=OFF) after a
-while. For example, a Philips TV Remote Control sends a series of 32 LED
-ON/OFF pulses (length 10us ON, 17.5us OFF each) instead of a permanent
-880us LED ON signal. Even though being generally CGB compatible, the GBA
-does not include an infra-red port.
+Infrared Port. For more information, see [CGB Infrared](./CGB_Infrared.md).
 
 ### FF6C - OPRI - CGB Mode Only - Object Priority Mode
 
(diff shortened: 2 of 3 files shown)


src/CGB_Infrared.md
```diff
@@ -0,0 +1,49 @@
...
+----|------------------|---------------------------------|-------------
+6-7 | Data read enable | 0=Disable, 3=Enable             | Read/Write
+ 1  | Read data        | 0=Receiving IR Signal, 1=Normal | Read Only
+ 0  | Write data       | 0=LED Off, 1=Led On             | Read/Write
+
+In order to read data, both Bit 6 and 7 must be set. Note that if Bit 0
+is not cleared, the CGB will end up reading its own IR signal.
+
+Both sending and receiving IR data increase power consumption - when not
+used, RP should be reset to 00h.
```
ISSOtm, line 26:
```suggestion
used, RP should be reset to $00.
```

src/CGB_Infrared.md
```diff
@@ -0,0 +1,49 @@
+# CGB Infrared
+
+The GBC comes with an infrared port on the very top of the handheld,
+which can be used to send and receive IR signals via its two separate
+diodes. This functionality is not limited to other GBCs - it can be
+used to communicate with external devices as well.
```
ISSOtm, line 6:
```suggestion
used to communicate with other kinds of devices as well.
```
Maybe?

src/CGB_Infrared.md
```diff
@@ -0,0 +1,49 @@
...
+
+Note that the GBA does not include an infrared port and, as such, this
+method of communication is unavailable in its CGB compatibility mode.
+
+## FF56 - RP - Infrared Communications Port
+
+This register allows data to be transmitted and received through the
+infrared port.
+
+Bit | Name             | Usage notes                     | R/W?
```
ISSOtm, line 16: The format of these is being discussed in #318, feel free to weigh in there if this is the style you prefer (maybe not, since it looks like you may have sliced this section from the original docs verbatim).

In the meantime, I'd like this be kept consistent with the style used throughout the rest of the document ([such as here](https://gbdev.io/pandocs/STAT.html#ff41---stat-lcd-status-rw)).

src/CGB_Infrared.md
```diff
@@ -0,0 +1,49 @@
...
+----|------------------|---------------------------------|-------------
+6-7 | Data read enable | 0=Disable, 3=Enable             | Read/Write
+ 1  | Read data        | 0=Receiving IR Signal, 1=Normal | Read Only
+ 0  | Write data       | 0=LED Off, 1=Led On             | Read/Write
+
+In order to read data, both Bit 6 and 7 must be set. Note that if Bit 0
+is not cleared, the CGB will end up reading its own IR signal.
```
ISSOtm, line 23:
```suggestion
is also set, the CGB will end up reading its own IR signal.
```

src/CGB_Infrared.md
```diff
@@ -0,0 +1,49 @@
...
+GBC receives a sustained IR signal beyond a certain amount of time,
+eventually the sensor treats this as a new "normal" level of light,
+which causes Bit 1 of RP to go back to 1. This is called the signal
+"fade" because it may appear as if the signal disappears.
+
+The following should be kept in mind while working with the infrared
+port:
+
+* IR communication protocols should be designed with this effect in mind.
+  For example, a TV remote control may send a series of 32 LED ON/OFF
```
ISSOtm, line 41:
```suggestion
  For example, some TV remotes send a series of 32 LED ON/OFF
```
(Note sure if that's *actually* true, but I think it doesn't matter much while making the wording less confusing.)

src/CGB_Registers.md
```diff
@@ -191,24 +191,7 @@ TODO: confirm whether interrupts can occur (just the joypad one?) during the pau
 ### FF56 - RP - CGB Mode Only - Infrared Communications Port
```
ISSOtm, line 191: Since other CGB-only regs are mentioned in other pages only (such as palettes), this section should be done away with entirely.

src/SUMMARY.md
```diff
@@ -32,6 +32,7 @@
 - [Interrupts](./Interrupts.md)
   - [Interrupt Sources](./Interrupt_Sources.md)
 - [CGB Registers](./CGB_Registers.md)
+  - [CGB Infrared](./CGB_Infrared.md)
```
ISSOtm, line 35: This is a largely reasonable place, but what about grouping it with Serial instead?

ISSOtm (changes requested):
Thanks!

src/CGB_Infrared.md
asiekierka: I've based it on the LCDC page, as I assumed this is the current "desired" style. The original docs's format differs.

src/SUMMARY.md
asiekierka: I think that'd only make sense if the CGB Registers document were to be targetted for splitting into respective sections in general (particularly HDMA and OPRI).

src/CGB_Infrared.md
ISSOtm: Yeah, we know the style is not used consistently, hence that issue. :/

src/CGB_Registers.md
asiekierka: Might be good to open an issue for this then and do away with this section entirely - except there *are* some things which I'm not sure belong elsewhere (WRAM banking, speed switching).

src/SUMMARY.md
ISSOtm: Agreed on splitting HDMA, but OPRI, KEY1 (and KEY0 which was mysteriously omitted?) are definitely related to the CGB itself, imo, so keeping them grouped in this document makes sense.

src/CGB_Infrared.md
ISSOtm: (By the way, I'm picking this one specifically as I believe it's the best compromise until the new syntax is implemented.)

src/CGB_Infrared.md
aaaaaa123456789: The wording here ("not cleared") implies that, for proper usage, you _should_ clear the bit. This is correct, so maybe the stronger wording is helpful.

ISSOtm: It's possible to write once to enable reading, and once again to disable the emitter, for example. So maybe "while bit 0 is also set, ..."?

ISSOtm:
Bump?

avivace:
> Bump?

@asiekierka do you have some time to work on this?

src/CGB_Infrared.md
```diff
@@ -0,0 +1,49 @@
...
+Bit | Name             | Usage notes                     | R/W?
+----|------------------|---------------------------------|-------------
+6-7 | Data read enable | 0=Disable, 3=Enable             | Read/Write
+ 1  | Read data        | 0=Receiving IR Signal, 1=Normal | Read Only
+ 0  | Write data       | 0=LED Off, 1=Led On             | Read/Write
+
+In order to read data, both Bit 6 and 7 must be set. Note that if Bit 0
+is not cleared, the CGB will end up reading its own IR signal.
+
+Both sending and receiving IR data increase power consumption - when not
```
avivace, line 25:
```suggestion
Since sending and receiving IR data increase power consumption - when not
```

src/CGB_Infrared.md
```diff
@@ -0,0 +1,49 @@
...
+port:
+
+* IR communication protocols should be designed with this effect in mind.
+  For example, a TV remote control may send a series of 32 LED ON/OFF
+  pulses (length 10us ON, 17.5us OFF each) instead of a permanent 880us
+  LED ON signal.
+* When enabling data read in the RP register under brighter light
+  conditions, one may initially read 0 from Bit 1, implying that an IR
+  signal is being received. As such, a short amount of time should pass
+  between writing 0xC0 to RP and attempting to read an IR signal. (TODO:
```
avivace, line 47:
```suggestion
  between writing $C0 to RP and attempting to read an IR signal. (TODO:
```

See https://github.com/gbdev/pandocs/wiki/Document-Style

avivace (changes requested):
A part from some little inconsistencies and possible improvements, this looks very good and mature to me.

This is anyway blocked by https://github.com/gbdev/pandocs/issues/318.

avivace:
Superseded by #366 

This PR got completely stale and I'm quite against having to maintain an additional piece of documentation when the Dan Docs are quite on-par with what we're trying to do here.

@asiekierka (or anyone else) if there's something worth porting over from this, feel free to open additional PRs.

Issue #16, opened 2020-02-12
Dan Docs


avivace:
[Dan Docs](https://shonumi.github.io/dandocs.html) documents more obscure Game Boy hardware.
It is rendered from here https://github.com/shonumi/shonumi.github.io/blob/master/dandocs.html, which in turns comes from GBE docs : https://github.com/shonumi/gbe-plus/blob/master/src/docs/technical/dandocs.html, manually converted to HTML from the TXT files in the same folder, like this one: https://github.com/shonumi/gbe-plus/blob/master/src/docs/technical/Mobile_Adapter_GB.txt

We could script something to automatically render those TXT as MD and then have our new Pan Docs website render an updated (an in sync with Shonumi/GBE docs) version of Dan Docs.

asiekierka:
I'm not sure that's the best approach - the Dan Docs seem to rely to a large extent on hand-written ASCII tables, as well as having some sections which overlap with existing sections in the Pan Docs (GBC Infrared). I would also refrain from the disassemblies and providing fragments of data (such as barcode IDs) used by games - focusing solely on the programming aspect of each hardware.

shonumi:
I should have commented on this a long time ago, I suppose. Apologies for the lateness. No one's showed any interest in merging Pan Docs and Dan Docs, so I'd forgotten all about this open issue altogether.

Dan Docs fortunately no longer contains the Zok Zok Heroes disassembly. It's been replaced with C-like psuedo code. I have no real opinion whether said psuedo code fits anywhere into Pan Docs or not, but it should be mentioned that the copyright concerns there have hopefully be taken care of now. That was a big worry of mine.

The IR section contains some useful programming information not currently mentioned in Pan Docs (obscure behavior with the RP register, more details on the signal fading). Rather than copying a dedicated IR section over from Dan Docs, the existing Pan Docs section could be updated with missing information in this case.

On that note, copy+pasting all of Dan Docs is probably not ideal to integrate it with Pan Docs. Each item/accessory would probably have to be rewritten to better conform with the tone and scope of the rest of Pan Docs, rewording and reworking it so everything feels and reads cohesively. It's a large task, but unless that approach is done, it's just two separate documents stuck together. Their primary goals are different after all (Pan Docs targets homebrew development, Dan Docs aims to incorporate as much data as possible). Despite the similar naming scheme, Dan Docs is modeled after GBATEK as opposed to Pan Docs.  

[...]

PR #328, opened 2021-07-20
Add 4-Player Adapter documentation, adapted from Dan Docs.

asiekierka:


ISSOtm:
I'm not sure about whether this is in the scope of Pan Docs, given that this accessory is quite rare. Discussion on that in #16, though.

src/Four_Player_Adapter.md
```diff
@@ -0,0 +1,197 @@
...
+transfer data (1KB/s), while setting it to 0xFF runs it close to the slowest
+speed (2042 bits-per-second).
+
+SIZE sets the length of packets exchanged between all Game Boys. Nothing fancy,
+just the number of bytes in each packet. It probably shouldn't be set to zero.
+
+### Transmission Phase
+
+When the master Game Boy (Player 1) is ready, it should send 4 bytes
+(`0xAA 0xAA 0xAA 0xAA`). (TODO: Are 4 bytes actually required?) This alerts
```
avivace, line 119: I'm not sure TODOs should be scattered around like this

src/Four_Player_Adapter.md
```diff
@@ -0,0 +1,197 @@
...
+with garbage data that is based on output the master Game Boy had sent during
+the ping phase. At this time, it is recommended to ignore the earliest packets
+received, however, it is safe to start putting new, relevant data into the
+buffer.
+
+### Restarting Ping Phase
+
+It's possible to restart the ping phase while operating in the transmission
+phase. To do so, the master Game Boy should send 4 or more bytes
+(`0xFF 0xFF 0xFF 0xFF`). (TODO: It's possible fewer 0xFF bytes need to be sent,
```
avivace, line 193: same as previous comment

avivace (changes requested):
LGTM and it definitely fits the scope of the accessories chapter of Pan Docs. My only worry here is how to keep up with upstream, since this adds an additional burden to sync with Dan Docs if it's a subject actively being investigated.

src/Four_Player_Adapter.md
asiekierka: Fair; what would you propose instead?

avivace:
As we didn't get any constructive progress in the linked issues, I'm not gonna hold this any longer and merge it, I'll just fix the "todos"

avivace:
TODOs removed from the text and added as Issues

PR #330, opened 2021-07-20
Redraw PPU timings SVG, can be affected by theme

ISSOtm:
Fixes one item from #322

PR #337, opened 2021-07-25
Fix image link in MBC5 page

QuinnPainter:
More specifically, the "Mild Rumble" chart near the bottom of the page.

### DIFF
--- a/src/MBC5.md
+++ b/src/MBC5.md
@@ -58,6 +58,6 @@ bit to 1 enables the rumble motor and keeps it enabled until the bit is reset ag
 To control the rumble's intensity, it should be turned on and off repeatedly,
 as seen with these two examples from Pokémon Pinball:
 
-{{#include imgs/MBC5_Rumble_Mild.svg:2:}}
+<img src="imgs/MBC5_Rumble_Mild.svg" width="950px">
 
 <img src="imgs/MBC5_Rumble_Strong.svg" width="950px">


avivace (approved):

ISSOtm:
Uh, there are three problems here.

1. Why was this committed? This is supposed to be part of #330
2. Why was this reverted? What was wrong with it?
3. Why is the width forced to a certain *pixel* size? This won't scale properly on smaller viewports.

avivace:
> Uh, there are three problems here.
> 
>     1. Why was this committed? This is supposed to be part of [Allow SVGs to be affected by theme #330](https://github.com/gbdev/pandocs/pull/330)
> 
>     2. Why was this reverted? What was wrong with it?
> 
>     3. Why is the width forced to a certain _pixel_ size? This won't scale properly on smaller viewports.

The image was literally broken on the rendered version of Pan Docs, resulting in the raw text just being displayed there randomly.  #330 is in Draft phase and *in any case* it's not a definitive solution for SVGs, but just another way to write them to allow to be styled. 

As usual, a sub-optimal but acceptable solution is good and welcome while a better one is in the works.

ISSOtm:
> The image was literally broken on the rendered version of Pan Docs, resulting in the raw text just being displayed there randomly.

Oh, you mean the `{{#include ...}}` text? Then yeah, that's normal. I was fearing that it was broken in a way i hadn't noticed in #330, but this looks like the problem I encountered there (which, for the record, is simply that the SVGs are generated during *rendering*, but `#include` occurs during *preprocessing*).

Then, sorry for pushing that! :bow:

PR #338, opened 2021-07-27
Reformat Specifications table

### LINKED ISSUE #332: Rearrange Specifications table
- [x] Use `colspan` for information true across several models
- [x] Switch SGB and CGB columns

ISSOtm:
Fixes #332

### DIFF
--- a/src/Specifications.md
+++ b/src/Specifications.md
@@ -41,7 +41,7 @@ td:first-child {
         <td>Palettes</td><td colspan="2">BG: 1&nbsp;×&nbsp;4, OBJ: 2&nbsp;×&nbsp;3</td><td>BG/OBJ: 1&nbsp;+&nbsp;4&nbsp;×&nbsp;3, border: 4&nbsp;×&nbsp;15</td><td>BG: 8&nbsp;×&nbsp;4, OBJ: 8&nbsp;×&nbsp;3<sup class="footnote-reference"><a href="#compat">2</a></sup></td>
     </tr>
     <tr>
-        <td>Colors</td><td>4 shades of green</td><td>4 shades of gray</td><td colspan="2">32768 colors (RGB555)</td>
+        <td>Colors</td><td>4 shades of green</td><td>4 shades of gray</td><td colspan="2">32768 colors (15-bit RGB)</td>
     </tr>
     <tr>
         <td>Horizontal sync</td><td colspan="2">9.198&nbsp;KHz</td><td>Complicated<sup class="footnote-reference"><a href="#sgb_vid">3</a></sup></td><td>9.198&nbsp;KHz</td>
--- a/src/Specifications.md
+++ b/src/Specifications.md
@@ -20,7 +20,7 @@ td:first-child {
       <td>CPU</td><td colspan="4">8-bit 8080-like Sharp CPU (speculated to be a SM83 core)</td>
     </tr>
     <tr>
-      <td>CPU freq</td><td colspan="2">4.194304&nbsp;MHz</td><td>Depends on model<sup class="footnote-reference"><a href="#sgb_clk">1</a></sup></td><td>Up to 8.388608&nbsp;MHz</td>
+      <td>CPU freq</td><td colspan="2">4.194304&nbsp;MHz</td><td>Depends on revision<sup class="footnote-reference"><a href="#sgb_clk">1</a></sup></td><td>Up to 8.388608&nbsp;MHz</td>
     </tr>
     <tr>
         <td>Work RAM</td><td colspan="3">8&nbsp;KiB</td><td>32&nbsp;KiB<sup class="footnote-reference"><a href="#compat">2</a></sup> (4&nbsp;+&nbsp;7&nbsp;×&nbsp;4&nbsp;KiB)</td>
--- a/src/Specifications.md
+++ b/src/Specifications.md
@@ -1,17 +1,71 @@
 # Specifications
 
- &#xfeff;       | Game Boy                       | Game Boy Pocket | Game Boy Color  | Super Game Boy |
-----------------|--------------------------------|-----------------|-----------------|----------------|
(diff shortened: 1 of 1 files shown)


src/Specifications.md
```diff
@@ -1,17 +1,71 @@
...
+    <tr>
+      <th></th><th>Game Boy (DMG)</th><th>Game Boy Pocket (MGB)</th><th>Super Game Boy (SGB)</th><th>Game Boy Color (CGB)</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>CPU</td><td colspan="4">8-bit 8080-like Sharp CPU (speculated to be a SM83 core)</td>
+    </tr>
+    <tr>
+      <td>CPU freq</td><td colspan="2">4.194304&nbsp;MHz</td><td>Depends on model<sup class="footnote-reference"><a href="#sgb_clk">1</a></sup></td><td>Up to 8.388608&nbsp;MHz</td>
```
avivace, line 23: Isn't Super Game Boy (SGB) the model? Are you referring to "revision"?

avivace (changes requested):

src/Specifications.md
ISSOtm: Yep, right.

src/Specifications.md
```diff
@@ -1,17 +1,71 @@
...
+        <td>Resolution</td><td colspan="2">160&nbsp;×&nbsp;144</td><td>160&nbsp;×&nbsp;144 within 256&nbsp;×&nbsp;224 border</td><td>160&nbsp;×&nbsp;144</td>
+    </tr>
+    <tr>
+        <td>OBJ ("sprites")</td><td colspan="4">8&nbsp;×&nbsp;8 or 8&nbsp;×16 ; max 40 per screen, 10 per line</td>
+    </tr>
+    <tr>
+        <td>Palettes</td><td colspan="2">BG: 1&nbsp;×&nbsp;4, OBJ: 2&nbsp;×&nbsp;3</td><td>BG/OBJ: 1&nbsp;+&nbsp;4&nbsp;×&nbsp;3, border: 4&nbsp;×&nbsp;15</td><td>BG: 8&nbsp;×&nbsp;4, OBJ: 8&nbsp;×&nbsp;3<sup class="footnote-reference"><a href="#compat">2</a></sup></td>
+    </tr>
+    <tr>
+        <td>Colors</td><td>4 shades of green</td><td>4 shades of gray</td><td colspan="2">32768 colors (RGB555)</td>
```
avivace, line 44: Why was RGB555 added here? I don't feel it's relevant - it's a "format", but from a specification point of view 32768 colors with 1/3 of the bits per color is pretty standard and 32768 colors on RGB gives already this information

avivace (changes requested):

src/Specifications.md
Rangi42: I think "RGB555" is most precise compared to "32768 colors", "15-bit color", "high color", etc. It clearly says that all channels are equal 5-bit, not a 15-bit palette index or something weird like RGB465.

ISSOtm: Yes, this was my line of thinking.

avivace: I don't think is most precise. On the contrary it's confusing because specifying a particular palette or encoding implies you are trying to communicate something non-standard, while actually, by standard (if not otherwise stated), every channel is encoded by the same amount of information: 32768 colors need 2^15 bits, I don't see why I should think each channel is not getting encoded by (15/3) bits anyway

avivace, line 44:
```suggestion
        <td>Colors</td><td>4 shades of green</td><td>4 shades of gray</td><td colspan="2">32768 colors (RGB)</td>
```

Rangi42: There are some pretty unusual color encodings in https://en.wikipedia.org/wiki/List_of_video_game_console_palettes, including for the NES, which don't just divide the bits into equal RGB (nor even YUV, HSL, etc) components.

avivace: The fact you define them "unusual" proves my point. You mention the encoding when it's unusual and it what way, not when it's standard.
In the exact same page you linked, under "Super NES", you can read "15-bit RGB (32,768 color)" for what we are discussing here.

Rangi42: "32768 colors (15-bit)" then?

avivace: You are suggesting something completely unrelated to what we are discussing now. ISSO slipt information about how the channels are encoded, you now want to enforce (re-explain) that to encode 32768 colors you need 15 bits

avivace: Which for me is again wrong. The wikipedia page is not a manual and the target is different. This is an introductory specification table aimed to give an *overview* in a technical document.
Why are you feeling additional information is needed? What information you can't get from "32768 colors (RGB)"?

Rangi42: What I meant is, if you're against the extra information, just go ahead and say "32,768 colors" because "(RGB)" doesn't clarify anything. I was suggesting "15-bit" or "15-bit RGB" as alternatives to "RGB555"; IMO any of those would clarify *how* those 32,768 colors are determined.

ISSOtm: Imo "32768 RGB colors" without further precision strikes a good deal between disambiguation ("it's RGB, not YUV or a fixed palette") and terseness ("32768 → 15-bit, if it's RGB and no precisions are given, it's likely RGB555").

avivace: I've been talked in how "unusual" this can result and how it can be useful anyway to specify RGB555 so `32,768 colors (15 bit RGB)` is okay for me but if you really think this should go on a introductory specification table I will accept `32,768 (RGB555)`, too

ISSOtm: I think "15-bit RGB" strikes the best balance we all agree on, so let's use that. Thanks!

PR #341, opened 2021-07-29
Apply misc. graphics-related fixups

ISSOtm:
Up for debate, of course.

### DIFF
--- a/src/Palettes.md
+++ b/src/Palettes.md
@@ -22,72 +22,82 @@ Value | Color
   2   | Dark gray
   3   | Black
 
-In CGB Mode the Color Palettes are taken from [CGB Palette Memory](<#LCD Color Palettes (CGB only)>)
+In CGB Mode the color palettes are taken from [CGB palette memory](<#LCD Color Palettes (CGB only)>)
 instead.
 
-### FF48 - OBP0 (Object Palette 0 Data) (R/W) - Non CGB Mode Only
+### FF48 - OBP0 (OBJ Palette 0 Data) (R/W), FF49 - OBP1 (OBJ Palette 1 Data) (R/W) - Both Non CGB Mode Only
 
-This register assigns gray shades to the color indexes of the OBJs that use this palette. It works exactly
-like BGP (FF47), except that the lower two bits are ignored because
-sprite index 00 means transparent.
-
-### FF49 - OBP1 (Object Palette 1 Data) (R/W) - Non CGB Mode Only
-
-This register assigns gray shades to the color indexes of the OBJs that use this palette. It works exactly
-like BGP (FF47), except that the lower two bits are ignored because
-sprite index 00 means transparent.
+These registers assigns gray shades to the color indexes of the OBJs that use the corresponding palette.
+They work exactly like BGP, except that the lower two bits are ignored because color index 0 is transparent for OBJs.
 
 ## LCD Color Palettes (CGB only)
 
+The CGB has a small amount of RAM used to store its color palettes. Unlike most
+of the hardware interface, palette RAM (or *CRAM* for *Color RAM*) is not
+accessed directly, but instead through the following registers:
+
 ### FF68 - BCPS/BGPI (Background Color Palette Specification or Background Palette Index) - CGB Mode Only
 
-This register is used to address a byte in the CGBs Background Palette
-Memory. Each two byte in that memory define a color value. The first 8
-bytes define Color 0-3 of Palette 0 (BGP0), and so on for BGP1-7.
+This register is used to address a byte in the CGB's background palette RAM.
+Since there are 8 palettes, 8 palettes × 4 colors/palette × 2 bytes/color = 64 bytes
+can be addressed.
 
 ```
 Bit 7     Auto Increment  (0=Disabled, 1=Increment after Writing)
-Bit 5-0   Index (00-3F)
+Bit 5-0   Address ($00-3F)
 ```
 
-Data can be read/written to/from the specified index address through
-Register FF69. When the Auto Increment bit is set then the index is
-automatically incremented after each **write** to FF69. Auto Increment has
-no effect when **reading** from FF69, so the index must be manually
-incremented in that case. Writing to FF69 during rendering still causes
-auto-increment to occur.
+First comes BGP0 color number 0, then BGP0 color number 1, BGP0 color number 2, BGP0 color number 3,
+BGP1 color number 0, and so on. Thus, address $03 allows accessing the second (upper)
+byte of BGP0 color #1 via BCPD, which contains the color's blue and upper green bits.
 
-Unlike the following, this register can be accessed outside VBlank and
-HBlank.
+Data can be read from or written to the specified CRAM address through
+BCPD/BGPD. If the Auto Increment bit is set, the index gets
+incremented after each **write** to BCPD. Auto Increment has
+no effect when **reading** from BCPD, so the index must be manually
+incremented in that case. Writing to BCPD during rendering still causes
+auto-increment to occur, despite the write being blocked.
+
+Unlike BCPD, this register can be accessed outside VBlank and HBlank.
 
 ### FF69 - BCPD/BGPD (Background Color Palette Data or Background Palette Data) - CGB Mode Only
 
-This register allows to read/write data to the CGBs Background Palette
-Memory, addressed through Register FF68. Each color is defined by two
-bytes (Bit 0-7 in first byte).
+This register allows to read/write data to the CGBs background palette memory,
+addressed through BCPS/BGPI. Each color is stored as little-endian RGB555:
 
 ```
-Bit 0-4   Red Intensity   (00-1F)
-Bit 5-9   Green Intensity (00-1F)
-Bit 10-14 Blue Intensity  (00-1F)
+Bit 0-4   Red Intensity   ($00-1F)
+Bit 5-9   Green Intensity ($00-1F)
+Bit 10-14 Blue Intensity  ($00-1F)
 ```
 
-Much like VRAM, data in Palette Memory cannot be read/written during the
-time when the LCD Controller is reading from it. (That is when the STAT
-register indicates Mode 3). Note: All background colors are initialized
-as white by the boot ROM, but it's a good idea to initialize at least
-one color yourself (for example if you include a soft-reset mechanic).
+Much like VRAM, data in palette memory cannot be read or written during the time
+when the PPU is reading from it, that is, [Mode 3](<#LCD Status Register>).
 
-### FF6A - OCPS/OBPI (Object Color Palette Specification or Sprite Palette Index), FF6B - OCPD/OBPD (Object Color Palette Data or Sprite Palette Data) - Both CGB Mode Only
+::: tip NOTE
 
-These registers are used to initialize the Sprite Palettes OBP0-7,
-identically as described above for Background Palettes. Note that four
-colors may be defined for each OBP Palettes - but only Color 1-3 of each
-Sprite Palette can be displayed, Color 0 is always transparent, and can
-be initialized to a don't care value or plain never initialized.
+All background colors are initialized as white by the boot ROM, however it is a
+good idea to initialize all colors yourself, e.g. if implementing
+a soft-reset mechanic.
 
-Note: All sprite colors are left uninitialized by the boot ROM, and are
-somewhat random.
+:::
+
+### FF6A - OCPS/OBPI (OBJ Color Palette Specification / OBJ Palette Index), FF6B - OCPD/OBPD (OBJ Color Palette Data / OBJ Palette Data) - Both CGB Mode Only
+
+These registers function exactly like BCPS and BCPD respectively; the 64 bytes
+of OBJ palette memory are entirely separate from Background palette memory, but
+function the same.
+
+Note that while 4 colors are stored per OBJ palette, color #0 is never used, as
+it's always transparent. It's thus fine to write garbage values, or even leave
+color #0 uninitialized.
+
+::: tip NOTE
+
+The boot ROM leaves all object colors uninitialized (and thus somewhat random),
+aside from setting the first byte of OBJ0 color #0 to $00, which is unused.
+
+:::
 
 ### RGB Translation by CGBs
 
@@ -96,14 +106,14 @@ somewhat random.
 When developing graphics on PCs, note that the RGB values will have
 different appearance on CGB displays as on VGA/HDMI monitors calibrated
 to sRGB color. Because the GBC is not lit, the highest intensity will
-produce Light Gray color rather than White. The intensities are not
-linear; the values 10h-1Fh will all appear very bright, while medium and
-darker colors are ranged at 00h-0Fh.
+produce light gray rather than white. The intensities are not
+linear; the values $10-$1F will all appear very bright, while medium and
+darker colors are ranged at $00-0F.
 
 The CGB display's pigments aren't perfectly saturated. This means the
-colors mix quite oddly; increasing intensity of only one R,G,B color
-will also influence the other two R,G,B colors. For example, a color
-setting of 03EFh (Blue=0, Green=1Fh, Red=0Fh) will appear as Neon Green
+colors mix quite oddly: increasing the intensity of only one R/G/B color
+will also influence the other two R/G/B colors. For example, a color
+setting of $03EF (Blue=$00, Green=$1F, Red=$0F) will appear as Neon Green
 on VGA displays, but on the CGB it'll produce a decently washed out
 Yellow. See the image above.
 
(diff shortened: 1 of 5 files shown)


ISSOtm:
Bump

avivace:
> Bump

I'm on this

src/Palettes.md
```diff
@@ -22,72 +22,89 @@ Value | Color
...
+Bit 5-0   Address ($00-3F)
 ```
 
-Data can be read/written to/from the specified index address through
-Register FF69. When the Auto Increment bit is set then the index is
-automatically incremented after each **write** to FF69. Auto Increment has
-no effect when **reading** from FF69, so the index must be manually
-incremented in that case. Writing to FF69 during rendering still causes
-auto-increment to occur.
+First comes BGP0 color #0, then BGP0 color #1, BGP0 color #2, BGP0 color #3,
```
avivace, line 57: Why are we using `#` for "number" here? Is is something we do somewhere else ?

src/Palettes.md
```diff
@@ -22,72 +22,89 @@ Value | Color
...
+BGP1 color #0, and so on. Thus, address $03 allows accessing the second (upper)
+byte of BGP0 color #1 via BCPD, which contains the color's blue and upper green bits.
+
+Data can be read from / written to the specified CRAM address through
```
avivace, line 61:
```suggestion
Data can be read from or written to the specified CRAM address through
```

src/Palettes.md
```diff
@@ -22,72 +22,89 @@ Value | Color
...
+BCPD/BGPD. If the Auto Increment bit is set, the index is
+automatically incremented after each **write** to BCPD. Auto Increment has
```
avivace, line 63:
```suggestion
incremented after each **write** to BCPD. Auto Increment has
```

src/Palettes.md
```diff
@@ -22,72 +22,89 @@ Value | Color
...
-automatically incremented after each **write** to FF69. Auto Increment has
-no effect when **reading** from FF69, so the index must be manually
-incremented in that case. Writing to FF69 during rendering still causes
-auto-increment to occur.
+First comes BGP0 color #0, then BGP0 color #1, BGP0 color #2, BGP0 color #3,
+BGP1 color #0, and so on. Thus, address $03 allows accessing the second (upper)
+byte of BGP0 color #1 via BCPD, which contains the color's blue and upper green bits.
+
+Data can be read from / written to the specified CRAM address through
+BCPD/BGPD. If the Auto Increment bit is set, the index is
```
avivace, line 62:
```suggestion
BCPD/BGPD. If the Auto Increment bit is set, the index gets
```

src/Palettes.md
```diff
@@ -22,72 +22,89 @@ Value | Color
...
+Bit 5-9   Green Intensity ($00-1F)
+Bit 10-14 Blue Intensity  ($00-1F)
 ```
 
-Much like VRAM, data in Palette Memory cannot be read/written during the
-time when the LCD Controller is reading from it. (That is when the STAT
-register indicates Mode 3). Note: All background colors are initialized
-as white by the boot ROM, but it's a good idea to initialize at least
-one color yourself (for example if you include a soft-reset mechanic).
+Much like VRAM, data in palette memory cannot be read/written during the time
```
avivace, line 81:
```suggestion
Much like VRAM, data in palette memory cannot be read or written during the time
```

I'd limit the use of slashes. They are for informal writing (which our is not) and we already over use them

src/Palettes.md
```diff
@@ -22,72 +22,89 @@ Value | Color
...
+when the PPU is reading from it, that is, [Mode 3](<#LCD Status Register>).
+
+::: tip NOTE
+
+All background colors are initialized as white by the boot ROM, however it is a
+good idea to initialize all colors yourself, for example if you plan to include
```
avivace, line 87:
```suggestion
good idea to initialize all colors yourself, e.g. to implement
```

src/Palettes.md
```diff
@@ -22,72 +22,89 @@ Value | Color
...
+a soft-reset mechanic.
+
+:::
+
+### FF6A - OCPS/OBPI (OBJ Color Palette Specification or OBJ Palette Index), FF6B - OCPD/OBPD (OBJ Color Palette Data or OBJ Palette Data) - Both CGB Mode Only
+
+These registers function exactly like BCPS and BCPD respectively; the 64 bytes
+of OBJ palette memory are entirely separate from Background palette memory, but
+function just the same.
```
avivace, line 96:
```suggestion
function the same.
```

src/Palettes.md
```diff
@@ -96,14 +113,14 @@ somewhat random.
...
-darker colors are ranged at 00h-0Fh.
+produce light gray rather than white. The intensities are not
+linear; the values $10-$1F will all appear very bright, while medium and
+darker colors are ranged at $00-0F.
 
 The CGB display's pigments aren't perfectly saturated. This means the
-colors mix quite oddly; increasing intensity of only one R,G,B color
-will also influence the other two R,G,B colors. For example, a color
-setting of 03EFh (Blue=0, Green=1Fh, Red=0Fh) will appear as Neon Green
+colors mix quite oddly; increasing the intensity of only one R/G/B color
```
avivace, line 121:
```suggestion
colors mix quite oddly: increasing the intensity of only one R/G/B color
```

src/Palettes.md
```diff
@@ -113,18 +130,18 @@ Even though GBA is described to be compatible to CGB games, most CGB
 games are completely unplayable on older GBAs because most colors are
 invisible (black). Of course, colors such like Black and White will
 appear the same on both CGB and GBA, but medium intensities are arranged
-completely different. Intensities in range 00h..07h are invisible/black
+completely different. Intensities in range $00..07 are invisible/black
```
avivace, line 133: Don't we use `-` for ranges?

src/Palettes.md
```diff
@@ -113,18 +130,18 @@ Even though GBA is described to be compatible to CGB games, most CGB
...
 darker colors.
 
 Newer CGB games may avoid this effect by changing palette data when
-detecting GBA hardware ([see
-how](<#Detecting CGB (and GBA) functions>)).
-Based on measurement of GBC and GBA palettes using the [144p Test
-Suite](https://github.com/pinobatch/240p-test-mini/tree/master/gameboy) ROM, a fairly close approximation is GBA = GBC \* 3/4 + 8h for
-each R,G,B intensity. The result isn't quite perfect, and it may turn
+detecting GBA hardware ([see how](<#Detecting CGB (and GBA) functions>)).
+Based on measurements of GBC and GBA palettes using the
```
avivace, line 141: I think this should go in a TIP box

src/STAT.md
```diff
@@ -42,34 +42,34 @@ Mode 0  ___000___000___000___000___000___000________________000
 Mode 1  ____________________________________11111111111111_____
 ```
 
-When the PPU is reading a particular part of video memory,
-that memory is inaccessible to the CPU.
+When the PPU is accessing some video-related memory, that memory is inaccessible
+to the CPU: writes are ignored, and reads return garbage values (usually $FF).
```
avivace, line 46:
```suggestion
to the CPU: writes are ignored, and reads return garbage values (usually $FF).
```
Why is there a comma before and here?

src/STAT.md
```diff
@@ -42,34 +42,34 @@ Mode 0  ___000___000___000___000___000___000________________000
...
-  2  | Searching OAM for OBJs whose (X,Y) coordinates overlap this line | 80 dots (19 µs)                                                    | VRAM, CGB palettes
-  3  | Reading OAM and VRAM to generate the picture                     | 168 to 291 dots (40 to 60 µs) depending on sprite count            | None
-  0  | Horizontal blanking                                              | 85 to 208 dots (20 to 49 µs) depending on previous mode 3 duration | VRAM, OAM, CGB palettes
-  1  | Vertical blanking                                                | 4560 dots (1087 µs, 10 scanlines)                                  | VRAM, OAM, CGB palettes
+  2  | Searching OAM for OBJs whose Y coordinate overlap this line | 80 dots (19 µs)                                                    | VRAM, CGB palettes
+  3  | Reading OAM and VRAM to generate the picture                | 168 to 291 dots (40 to 60 µs) depending on sprite count            | None
+  0  | Nothing (horizontal blanking)                               | 85 to 208 dots (20 to 49 µs) depending on previous mode 3 duration | VRAM, OAM, CGB palettes
```
avivace, line 56:
```suggestion
  0  | Nothing (HBlank)                               | 85 to 208 dots (20 to 49 µs) depending on previous mode 3 duration | VRAM, OAM, CGB palettes
```

src/STAT.md
```diff
@@ -42,34 +42,34 @@ Mode 0  ___000___000___000___000___000___000________________000
...
+  1  | Nothing (vertical blanking)                                 | 4560 dots (1087 µs, 10 scanlines)                                  | VRAM, OAM, CGB palettes
```
avivace, line 57:
```suggestion
  1  | Nothing (VBlank)                                 | 4560 dots (1087 µs, 10 scanlines)                                  | VRAM, OAM, CGB palettes
```

src/STAT.md
```diff
@@ -42,34 +42,34 @@ Mode 0  ___000___000___000___000___000___000________________000
...
+- Background scrolling: If `SCX % 8` is not zero at the start of the scanline, rendering is paused for that many dots while the shifter discards that many pixels from the leftmost tile.
 - Window: An active window pauses for at least 6 dots, as the background fetching mechanism starts over at the left side of the window.
-- Sprites: Each sprite usually pauses for `11 - min(5, (x + SCX) mod 8)` dots. Because sprite fetch waits for background fetch to finish, a sprite's cost depends on its position relative to the left side of the background tile under it. It's greater if a sprite is directly aligned over the background tile, less if the sprite is to the right. If the sprite's left side is over the window, use `255 - WX` for `SCX` in this formula.
+- Sprites: Each sprite usually pauses for `11 - min(5, (x + SCX) % 8)` dots. Because sprite fetch waits for background fetch to finish, a sprite's cost depends on its position relative to the left side of the background tile under it. It's greater if a sprite is directly aligned over the background tile, less if the sprite is to the right. If the sprite's left side is over the window, use `255 - WX` for `SCX` in this formula.
```
avivace, line 72:
```suggestion
- Sprites: Each sprite usually pauses for `11 - min(5, (x + SCX) % 8)` dots. Because sprite fetch waits for background fetch to finish, a sprite's cost depends on its position relative to the left side of the background tile under it. It's greater if a sprite is directly aligned over the background tile, less if the sprite is to the right. If the sprite's left side is over the window, use `255 - WX` instead of `SCX` in this formula.
```

src/Scrolling.md
```diff
@@ -1,19 +1,19 @@
 
 # LCD Position and Scrolling
 
-These registers can be accessed even during Mode 3, but they have no
-effect until the end of the current scanline.
+These registers can be accessed even during Mode 3, but they may not take
```
avivace, line 4:
```suggestion
These registers can be accessed even during Mode 3, but they may not have 
```

src/Scrolling.md
```diff
@@ -1,19 +1,19 @@
...
+effect immediately.
```
avivace, line 5: I don't think the change was an improvement. Immediatly can mean anything in this context while the previous formulation had more information about until when they can (not) show any effect.

src/Scrolling.md
```diff
@@ -32,10 +32,18 @@ The Window is visible (if enabled) when both coordinates are in the ranges
...
-WX values 0-6 and 166 are unreliable due to hardware bugs. If WX is set
-to 0, the window will "stutter" horizontally when SCX changes.
-(Depending on SCX modulo 8, behavior is a little complicated so you
-should try it yourself.)
+::: warning
+
+WX values 0 and 166 are unreliable due to hardware bugs.
+
+If WX is set to 0, the window will "stutter" horizontally when SCX changes
+(depending on SCX modulo 8, behavior is a little complicated so you
```
avivace, line 40:
```suggestion
(depending on SCX % 8, behavior is a little complicated so you
```

avivace, line 40:
```suggestion
(depending on SCX modulo 8, detailed behaviour is unknown)
```

src/Scrolling.md
```diff
@@ -32,10 +32,18 @@ The Window is visible (if enabled) when both coordinates are in the ranges
...
+should try it yourself).
```
avivace, line 41:
```suggestion

```

avivace (changes requested):

src/Palettes.md
```diff
@@ -22,72 +22,89 @@ Value | Color
   2   | Dark gray
   3   | Black
 
-In CGB Mode the Color Palettes are taken from [CGB Palette Memory](<#LCD Color Palettes (CGB only)>)
+In CGB Mode the color palettes are taken from [CGB palette memory](<#LCD Color Palettes (CGB only)>)
 instead.
 
-### FF48 - OBP0 (Object Palette 0 Data) (R/W) - Non CGB Mode Only
+### FF48 - OBP0 (OBJ Palette 0 Data) (R/W) - Non CGB Mode Only
```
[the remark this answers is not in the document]
ISSOtm: Yep, [in the overview](https://gbdev.io/pandocs/Rendering.html#background).

src/Palettes.md
```diff
@@ -22,72 +22,89 @@ Value | Color
...
+this palette. It works exactly like BGP, except that the lower two bits are
+ignored because color index 0 is transparent for OBJs.
 
-### FF49 - OBP1 (Object Palette 1 Data) (R/W) - Non CGB Mode Only
+### FF49 - OBP1 (OBJ Palette 1 Data) (R/W) - Non CGB Mode Only
 
-This register assigns gray shades to the color indexes of the OBJs that use this palette. It works exactly
-like BGP (FF47), except that the lower two bits are ignored because
-sprite index 00 means transparent.
+This register assigns gray shades to the color indexes of the OBJs that use
```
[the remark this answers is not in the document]
ISSOtm: Yeah, the two sections should probably be merged.

src/Palettes.md
ISSOtm: No, according to some research, spelling out "number" or maybe `n°`?

src/Palettes.md
ISSOtm: I'mmmm... not entirely sure. I'd ask a native speaker for advice here.

src/Palettes.md
ISSOtm:
```suggestion
good idea to initialize all colors yourself, e.g. if implementing
```?

File: src/Palettes.md (hunk 9 above)
Comment (ISSOtm) [reply src/Palettes.md:133]: `–`, to be precise, but yes. Should probably go in the [style guide](https://github.com/gbdev/pandocs/wiki/Document-Style).

File: src/STAT.md (hunk 11 above)
Comment (ISSOtm) [reply src/STAT.md:46]: Because I found its lack makes the sentence harder to process.

File: src/STAT.md (hunk 12 above)
Comment (ISSOtm) [reply src/STAT.md:56]: I figured spelling out the name in full wouldn't hurt, since we have some room to spare.

File: src/Scrolling.md (hunk 16 above)
Comment (ISSOtm) [reply src/Scrolling.md:5]: The previous formulation was wrong, actually. The scroll registers are re-read on each tile fetch, except for the low 3 bits of `SCX`, which are only read at the beginning of the scanline (during the initial shifting).

I figured that explaining this nuance (which also opens the can of worms of "what is a tile fetch?") here would be detrimental to comprehension.

File: src/Scrolling.md (hunk 17 above)
Comment (ISSOtm) [reply src/Scrolling.md:40]:
```suggestion
(depending on SCX modulo 8, exact behavior is a little complicated).
```
It *is* properly understood, I believe.

src/Palettes.md
avivace: I agree

src/Scrolling.md
ISSOtm: Either "take effect" or "have an effect", AIUI. Also, `s/they/any modifications/`.

src/STAT.md
avivace: I don't think it's a question of space here, as in the entire document (after the definition we use HBlank and VBlank), one can think we are talking about something different if go back at this spelling in some tables..

ISSOtm: Fair enough. (And one more reason to #194, lol.)

src/Scrolling.md
avivace: Got it and I agree with you. Maybe put an info box with the tile fetch thing? It's interesting.

ISSOtm: Or a link to the FIFO article?

src/Scrolling.md
avivace: Then why is the reader invited to "try it himself because it's too complicated to explain"?

ISSOtm: I think because I couldn't find it documented anywhere (guess I could have taken a look at SameBoy's source?), and I thought that it wouldn't be worth explaining in full there. I was probably wrong with that, but then what shall we do?

avivace: We can link SameBoy source if that's implemented there.

src/Scrolling.md
avivace: I prefer adding a brief explanation like the one you gave me here and link to the FIFO article

src/Scrolling.md
avivace: Ok

src/Palettes.md
avivace: I feel like `#' can be used but it must be a document-wise thing. In the mean time, both of the following suggestions would work for me

src/Palettes.md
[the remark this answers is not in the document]
avivace: Agree

src/Scrolling.md
ISSOtm: Footnote it is, then.

src/Palettes.md
ISSOtm: I want to avoid the former because of potential ambiguity.

src/Scrolling.md
ISSOtm: I instead went and turned the "mid-frame" tip box into a section, where I also added this info.

src/Scrolling.md
ISSOtm: I'm not fond of that because AFAIK there is no test ROM for it.

avivace: If we don't have a source, we can't provide details then we shouldn't mention it at all. Just open an Issue and mark it as research. Makes no sense to add confusion here mentioning stuff in half.

ISSOtm: Then just
```suggestion
(depending on SCX modulo 8).
```
?

avivace: Agree!

avivace: (But still an Issue is worth opening)

avivace (approved):
LGTM! Thanks @ISSOtm

PR #343, opened 2021-08-02
Document halt bug

ISSOtm:
In its two forms, and with a test ROM as citation, too!

### DIFF
--- a/src/Interrupts.md
+++ b/src/Interrupts.md
@@ -22,8 +22,9 @@ RETI   ;Enables interrupts and returns (same as the instruction sequence EI, RET
 where \<INT\> means the operation which is automatically executed by the
 CPU when it executes an interrupt.
 
-The effect of EI is delayed by one instruction. This means that EI
+The effect of `ei` is delayed by one instruction. This means that `ei`
 followed immediately by DI does not allow any interrupts between them.
+This interacts with the [`halt` bug](<#halt bug>) in an interesting way.
 
 ## FFFF - IE - Interrupt Enable (R/W)
 
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -31,6 +31,7 @@
   - [Timer Obscure Behaviour](./Timer_Obscure_Behaviour.md)
 - [Interrupts](./Interrupts.md)
   - [Interrupt Sources](./Interrupt_Sources.md)
+  - [HALT](./halt.md)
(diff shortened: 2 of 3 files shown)


avivace (approved):

avivace:
Looks good to me but I'll wait for the PR on SameBoy to be merged.
Also, shouldn't we try to set up a repository with test ROMs?

CasualPokePlayer:
> 
> 
> Looks good to me but I'll wait for the PR on SameBoy to be merged.
> Also, shouldn't we try to set up a repository with test ROMs?

The PR is on SameBoy, it's on SameSuite (LIJI's testrom suite), and the PR has been sitting there for months now. SameBoy also has the same behavior demonstrated in the testrom already (along with other emulators such as BGB and Gambatte).

avivace:
> > Looks good to me but I'll wait for the PR on SameBoy to be merged.
> 
> The PR is on SameBoy, it's on SameSuite (LIJI's testrom suite), and the PR has been sitting there for months now. SameBoy also has the same behavior demonstrated in the testrom already (along with other emulators such as BGB and Gambatte).

Still, linking a file uploaded in a pull request as reference or source to a fact is not acceptable. Either wait for it to be merged or host it somewhere else. Source branches get deleted when merging.

CasualPokePlayer:
> Source branches get deleted when merging.
I'm not aware of any sort of automatic deletion of branches, only a deletion prompt which the user can ignore.

Even still, even if you delete the branch, [that would not delete the commit from GitHub](https://github.com/LIJI32/SameSuite/commit/ee31c156deb5bd96b76a8c47a48bff7a516c94cb), at least not without completely deleting the host repository/contacting GitHub support/(more effectively like a delete) privating the repo. You can just swap the link with the commit I've linked and you effectively have no issues.

ISSOtm:
beware mentions that *Torpedo Range* may trigger this; mentioning that fact would be cool, so this should be verified if possible.

ISSOtm:
[Liji just merged the PR](https://github.com/LIJI32/SameSuite/pull/11#event-5256354908) after some discussion, so I'll update this to refer to the new upstream.

They mentioned that they're planning to release another test ROM that would supersede that one, but we can just update the link when the weekly link check reports that it's down, or we can use a permalink (which I'm not too fond of, though).

avivace (approved):

avivace:
Great, thanks!

PR #344, opened 2021-08-05
Mark NR31 and NR41 as write only

ICanNotThinkOfAGoodUsername:
This is seen in both [SameBoy's APU read function](https://github.com/LIJI32/SameBoy/blob/master/Core/apu.c#L746-L787) and the dmg_sound test rom. Both use a mask of `$FF`, meaning it is impossible to read these registers and effectively making them write only.

### DIFF
--- a/src/Sound_Controller.md
+++ b/src/Sound_Controller.md
@@ -152,10 +152,10 @@ square wave. This channel doesn't have a volume envelope register.
 Bit 7 - Sound Channel 3 Off  (0=Stop, 1=Playback)  (Read/Write)
 ```
 
-### FF1B - NR31 - Channel 3 Sound Length (R/W)
+### FF1B - NR31 - Channel 3 Sound Length (W)
 
 ```
-Bit 7-0 - Sound length (t1: 0 - 255)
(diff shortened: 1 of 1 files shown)


ISSOtm:
I'm surprised that we'd be getting this wrong (although it wouldn't be the first time, HDMA1–4 suffered from the same issue), so I think we should actually test this on hardware. Because the test ROM is not checking if they are readable (AIUI) doesn't mean they aren't. So maybe those are wrong, and I'd like to double-check that.

Still, thanks for bringing this up! ^^

ICanNotThinkOfAGoodUsername:
I wrote a quick test rom that tries writing $00 and $FF to both locations and shows the value of a read immediately after. This is the output I get in Same Boy (mGBA is exactly the same).
Screenshot from 2021-08-11 22-17-06

I asked someone on the gbdev Discord server to verify and they get that on hardware too. https://discord.com/channels/303217943234215948/303217943234215948/875199303184941106
Screenshot from 2021-08-11 22-20-09

The code for the rom can be found at https://github.com/ICanNotThinkOfAGoodUsername/nr31nr41test

ISSOtm (approved):

PR #345, opened 2021-08-14
Remove useless DB.

GreenAndEievui:
Someone in the Discord server was confused by this DB. It doesn't have any purpose within the function, and isn't even valid syntax.


ISSOtm:
This `db` doesn't make any sense, and the `WRAM0` part wouldn't even compile, lol.

ISSOtm (approved):

PR #346, opened 2021-08-20
Fix descriptions of sprite position bytes

bfabio:


ISSOtm (approved):

rondnelson99:
Is it just me or are these changes wrong? For instance if a sprite is on the left edge of the screen, the OAM X position is screen position plus 8, so 0+8=8. If it was 0-8 it would be -8 which is not correct. I might be just reading something wrong though.

ISSOtm:
Depends on how you read it... "screen position plus 8" can be interpreted as "I must write the desired position plus 8", but also as "The OAM position is the on-screen position plus 8, so it's shifted downwards by 8 pixels".

Maybe more unambiguous wording is preferable, then, talking unambiguously about the coordinate offsets. By the way, do we mention the orientation of the X and Y axes? If not, this would be a good opportunity to add that.

bfabio:
@rondnelson99 as @ISSOtm said, it got me confused because of the interpretation I gave it. I agree we should reword it altogether with something else.

avivace (changes requested):
I don't think this solves the issue. As ISSO mentined, this requires a better wording overall, clarifying, maybe with an example, what is being and in what way (i.e. where is the sprite being drawn in respect of what did I see in Y and X positions?)

ISSOtm:
Since we agree that a clarification is in order, even though this PR doesn't solve the issue, shouldn't we at least keep it open as a reminder of the work item? Making it into a WIP should be enough to point out that the PR's contents are not to be merged as-is.

avivace:
> Since we agree that a clarification is in order, even though this PR doesn't solve the issue, shouldn't we at least keep it open as a reminder of the work item? Making it into a WIP should be enough to point out that the PR's contents are not to be merged as-is.

I will open an issue about this.

PR #350, opened 2021-09-01
Overhaul audio section

### LINKED ISSUE #119: CH2 register descriptions are copy-pasted from CH1
Since both pulse channels work exactly identically, save for CH1's sweep, shouldn't CH2's reg descriptions be made to point to CH1's?

### LINKED ISSUE #120: Clarify which side S01 and S02 terminals are

### LINKED ISSUE #347: Document APU DAC behavior
For example, it is currently not documented that writing %0000_0xxx to NRx2 (except NR32) or clearing NR30 bit 7 immediately kills the channel.

References in SameBoy's source: [[1]](https://github.com/LIJI32/SameBoy/blob/3c6d0947635ddc2da9809bf6e657a8302aca4894/Core/apu.c#L999-L1004) [[2]](https://github.com/LIJI32/SameBoy/blob/3c6d0947635ddc2da9809bf6e657a8302aca4894/Core/apu.c#L1150-L1156)
Those are likely covered by some test ROMs that we can cite instead, though.

ISSOtm:
Waaay not finished, only posting for discussion and early review. (Anything changed and not marked with a TODO is intended to be final.)

I won't be requesting any reviews until it's finished, but please feel free to chime in so as to make the process piecemeal as possible :)

TODO:

- [x] Rename channel "frequency" to "wavelength" or "period", since it's an unit of time and not a frequency.
- [x] Merge CH2 descriptions into CH1? All CH2 stuff would either be duplicates or "c.f.", so why not merge them? (How, though?)

Fixes #347 and fixes #119.

~~An issue that @LIJI32 raised is the time units. Currently, I only use "APU ticks", being 2 MiHz (the smallest common denominator), and provide measurements in multiples of those. Liji instead suggested introducing separate units whenever necessary (e.g. "1 128 Hz tick" instead of "16 [APU] ticks"). The pro is that the units make more sense (no multiples of powers of 2), the con is potentially more mental complexity.~~

~~I'm not attached to either, so I used the solution I originally came up with in this PR, please feel free to discuss and provide feedback in the comments.~~

~~Other names for the time unit were suggested: "sample" (given that 2 MiHz *is* the final sample rate), "boop", maybe others I missed?~~


Lycoder:
I think the "Common concepts" section should precede everything else, otherwise, really good job!

ISSOtm:
Thanks!

It's not possible to talk about any of these concepts without introducing the architecture (e.g. channels). That said, maybe moving the "APU" section before the architecture would indeed be better? I also want to introduce the diagram early, so that it can serve as a visual aid to go back to for the rest of the page.

Lycoder (approved):

src/Audio.md
```diff
@@ -0,0 +1,103 @@
...
+Each channel's [**envelope**](https://en.wikipedia.org/wiki/Envelope_(music)) can be configured.
+A signal's [envelope](https://en.wikipedia.org/wiki/Envelope_(waves)) is, roughly speaking, the variation of the signal's amplitude.
+On the GB, "envelope" basically designates the channel's loudness.
+
+All channels can have their envelope controlled to modulate their volume (CH3 less precisely than other channels).
+In addition, all channels but CH3 benefit from a functionality called **envelope sweep**, which automatically adjusts the envelope over time.
+The parameters that can be controlled are how many times the envelope will be stepped, in which direction, and how long between steps.
+Envelope sweep ticks at 64 Hz, and every 1&ndash;7 of those ticks, the envelope will be increased or decreased.
+
+Unfortunately, of [ADSR envelopes](https://en.wikipedia.org/wiki/ADSR_envelope), the Game Boy only provides for Attack and Sustain, via envelope sweep and sound length respectively.
```
nitro2k01, line 68: This is not really correct. In envelope terms, the GB envelope generators would be either attack/sustain or just decay, depending on the direction. And while the sound length register could be used to emulate (effectively) sustain and instant release, it could also be used to stop the channel during the decay phase. I don't think it's necessarily a good idea to try to fit the GB envelope control into exact ADSR terminology, but rather try to explain what each function does on its own.

src/APU_details.md
```diff
@@ -0,0 +1,98 @@
...
+-   CH3 output level control does not, in fact, alter the output level.
+    It shifts the **digital** value CH3 is outputting (read below), not
+    the analog value.
+-   On GBA, CH3 is inverted. This causes the channel to output a loud
+    spike when disabled; it's a good idea to "remove" the channel
+    using NR51 before refreshing wave RAM.
+
+## APU detailed architecture
+
+{{#include imgs/apu_detailed.svg:2:}}
```
nitro2k01, line 38: The "amplifiers" are actually attenuators, I'm pretty sure. Also, does any Gameboy actually have a HP filter in the SoC? I know DMG and GBC have HPF->Volume dial->amp->HPF external to the SoC, at least. Not sure what level of abstraction we should explain this at though, but it may be a good idea to distinguish between what's happening inside and outside the SoC.

nitro2k01 (changes requested):

src/Audio.md
```diff
@@ -0,0 +1,103 @@
...
+Unlike other sound chips, such as the SNES' S-DSP or the Atari 5200's POKEY, each sound channel is largely different from the others.
+
+Each channel generates an electronic signal; these signals are then mixed into two new channels (for stereo: one for the left ear, one for the right ear), which are then individually amplified, and then output either to the headphone jack, or the speaker[^speaker_mono].
+
+Channels 1 and 2, the "pulse channels", produce [pulse (rectangular) waves](https://en.wikipedia.org/wiki/Pulse_wave).
+Channel 3, the "wave" channel, produces user-defined waves.
+Channel 4 is the "[noise](https://en.wikipedia.org/wiki/Noise_in_music)" channel, producing a pseudo-random wave.
+
+The VIN channel is a signal received directly from the cartridge, allowing external hardware to supply a fifth sound channel.
+No licensed games used this feature, and it was omitted from the Game Boy Advance.
```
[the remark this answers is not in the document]
ISSOtm: Got some links?

src/Audio.md
```diff
@@ -0,0 +1,103 @@
...
+
+::: tip POCKET MUSIC
+
+Despite rumors, *Pocket Music* does not use VIN. It refuses to run on the GBA for a different reason:
```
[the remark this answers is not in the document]
ISSOtm: I don't know, I believe @pinobatch added this line.

src/Audio.md
```diff
@@ -0,0 +1,103 @@
...
+:::
+
+### Envelope
+
+Each channel's [**envelope**](https://en.wikipedia.org/wiki/Envelope_(music)) can be configured.
+A signal's [envelope](https://en.wikipedia.org/wiki/Envelope_(waves)) is, roughly speaking, the variation of the signal's amplitude.
+On the GB, "envelope" basically designates the channel's loudness.
+
+All channels can have their envelope controlled to modulate their volume (CH3 less precisely than other channels).
+In addition, all channels but CH3 benefit from a functionality called **envelope sweep**, which automatically adjusts the envelope over time.
```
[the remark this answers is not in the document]
ISSOtm: Envelope is a modulation of amplitude, here a volume setting, so the terminology seems acceptable imo.

It's not possible to change this terminology, either, because it's used literally everywhere (including [hardware.inc](https://github.com/gbdev/hardware.inc/blob/05f5a9b6c7172abe1d7488080c1c050284c09226/hardware.inc#L230-L237)).

src/Audio.md
```diff
@@ -0,0 +1,103 @@
...
+This can be used to time the "sustain" part of ADSR, though the sound will cut fairly abruptly once the sound length "expires", instead of decaying.
+
+A channel's sound length is initially set between 0 and 63, and, if enabled, ticks up[^len_cnt_dir] at 256 Hz (tied to [DIV-APU](<#DIV-APU>)).
+When the counter reaches 64, the channel is turned off.
+
+[^len_cnt_dir]:
+Actually, the sound length is inverted when written, and *that* ticks down until it reaches 0.
+But the effect is as if the counter ticked up.
+
+### Wave length
```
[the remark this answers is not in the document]
ISSOtm: Ok with contracting.

What's the difference between the "conversion formula" and the "underlying function"?

Calling it "frequency" is outright wrong, since it's not what's being controlled, and has proven to cause confusion. This is why I'm changing it.

src/Audio.md
ISSOtm: Okay, if you say so.

src/APU_details.md
ISSOtm: What would be the point?

src/Audio.md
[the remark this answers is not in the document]
nitro2k01: The envelope is technically implemented as a sweep in the sense that it interpolates from one value to another. But hardware.inc, which copied the comment from nocash Pan Docs, seems to refer to the lower 3 bits of NRx2 as "Number of envelope sweep", and not the whole function of generating the envelope. In the overview, it was described like this with no mention of sweep (except referring to ch1's pitch sweep.)

>    Quadrangular wave patterns with sweep and envelope functions.
>    Quadrangular wave patterns with envelope functions.
>    Voluntary wave patterns from wave RAM.
>    White noise with an envelope function.

If anything, this seems like one of many quirks in Martin's English which should probably be improved instead of embraced.

Also note that nocash Pan Docs and hardware.inc don't refer to ch3 (`rNR32`) as having an envelope/being an envelope setting, but an output level.

src/Audio.md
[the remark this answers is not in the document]
nitro2k01: By conversion formula, I mean literally the equation "Frequency = 131072/(2048-x) Hz" or some variant of it. 

By underlying function, I mean (a description of) how and why the oscillator functions the way it does.

And I'm not suggesting that the value itself should be referred to as a frequency, since it's obviously not. What I'm saying is that the section should possibly be called frequency, and things should be explained in terms of controlling the frequency. This makes educational sense because it's describing the hardware starting form familiar territory. (Assuming you have knowledge about music/sound theory.)

As a first approximation, this could be described as a special pitch value used by the hardware which can be calculated from the frequency.

Then a later paragraph can describe the exact meaning of the value, and how it's used internally by the hardware.

src/APU_details.md
nitro2k01: Firstly, the HPF is not part of the APU itself, as far as I can tell, so describing it as being part of the APU seems out of place and slightly inaccurate.

Secondly, describing the external audio circuitry (in one way or the other, not necessarily in the same diagram) can be useful information, for example for understanding the motivation for the "Prosound" hardware modification.

src/Audio.md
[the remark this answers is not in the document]
nitro2k01: Nanoloop: https://www.nanoloop.com/mono/index.html

PocketVoice:
http://n64.icequake.net/mirror/www.bung.com.hk/porducts/pocketvoice.htm
http://n64.icequake.net/mirror/www.bung.com.hk/support/download.htm

Nanoloop is still being sold, whereas PocketVoice hasn't been sold in about 20 years since Bung went under, and was probably one of their more obscure products. I think it's enough for now to just mention in passing that "no licensed games used the VIN pin, although two unlicensed cartridges, Bung PocketVoice, and Nanoloop Mono did."

Speaking of which... we shold probably also explain how...

- the VIN pin is sometimes used as a flash /WR pin.
- how it therefore often has a pullup resistor.
- how this means the behavior of the modulating VIN can produce different results depending on the cartridge it's running from.
- How one homebrew game (which I don't remember the name of) mistakenly used 'rNR50` as a 2x4 bit PCM register, because on the cart it was developed on the DC offset on VIN was just right for this to work ok, but won't always work as per the previous point.

src/APU_details.md
ISSOtm: Even for a "detailed" section of Pan Docs, there's a limit what the reader can mentally process... and also to what I can write. I would prefer not to make that distinction; if you deem it necessary, we could rename the section and/or page from "APU" to "audio", I guess.

src/Audio.md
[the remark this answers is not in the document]
ISSOtm: Resolved locally by using "Frequency" as the section name. The conversion formula is not provided in the overview, simply because it depends on the channel.

src/Audio.md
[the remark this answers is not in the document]
ISSOtm: Mm, honestly, I'd rather not touch VIN. This PR is enough of a brain space nightmare as-is. Open a separate issue, maybe?

aaaaaa123456789:
Minor trivial feedback: use either "wave length" or "wavelength" consistently; you're currently using both.

ISSOtm:
As per https://github.com/gbdev/pandocs/pull/350#discussion_r709713631, it should be "wavelength". That should be added to the document style, and it'll have to be caught during review due to size of the change.

avivace:
Hi @ISSOtm did you have any chance to come back on this? I will be  fixing the conflict/rebasing, splitting it and merging what we have if no activity is expected in the near future as I don't want this to become blocking or get lost in time

ISSOtm:
I am not planning to work on this in the foreseeable future due to being busy on many other things (as you know), but I can provide guidance to anyone else willing to pick this up. I don't think it's production-ready in its current state due to how incomplete it is—it doesn't have all the info the page it's meant to replace has.

Most of the info is available though not necessarily readily, if only from the amount of people writing GB emulators by the day.

src/Audio.md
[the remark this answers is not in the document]
pinobatch: I apologize for not having saved receipts when I first read this rumor.

ISSOtm:
Let's not let it get lost to time. :)

ISSOtm:
"Audio details" is also almost complete, and ready for feedback.

src/Audio_Registers.md
```diff
@@ -0,0 +1,383 @@
...
+
+This register stores the low 8 bits of the channel's 11-bit "[wavelength](<#Frequency>)".
+The upper 3 bits are stored in the low 3 bits of `NR34`.
+
+The actual signal frequency is <math><mfrac><mn>65536</mn><mrow><mn>2048</mn><mo>-</mo><mi>wavelen</mi></mrow></mfrac></math> Hz: the higher the value, the higher the frequency.
+This is the whole wave's frequency; the rate at which the channel steps through the 8 "indices" in its wave form is 32 times that, i.e. <math><mfrac><mn>2097152</mn><mrow><mn>2048</mn><mo>-</mo><mi>wavelen</mi></mrow></mfrac></math>) Hz = <math><mfrac><mn>2</mn><mrow><mn>2048</mn><mo>-</mo><mi>wavelen</mi></mrow></mfrac></math> MiHz.
+
+::: warning DELAY
+
+Modifications of the frequency (via `NR33` or `NR34`) only take effect after the following time wave RAM is read.
```
avivace, line 266:
```suggestion
Changes of frequency (via `NR33` or `NR34`) only take effect after the following time wave RAM is read.
```

src/Audio_Registers.md
```diff
@@ -0,0 +1,383 @@
...
+
+On monochrome consoles only, retriggering CH3 while it's about to read a byte from wave RAM causes wave RAM to be corrupted in a generally unpredictable manner.
+
+:::
+
+Bit 6 takes effect immediately upon writing to this register.
+
+::: warning PLAYBACK DELAY
+
+Triggering the wave channel does not immediately start playing wave RAM; instead, the *last* sample ever read (which is reset to 0 when the APU is off) is output until the channel next reads a sample.
```
avivace, line 292:
```suggestion
Triggering the wave channel does not immediately start playing wave RAM; instead, the *last* sample ever read (which is reset to 0 when the APU is off) is played until the channel next reads a sample.
```

src/Audio.md
```diff
@@ -0,0 +1,110 @@
...
+
+### APU
+
+The Game Boy's sound chip is called the <abbr title="Audio Processing Unit">APU</abbr>.
+
+The APU runs off the same master clock as the rest of the Game Boy, which is to say, it is fully synced with the CPU and [PPU](<#Rendering Overview>).
+This also means that the APU runs about 2.4% faster on the SGB1, increasing frequencies by as much and thus sounding slightly higher-pitched.
+The SGB2 rectifies this issue.
+
+The APU processes a lot of events at a fixed rate dependent on the event's type.
```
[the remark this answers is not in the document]
ISSOtm: Which rate it is depends on the event's type, but it's fixed for each type.
```suggestion
The APU processes a lot of events at fixed rates dependent on the event's type.
```
```suggestion
The APU processes a lot of events, at a different fixed rate depending on the event's type.
```

ISSOtm:
Done! Commence the bikeshedding :)

src/Audio_details.md
```diff
@@ -0,0 +1,192 @@
...
+## Game Boy Advance audio
+
+The APU was reworked pretty heavily for the GBA, which introduces some slightly different behavior:
+- Instead of mixing being done analogically, it's instead done digitally; then, sound is converted to an analog signal and an offset is added (see `SOUNDBIAS` in [GBATEK](http://problemkaputt.de/gbatek.htm#gbasoundcontrolregisters) for more details).
+- This also means that the GBA APU has no DACs.
+  Instead, they are emulated digitally such that a disabled "DAC" behaves like an enabled DAC receiving 0 as its input.
+- Additionally, CH3's DAC has its output inverted.
+  In particular, this causes the channel to emit a loud spike when disabled; therefore, it's a good idea to "disconnect" the channel using NR51 before accessing wave RAM.
+
+None of the "enhanced" features (more wave RAM, digital FIFOs, etc.) are available to CGB programs.
```
avivace, line 192:
```suggestion
None of the additional features (more wave RAM, digital FIFOs, etc.) are available to CGB programs.
```

avivace (changes requested):

src/Audio_Registers.md
avivace: "or 0 if none was since the APU was last turned on" sounds still very confusing to read ..

ISSOtm: Edited, I had a word.

src/Audio_details.md
```diff
@@ -0,0 +1,192 @@
...
+CH3 has an internal "sample index" counter.
+Each time it is ticked (as determined by the "wavelength" in NR33/NR34), that "sample index" is incremented, and then the corresponding "sample" (nibble) is read from wave RAM.
+(This means that sample #0 is skipped when first starting up CH3.)
+
+CH3 does not emit samples directly, but stores every sample read into a buffer, and emits that continuously; (re)triggering the channel does *not* clear nor refresh this buffer, so the last sample ever read will be emitted again.
+This buffer *is* cleared when turning the APU on, so CH3 will emit a "digital 0" when first powered on.
+
+CH3 output level control does not, in fact, alter the output level.
+It shifts the **digital** value CH3 is outputting, not the analog value.
+This only matters when changing the setting mid-playback: the digital values being shifted bias them towards 0, which biases the analog output towards "1"; the HPF will smooth this over time, but not instantly.
```
avivace, line 159: I'm not super sure about how we are using `bias` as a verb here. We should try to find something more technical.

src/Audio_details.md
```diff
@@ -0,0 +1,192 @@
...
+The opposite is not true, however: a disabled channel outputs 0, which an enabled DAC will dutifully convert into "analog 1".
+
+A channel can be deactivated in one of the following ways:
+- Turning off its DAC
+- Its [length timer](<#Length timer>) expiring
+- (CH1 only) [Frequency sweep](<#FF10 — NR10: Channel 1 sweep>) overflowing the frequency
+
+### Pulse channels (CH1, CH2)
+
+Each pulse channel has an internal "duty step" counter, which is used to index into [the selected waveform](<#FF11 — NR11: Channel 1 length timer & duty cycle>) (each background stripe corresponds to one "duty step")[^pulse_lut].
```
avivace, line 137: This doesn't read very well.. The internal counter is used to index (what?) into a waveform? What does it mean to index into a waveform?

src/Audio_details.md
```diff
@@ -0,0 +1,192 @@
...
+The nature of this fade is not entirely deterministic and varies between models.
+
+NR52's low 4 bits report whether the channels are turned on, not their DACs.
+
+### Channels
+
+A channel is activated by a write to NRx4's MSB, unless its DAC is off, which forces it to be disabled as well.
+The opposite is not true, however: a disabled channel outputs 0, which an enabled DAC will dutifully convert into "analog 1".
+
+A channel can be deactivated in one of the following ways:
```
avivace, line 130:
```suggestion
A channel is deactivated in the following cases:
```

src/Audio_details.md
```diff
@@ -0,0 +1,192 @@
...
+- Turning off its DAC
```
avivace, line 131:
```suggestion
- Its DAC is turned off
```

src/Audio_details.md
```diff
@@ -0,0 +1,192 @@
...
+- Its [length timer](<#Length timer>) expiring
```
avivace, line 132:
```suggestion
- Its [length timer](<#Length timer>) expires
```

src/Audio_details.md
```diff
@@ -0,0 +1,192 @@
...
+- (CH1 only) [Frequency sweep](<#FF10 — NR10: Channel 1 sweep>) overflowing the frequency
```
avivace, line 133:
```suggestion
- (CH1 only) The [Frequency sweep](<#FF10 — NR10: Channel 1 sweep>) overflows the frequency
```

src/Audio_details.md
```diff
@@ -0,0 +1,192 @@
...
+
+The HPF is more aggressive on GBA than on GBC, which itself is more aggressive than on DMG.
+(The more "aggressive" a HPF, the faster it pulls the signal towards "analog 0"; this reduces the pops, but tends to also distort waveforms.)
+
+### DACs
+
+Channel <var>x</var>'s DAC is enabled if and only if `[NRx2] & $F8 != 0`; the exception is CH3, whose DAC is directly controlled by bit 7 of NR30 instead.
+Note that the envelope functionality changes the volume, but not the value stored in NRx2, and thus doesn't disable the DACs.
+
+If a DAC is enabled, the digital range $0 to $F is lineraly translated to the analog range -1 to 1, in arbitrary units.
```
avivace, line 117:
```suggestion
If a DAC is enabled, the digital range $0 to $F is linearly translated to the analog range -1 to 1, in arbitrary units.
```

avivace, line 117: What does "in arbitrary units" wants to communicate here? Aren't we in arbitrary units already in the digital range?

src/Audio_details.md
```diff
@@ -0,0 +1,192 @@
...
+Sound length   | 2             | 256 Hz
+CH1 freq sweep | 4             | 128 Hz
+
+[^div_apu_freq]:
+Indicated values are under normal operation; the frequencies will obviously differ if writing to `DIV` to increase the counter faster.
+
+### Mixer
+
+A high-pass filter (HPF) removes constant biases over time.
+The HPFs therefore remove the DC offset created by inactive channels with an enabled DAC, and off-center waveforms.
```
avivace, line 104: How we moved from a single HPF to more than one? Should this be clarified?

src/Audio_details.md
```diff
@@ -0,0 +1,192 @@
...
+
+Enabling or disabling a DAC ([see below](#DACs)), adding or removing it using NR51, or changing the volume in NR50, will cause an audio pop.
```
avivace, line 106: What do you mean here with "audio pop"?

src/Audio_details.md
```diff
@@ -0,0 +1,192 @@
...
+(All of these actions cause a change in DC offset, which is smoothed out by the HPFs over time, but still creates a pop.)
+
+The HPF is more aggressive on GBA than on GBC, which itself is more aggressive than on DMG.
```
avivace, line 109: This should be reworded in a way that doesn't require the explanation of what it means to be aggressive *afterwards*.

src/Audio_details.md
```diff
@@ -0,0 +1,192 @@
...
+Same, but with channels 3 and 4.
+
+## Finer technical explanation
+
+### DIV-APU
+
+A "DIV-APU" counter is increased every time `DIV`'s bit 4 (5 in [double-speed mode](<#FF4D — KEY1 (CGB Mode only): Prepare speed switch>)) goes from 1 to 0, therefore at a frequency of 512 Hz (regardless of whether double-speed is active).
+Thus, the counter can be made to increase faster by writing to `DIV` while its relevant bit is set (which clears `DIV`, and triggers the falling edge).
+
+Some events are tied to the DIV-APU counter:
```
avivace, line 90:
```suggestion
The rate of the following events are tied to the DIV-APU counter:
```

src/Audio_details.md
```diff
@@ -0,0 +1,192 @@
...
+
+Event          | Every N ticks | Frequency[^div_apu_freq]
```
avivace, line 92: I feel like there is a better way to word "Every N ticks" but can't come up with one right now

src/Audio_details.md
```diff
@@ -0,0 +1,192 @@
...
+To be clear: digital values are discrete and clear-cut; conversely, the analog domain is continuous.
+The former is what computers use, the latter is what the real world is made of.
+
+[^vin]:
+Actually, VIN acts as a 5<sup>th</sup> channel fed into the mixer, whose control bits are in NR50 instead of NR51.
+This was omitted from the diagram for simplicity.
+
+## PCM registers
+
+These two registers, only present in the Game Boy Color and ulterior models, allow reading the output of the generation circuits directly; this is very useful for testing "internal" APU behavior.
```
avivace, line 72: What is "ulterior" here? Do you mean from more recent than GBC? Maybe "further" models?

src/Audio_details.md
```diff
@@ -0,0 +1,192 @@
...
+These registers are not documented in any known Nintendo manual.
```
avivace, line 73: What is the purpose of this line? If we specify here should we do it elsewhere through the document as well?

src/Audio_details.md
```diff
@@ -0,0 +1,192 @@
...
+Thus, the analog range of those outputs is 4× that of each channel, -4 to 4.
+
+Then, both of these two get their respective volume scaled, once from [`NR50`](<#FF24 — NR50: Master volume & VIN panning>), and once from the volume knob (if the console has one).
+Note that the former step never mutes a non-silent input, but the latter can.
+
+Each of the two analog outputs then goes through a [high-pass filter](https://en.wikipedia.org/wiki/High-pass_filter) (HPF).
+For short, a HPF constantly tries to "pull" the signal towards analog 0 (neutral); the reason for that is explained further below.
+
+[^digital_analog]:
+To be clear: digital values are discrete and clear-cut; conversely, the analog domain is continuous.
```
avivace, line 63: I feel like this clarification shouldn't be here. Maybe it's something we can add to an initial glossary/list of necessary concepts needed to read Pan Docs?

avivace (changes requested):

src/Audio_details.md
ISSOtm: The "waveform" here is meant as an array of booleans, which is being indexed into.

src/Audio_details.md
ISSOtm: I'm trying to say that the list is exhaustive;

src/Audio_details.md
ISSOtm: Yes, but those are different unary units, and the usual unit for analog signals is volts, and we don't want to imply that sound ranges from -1 V to 1 V. I'm not sure the exact values are known, and they would muddy the actual explanations throughout the page.

src/Audio_details.md
ISSOtm: There have always been two HPFs, as per the diagram.

src/Audio_details.md
ISSOtm: An audio... "pop"! It's a sound.

src/Audio_details.md
ISSOtm: That would break the flow of the comparison and make it hard to read. Introducing the concept first would be confusing, as in "why am I being explained this?".

src/Audio_details.md
ISSOtm: It's not just the rate, it's their occurrence. Repeatedly writing to DIV allows triggering them irregularly, or not at all.

src/Audio_details.md
ISSOtm: In theory, "Period (ticks)", but I avoided it because I'm pretty sure it would be more confusing.

src/Audio_details.md
ISSOtm: Yes, this is what I meant; however, "further" would be incorrect here.

src/Audio_details.md
ISSOtm: Pan Docs currently does not document any such registers [besides these](https://gbdev.io/pandocs/CGB_Registers.html#undocumented-registers), which I am lifting PCMxx from. The line's purpose is to give context, explain why no commercial software uses it, etc. They might be debugging leftovers, for all we know.

src/Audio_details.md
ISSOtm: Sure, but this PR has waited long enough that I don't want it to become blocked by #194 (which itself has been twiddling its thumbs for a long while now).

src/Audio_details.md
avivace: Which is what rate conveys here. If I understand this correctly, it's not really "irregular", it's still every *N* ticks. Then the fact that those ticks can happen or not it's a separate concept. The event A is still going to have a rate of 1/N DIV-APU ticks.

src/Audio_details.md
ISSOtm: I believe "to bias" is a verb used fairly often in signal processing.

src/Audio_details.md
ISSOtm:
```suggestion
The following events occur every <var>N</var> DIV-APU ticks:
```
Then we'd replace the "Every N ticks" column header with "Rate".

src/Audio.md
```diff
@@ -0,0 +1,110 @@
+# Audio Overview
+
+Game Boy audio is sometimes called "8-bit", but this doesn't mean much: the term is used to designate audio similar to that produced by 8-bit consoles.
+
+A better description would be "chiptune": tunes produced by a chip.
```
nitro2k01, line 5: No! No! No! Please avoid the use of the word chiptune. It's one of the most controversial terms in the chiptune scene itself, when it comes to an exact definition. It can means anything from an Amiga module with short loopes, so called "chipped", samples (the origin of the term) to a purist view that it can only refer to music generated by hardware, to a more liberal view that it can mean music in a cetain style, allowing a song to be made with a Game Boy but not be "chiptune". In short it can mean anything and everything depending on who you ask, so in the end it means nothing.

ISSOtm: Do you have a better wording for that sentence, or should it be removed entirely?

src/Audio.md
```diff
@@ -0,0 +1,110 @@
...
+## Architecture
+
+{{#include imgs/apu_overview.svg:2:}}
+
+The Game Boy has four sound generation units, called **channels** 1 through 4, notated "CH1", "CH2", etc.
+Unlike some other sound chips, such as the SNES' S-DSP or the Atari 5200's POKEY, each sound channel is specialized in a way largely different from the other channels.
+
+Each channel generates an electronic signal; these signals are then mixed into two new channels (for stereo: one for the left ear, one for the right ear), which are then individually amplified, and then output either to the headphone jack, or the speaker[^speaker_mono].
+
+Channels 1 and 2, the "pulse channels", produce [pulse (rectangular) waves](https://en.wikipedia.org/wiki/Pulse_wave).
```
nitro2k01, line 17: Not sure about calling it rectangular wave. This seems like a more academic terminology, and is in my experience not commonly used in synthesis for music. (A legacy from Martin's "quadrangular" waves?) I would rather say something like "produce [pulse width modulated waves](https://en.wikipedia.org/wiki/Pulse-width_modulation) with 4 fixed pulse width settings."

src/Audio.md
```diff
@@ -0,0 +1,110 @@
...
+Channel 3, the "wave" channel, produces arbitrary user-supplied waves.
+Channel 4 is the "[noise](https://en.wikipedia.org/wiki/Noise_in_music)" channel, producing a pseudo-random wave.
+
+The VIN channel is a signal received directly from the cartridge, allowing external hardware to supply a fifth sound channel.
```
nitro2k01, line 21: I would specify "is an analog signal..." here.

src/Audio.md
ISSOtm: So this?
```suggestion
Channels 1 and 2, the "pulse channels", produce [pulse width modulated waves](https://en.wikipedia.org/wiki/Pulse-width_modulation) with 4 fixed pulse width settings.
```

src/Audio.md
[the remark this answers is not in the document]
nitro2k01: In my opinion, this whole sentence is at the wrong level of abstraction for this section of the document and should probably be removed entirely.

src/Audio.md
ISSOtm:
```suggestion
The VIN channel is an analog signal received directly from the cartridge, allowing external hardware to supply a fifth sound channel.
```

src/Audio.md
```diff
@@ -0,0 +1,110 @@
...
+
+**Triggering** a channel causes it to turn on if it wasn't[^trig_dac_off], and to start playing its wave from the beginning[^pulse_restart].
+Most changes to a channel's parameters take effect immediately, but some require re-triggering the channel.
+
+### Volume & envelope
+
+The volume can be controlled in two ways: there is a "master volume" control[^vol_knob] (which has separate settings for the left and right outputs), and each channel's volume can be individually set as well (CH3's less precisely than the others).
+
+Additionally, an [**envelope**](https://en.wikipedia.org/wiki/Envelope_(music)) can be configured for all channels except CH3, which allows automatically adjusting the volume over time.
+The parameters that can be controlled are how many times the envelope will be stepped, in which direction, and how long between steps.
```
nitro2k01, line 67: Too heavy on technical primitives. I'd describe this as something like:
> The envelope register allows setting the initial volume, the envelope direction and envelope speed.

ISSOtm: I'd avoid the word "register" here too, then:
```suggestion
The parameters that can be controlled are the initial volume, the envelope's direction, and its speed.
```

src/Audio.md
```diff
@@ -0,0 +1,110 @@
...
+All envelopes are ticked at 64 Hz, and every 1–7 of those ticks, the volume will be increased or decreased.
```
nitro2k01, line 68: This is fine as is, after the above change, except maybe it should start with "Internally, all envelopes..."

src/Audio.md
```diff
@@ -0,0 +1,110 @@
...
+
+The term "wavelength" throughout this document does *not* designate the inverse of the frequency, but instead a quantity akin to it.
+See the description of each NRx3 register for more information.
+
+:::
+
+---
+
+[^trig_dac_off]:
+If [the channel's DAC](#DACs) is off, the channel will not turn on.
```
nitro2k01, line 92: This footnote makes little sense in the context it's linked from. "The channel will not turn on" in which situation? I think what this means to say is that the channel doesn't initially turn on just by writing other registers than bit 7 of the NRx4 but this doesn't come across.

ISSOtm: I meant that setting NRx4 bit 7 while the DAC is still off will not turn the channel on. The DAC must be turned on first, then the channel triggered.

src/Audio.md
```diff
@@ -0,0 +1,110 @@
...
+
+[^pulse_restart]:
+Except for pulse channels, whose index into their waves is only ever reset by turning the APU off.
+This is not noticeable when playing sound, and only matters for "advanced" APU usage.
```
nitro2k01, line 96: Following the theme of not describing things solely in terms of primitives, "index into their waves" should probably be "phase" or "phase position". Also, I'm not sure this information should be hidden away in a footnote since it applies to half of the available channels. Also, not sure exactly what counts as advanced APU usage, but a fairly common trick is to play two slightly detuned notes on two channels, which can sound different every time due to this effect.

src/Audio.md
```diff
@@ -0,0 +1,110 @@
...
+[^vol_knob]:
+This is separate from the physical volume knob located on the side of the console.
+
+[^len_cnt_dir]:
+Internally, the length timer is inverted when written, and *that* ticks down until it reaches 0.
+But the effect is as if the counter ticked up.
+
+[^pitch]:
+There is also **pitch**, which is merely a measure of how we perceive frequency.
+The higher the frequency, the higher the pitch; therefore, pitch will be omitted from the rest of the document.
```
nitro2k01, line 107: Actually, rather than omitting pitch from the document, maybe using pitch as the preferred term instead of frequency is useful. Consider especially the noise channel, which for some settings doesn't have a clearly defined frequency, in terms of what the human ear can perceive. (For example 15-bit polynomial setting will produce a cycle that can be more than a second long.) Yet, it would be correct to say that we can vary the pitch of the sound.

src/Audio.md
ISSOtm: This is my turn being confused—what does "detuned" mean?

I think "phase position" would best describe the intent here.

src/Audio.md
ISSOtm: That doesn't mesh with the "registers" document, which expresses frequencies in terms of what is written to the registers. Pitch can be derived from that, but it's non-trivial.

src/Audio.md
nitro2k01: Register is fine, since that's the way the programmer interfaces with the hardware, and something you need to learn early to do anything with the hardware. What I really mean by technical primitives, is _hidden_ technical primitives. Ie, new extra concepts you have to understand, just to read a basic sentence. With that said, there's nothing wrong with the suggested change, per se.

ISSOtm: I'm generally aiming for `Audio.md` to be readable by non-programmers, `Audio_Registers.md` by programmers, and `Audio_details.md` by people who really need to dive into some of the details (programmers trying to understand/debug weird behaviour, emu devs, etc.)

src/Audio.md
nitro2k01: Ok, I have to admit I was unaware of this particular fact: 
> Channel <var>x</var>'s DAC is enabled if and only if `[NRx2] & $F8 != 0`;"
I was under the impression that this was not a prerequisite for turning on the channel (in terms of for example making it's phase counter start counting). Is the research for this behavior documented anywhere?

src/Audio.md
nitro2k01: Detuned just means the two notes have a slightly different frequency. The effect in question happens when two such notes create a "phasing" sound with each other. A simple way to illustrate the concept is for example using my bgbtest.gb ROM (which comes with BGB.) Press one of the buttons quickly repeatedly. Every second note is detuned, and if you do it quick enough, they overlap and create this effect.

What you can't do in a very predictable way, is an effect similar to a PWM that goes from a low percent duty to a 50% duty, since the phase position might be "random" on every note retrigger. I can make a sound example later.

src/Audio.md
nitro2k01: Not sure. I guess both the first two sentences could be omitted, but then again, I'm not sure how much context you need to explain, in particular for younger readers. (God, I feel old...) Maybe a fair assumption is that anyone reading the document has heard at least a few example of music/sound produced by the Game Boy in games?

> Game Boy audio is sometimes called "8-bit", but this doesn't mean much: the term is used to designate audio similar to that produced by 8-bit consoles.

As for this sentence. It's true that 8-bit can be used to designate audio similar to that produced by 8-bit consoles (similar to how chiptune can mean anything and everything) but that sentence almost seems to imply that "8-bit" is used only to describe _similar_ sounds, not the actual thing itself.

Maybe something like?

> Game Boy audio is sometimes called "8-bit". This does not refer to the bit depth of the sound generated, but rather that it has sound capabilities typical of 8-bit consoles. Like much of its contemporary hardware, the Game Boy produces sound generated by simple digital circuits.

src/Audio.md
```diff
@@ -0,0 +1,110 @@
...
+
+A better description would be "chiptune": tunes produced by a chip.
+This section describes how audio is generated on the Game Boy.
+
+## Architecture
+
+{{#include imgs/apu_overview.svg:2:}}
+
+The Game Boy has four sound generation units, called **channels** 1 through 4, notated "CH1", "CH2", etc.
+Unlike some other sound chips, such as the SNES' S-DSP or the Atari 5200's POKEY, each sound channel is specialized in a way largely different from the other channels.
```
nitro2k01, line 13: The SNES' S-DSP is one generation ahead. A better comparison might be the SID.

src/Audio.md
nitro2k01: Yes.

src/Audio_Registers.md
```diff
@@ -0,0 +1,375 @@
+# Audio Registers
+
+Audio registers are named following a `NRxy` scheme, where `x` is the channel number (or `5` for "global" registers), and `y` is the register's ID within the channel.
+Since many registers share common properties, a notation is often used where e.g. `NRx2` is used to designate `NR12`, `NR22`, `NR32`, and `NR42` at the same time, for simplicity.
+
+As a rule of thumb, for any `x` in `1`, `2`, `3`, `4`:
+- `NRx0` is some channel-specific feature (if present),
```
nitro2k01, line 7: Just state explicitly what it's used for, and that it's unused for the other two channels. Don't worry too much that this section must be completely generalized in its language at all cost.

src/Audio.md
ISSOtm: I don't know of any test ROM that checks this (maybe somewhere in blargg), but I just tested with [Numism's exerciser](https://github.com/pinobatch/numism/tree/main/gameboy/exercise#sound-exerciser), writing $80 to NR24 does nothing with NR22 = $00, but it does with e.g. $08.

src/Audio.md
avivace: Agree here, I'd avoid saying "index into their waves"

src/Audio.md
avivace: I like @nitro2k01 last suggestion.

src/Audio_details.md
avivace: "more recent"?

avivace: "following"?

src/Audio_details.md
avivace: It sounds better to me

src/Audio_details.md
ISSOtm: Either "more recent" or "later", I'd say.

src/Audio_details.md
avivace: I don't see V specified anywhere else here when speaking about analog ranges and/or signals. I think it's clear enough to refer to them without specifying units.

src/Audio_details.md
avivace: "later model" works for me too

src/Audio_Registers.md
ISSOtm: Considering that the list is all about "the rule of thumb", being generic is the name of the game. The goal of this paragraph is to give the reader a simple flowchart to quickly sort "uh, what *exactly* is NRxy?" (y = 1 ⇒ sound length, y = 2 ⇒ volume, y = 3 ⇒ freq low, y = 4 ⇒ freq hi + trigger)

ISSOtm: How would you word it?

ISSOtm:
Bump. I'd *really* like to merge this soon, please. It's going to become a blocker for other projects of mine soon.

avivace:
> Bump. I'd _really_ like to merge this soon, please. It's going to become a blocker for other projects of mine soon.

I can see some threads here still waiting for your follow up.. can you take a look?

ISSOtm:
I don't see any of the open threads currently waiting on me, can you give me some links?

avivace:
https://github.com/gbdev/pandocs/pull/350#discussion_r979757208

ISSOtm:
This one has been changed:

https://github.com/gbdev/pandocs/blob/a4049ecdb0fdfe63a8478e012511cf6b49198014/src/Audio.md?plain=1#L93

...and I'm waiting for approval or further modification requests on the updated text.

src/Audio_details.md
avivace: How is the counter influencing / "being indexed into" the waveform? Are you trying to say that it's being "encoded" in array used to generate the waveform?

avivace (changes requested):

src/Audio_details.md
ISSOtm: The waveform SVGs have stripes in the background, which are meant to indicate what the waveform's state is based on that "duty step"/phase counter. 0 being leftmost.

src/Audio_details.md
avivace: Added to that issue

src/Audio.md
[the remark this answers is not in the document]
ISSOtm: Updated.

src/Audio_details.md
ISSOtm: When reading documents just stating values without units, I default to assuming that it's a typo/oversight, and try to figure out what the unit is. Saying it explicitly clears that up.

src/Audio.md
ISSOtm: How is it now?

avivace: Ok to me, but I opened https://github.com/gbdev/pandocs/issues/445 to have the chance to continue this conversation

avivace:
A number of very interesting threads were opened by @nitro2k01, for some of which I feel they should've been blocking, but this PR has been sitting here for more than a year and I really don't feel like it's progressing in any sensible way since months.

Exceptionally, I "moved" those conversations into issues #444 #445 #446 and I'm going to merge what we have now in this chapter, which is really a block of awesome work and information previously scattered unorganically around.

Thanks @ISSOtm and @nitro2k01 ! Let's try to keep the conversation going in those issues and open more PRs to iterate upon this content.

PR #354, opened 2021-09-13
Fix misc. formatting in "Other MBCs" chapter

ISSOtm:


src/othermbc.md
```diff
@@ -16,22 +16,23 @@ outputs, but this appears not to have been done in licensed games.
...
 ### EMS
 
-PinoBatch learned the game selection protocol for EMS flash carts from
-beware, who in turn learned it from nitro2k01. Take this with a grain of
-salt, as it hasn't been verified on the authentic EMS hardware.
+::: warning
 
-A [header](<#The Cartridge Header>) matching any of the
-following is detected as EMS mapper:
+Take the following with a grain of salt, as it hasn't been verified on authentic EMS hardware.
```
avivace, line 21: We should use a "warning" box with "speculation" or "research needed" as title. This also applies to the other info/warning box around that start with something describing an unverified/speculated behaviour

src/othermbc.md
```diff
@@ -16,22 +16,23 @@ outputs, but this appears not to have been done in licensed games.
...
+- Header name is "GB16M", NUL-padded
+- Cartridge type ($0147) = $1B and region ($014A) = $E1
 
 Registers:
 
-$2000 write: Normal behavior, plus save written value in $2000 latch
-$1000 write: $A5 enables configure mode, $98 disables, and other values have no known effect
-$7000 write while configure mode is on: Copy $2000 latch to OR mask
+- $2000 write: Normal behavior, plus save written value in $2000 latch
+- $1000 write: $A5 enables configure mode, $98 disables, and other values have no known effect
```
avivace, line 34:
```suggestion
- $1000 write: $A5 enables configure mode, $98 disables it, and other values have no known effect
```

Also, did we ever decide if the british comma should be consistent through the document?

src/othermbc.md
```diff
@@ -40,70 +41,75 @@ nonsensical.
 
 The mapper does not support an outer bank for battery SRAM.
 
-To start a game, do the following in code run from RAM: Write $A5 to
-$1000, write game starting bank number to $2000, write any value to
-$7000, write $98 to $1000, write $01 to $2000 (so that 32K games
-work), jump to $0100.
+To start a game, do the following with code run from RAM:
```
avivace, line 44:
```suggestion
To start a game, execute the following steps with code running from RAM:
```

src/othermbc.md
```diff
@@ -40,70 +41,75 @@ nonsensical.
...
+consisting of a 74LS377 octal latch in addition to the ROM chip. Because
 the mapper consists of a single standard 74 series logic chip, it has
 two unusual properties:
 
-First, unlike a usual MBC, it switches the whole 32 kiB ROM area instead
-of just the $4000-$7FFF area. If you want to use the interrupt vectors
+First, unlike a usual MBC, it switches the whole 32 KiB ROM area instead
+of just the $4000-$7FFF area. Therefore, if you want to use [the interrupt vectors](<#Interrupt Handling>)
 with this cart, you should duplicate them across all banks.
 Additionally, since the initial state of the '377 can't be guaranteed,
```
avivace, line 62: What is the "initial state of the '377'? We are referring to the  74LS377 but it's not clear what the "initial state" consists of

src/othermbc.md
```diff
@@ -40,70 +41,75 @@ nonsensical.
...
 the ROM header and some code for switching to a known bank should also
 be included in every bank. This also means that the Wisdom Tree mapper
-could be used as a multicart mapper for 32 kiB ROMs, assuming there was
+could be used as a multicart mapper for 32 KiB ROMs, assuming there was
```
avivace, line 65:
```suggestion
can be used as a multicart mapper for 32 KiB ROMs, assuming there is
```
Why we switch to past tense here?

src/othermbc.md
```diff
@@ -40,70 +41,75 @@ nonsensical.
...
+    jp $7FFB
+```
+```rgbasm
+; At $7FFB in all banks
+    ld hl, $0100
+    ld [hl], a
+    jp hl
+```
+
+Second, because the '377 latches data on the *positive* write pulse edge,
```
avivace, line 82: Again '377, not a clear way to refer to the mapper.

avivace, line 82: I think a small reference to how positive-edge triggered latches should be linked (e.g. https://www.allaboutcircuits.com/textbook/digital/chpt-10/edge-triggered-latches-flip-flops/)

src/othermbc.md
```diff
@@ -40,70 +41,75 @@ nonsensical.
...
--   ROM contains "WISDOM TREE" or "WISDOM\\x00TREE" (the space can
-    be $20 or $00), $0147 = $00, $0148 = $00, size \> 32k. This
-    method works for the games released by Wisdom Tree, Inc.
--   $0147 = $C0, $014A = $D1. These are the values recommended by
-    beware for 3rd party developers to indicate that the ROM is
-    targeting Wisdom Tree mapper hardware. (See below.)
+- [ROM title](<#0134-0143 - Title>) is "WISDOM TREE" (the space may be a
+  $00 NUL character instead), $0147 = $00, $0148 = $00, size \> 32k.
+  This method works for the games released by Wisdom Tree, Inc.
+- $0147 = $C0, $014A = $D1. These are the values recommended by
```
avivace, line 96: "recommended by beware"  is not clear and such phrases should be avoided.
Who is "beware"? Is it even clear we are speaking of a person? 

"A standard proposed by BGB emulator proposes $0147 = $C0, $014A = $D1 to indicate this mapper as a target".

But even phrased this way. What standard? Where is the proposal? And under which assumptions it's a good choice? What's the rationale about the proposal? We should discuss and decide:

1. A place where to put those "proposals"
2. How to reference

src/othermbc.md
```diff
@@ -40,70 +41,75 @@ nonsensical.
...
+  beware for 3rd party developers to indicate that the ROM is
+  targeting Wisdom Tree mapper hardware. (See below.)
 
 ## Magic values for detection of multicarts in emulators
```
avivace, line 100: Why is there a chapter about detecting this type of multicart when it has just been stated how to detect it? This information should be merged

src/othermbc.md
```diff
@@ -40,70 +41,75 @@ nonsensical.
...
 
 Sometimes it may be useful to allow a ROM to be detected as a multicart
 in emulator, for example for development of a menu for physical
 multicart hardware. These are values suggested by beware, and supported
-in BGB, for signaling that your ROM is supposed to interface a multicart
+in at least BGB, for signaling that your ROM is supposed to interface a multicart
```
avivace, line 105: I'm not super sure stuff working only for a single emulator should be put with other information. There must be a visual way to tell we're explaining features of a single emulators (why?) and not documenting the mapper.

avivace (changes requested):
Overall looks great but the * Magic values for detection of multicarts in emulators* should be merged with previous information on detecting/being detected as a multicart.

As specified in the comments, "proposed standards" must be referenced and clearly exposed as behaviour of single emulators. 

I'm not even sure a particular feature of a single emulator is relevant here, to be quite fair. A link/reference should be enough and if we want to promote such proposals/standard we should discuss it in another place and then find a way to do so without mixing them up with actual hardware documentation.

src/othermbc.md
ISSOtm: I don't know, I just think it's more natural here because it's how you'd pronounce it. The comma here separates different propositions more than this is an enumeration.

src/othermbc.md
ISSOtm:
```suggestion
To start a game, do the following with code running from RAM:
```
or
```suggestion
To start a game, perform the following steps with code running from RAM:
```

Since all the steps are already formulated in imperative tense, I think "do the following" makes more sense than using the verb "execute", which implies code.

src/othermbc.md
ISSOtm: Should we use the full name everywhere?

src/othermbc.md
ISSOtm: Well, the warning box is already there, only the title needs to be added, then.

src/othermbc.md
ISSOtm: No idea.

src/othermbc.md
ISSOtm: Well, this should have been addressed when this info was merged; I'm not even touching this part.

src/othermbc.md
ISSOtm: On one hand there's documentation of the existing ROMs, and OTOH there's documentation for homebrew developers seeking to use those mappers (for any reason). There's also emulator developers, who still read Pan Docs, who may want to implement detection, and for whom documenting an agreed-upon heuristic is better than letting subtly incompatible variations proliferate.

I agree that it should be moved elsewhere as an actual standard, but this is not the scope of this PR.

src/othermbc.md
avivace: It wasn't merged. It was part of the baseline from where we started working (the wiki version):
https://gbdev.gg8.se/wiki/articles/Memory_Bank_Controllers#Magic_values_for_detection_of_multicarts_in_emulators

I think we move this issue in another PR as we are out of scope, but beware (lol) it'll be more time consuming to face those issues separatly (as we already pointed them down) and some of them are definitely blocking for this PR, so I'd honestly prefer switch this PR to draft and widen its scope.

src/othermbc.md
avivace: Whatever their scope, this information should be in the same place, and not splitted in "for emu dev" vs "for homebrew dev": it's the same suggestion and splitting it only creates confusion.

> It is proposed that homebrew targeting this mapper should [...]. 

Emulator developers can read this kind of information with no problems, and additional notes or info boxes can be appended to let them know additional/proposed information

src/othermbc.md
avivace: Second one works for me

src/othermbc.md
```diff
@@ -16,22 +16,23 @@ outputs, but this appears not to have been done in licensed games.
 
 ### EMS
 
-PinoBatch learned the game selection protocol for EMS flash carts from
-beware, who in turn learned it from nitro2k01. Take this with a grain of
-salt, as it hasn't been verified on the authentic EMS hardware.
+::: warning VERIFICATION NEEDED
```
avivace, line 19: We should decide on a consisted title for this kind of boxes. What about "Speculation"?

avivace (changes requested):

src/othermbc.md
ISSOtm: Or maybe follow Wikipedia and use "Verification needed"

PR #355, opened 2021-09-13
Fixes ppu_modes_timing.svg to more properly reflect the description b…

Kim-Dewelski:
…ox above. Also added extra detail in the description box found under Rendering -> Pixel FIFO to more accurately explain what is meant by a 'dot'.

### DIFF
--- a/src/imgs/ppu_modes_timing.svg
+++ b/src/imgs/ppu_modes_timing.svg
@@ -1,152 +1,307 @@
-<!-- Do not put any blank lines, or all hell will break loose! -->
-<svg viewBox="0 0 700 590" preserveAspectRatio="xMidYMid meet" xmlns="http://www.w3.org/2000/svg">
-	<defs>
-		<style type="text/css"><![CDATA[
-			text {
-				fill: var(--fg);
-			}
-			rect {
-				opacity: 75%;
-			}
-			.mode-box {
-				opacity: 25%;
-				height: 60px;
-			}
-			.scanline {
-				height: 20px;
-			}
-			.square {
-				height: 10px;
-				width: 10px;
-			}
-			.m2 { fill: #ffd966; }
-			.m3 { fill: #ff8000; }
-			.m0 { fill: #97d077; }
-			.m1 { fill: #00ffff; }
-			.m30 {
-				fill: url(#m30-gradient);
-			}
-			.darken {
-				fill: #000;
-				opacity: 50%;
-				font-weight: 500;
-			}
-			.centered { text-anchor: middle; }
-			.right    { text-anchor: end; }
-			.mode-id {
-				font-size: large;
-			}
-			.small {
-				font-size: small;
-			}
-			.middle-baseline {
-				dominant-baseline: middle;
-			}
-			.mode-id, .mode-descr {
-				font-variant: small-caps;
-			}
-		]]></style>
-		<linearGradient x1="0%" y1="0%" x2="100%" y2="0%" id="m30-gradient">
-			<stop offset="0%" stop-color="#ff8000"/>
-			<stop offset="100%" stop-color="#97d077"/>
-		</linearGradient>
-	</defs>
-	<rect class="mode-box m2" x="60" y="0" width="80"/>
-	<text class="centered mode-id" x="100" y="25">Mode 2</text>
-	<text class="centered mode-descr" x="100" y="45">OAM scan</text>
-	<text class="centered small" x="100" y="75">80 dots</text>
-	<text class="centered small" x="100" y="90">(= 20 cycles)</text>
-	<rect class="mode-box m3" x="145" y="0" width="172"/>
-	<text class="centered mode-id" x="231" y="25">Mode 3</text>
-	<text class="centered mode-descr" x="231" y="45">Drawing pixels</text>
-	<text class="centered small" x="231" y="75">172–289 dots</text>
-	<text class="centered small" x="231" y="90">(= 43–72 cycles)</text>
-	<rect class="mode-box m0" x="322" y="0" width="204"/>
-	<text class="centered mode-id" x="424" y="25">Mode 0</text>
-	<text class="centered mode-descr" x="424" y="45">Horizontal blank</text>
-	<text class="centered small" x="424" y="75">87–204 dots</text>
-	<text class="centered small" x="424" y="90">(= 22–51 cycles)</text>
-	<rect class="mode-box m1" x="130" y="490" width="230"/>
-	<text class="centered mode-id" x="245" y="515">Mode 1</text>
-	<text class="centered mode-descr" x="245" y="535">Vertical blank</text>
-	<!-- "Accessibility" spans -->
-	<text x="322" y="135">VRAM ($8000–9FFF) inaccessible</text>
-	<text x="322" y="155">CGB palettes inaccessible</text>
-	<path stroke="#ff8000" d="M 317,130
-	                          v 20
-	                          m 0,-10
-	                          h -172
-	                          m 0,10
-	                          v -20"/>
-	<text x="322" y="185">OAM inaccessible (except by DMA)</text>
-	<path stroke="#ffd966" d="M 317,170
-	                          v 20
-	                          m 0,-10
-	                          h -257
-	                          m 0,10
-	                          v -20"/>
-	<text class="centered" x="443" y="515">Everything</text>
-	<text class="centered" x="443" y="535">accessible</text>
-	<!-- Scanlines -->
-	<text class="middle-baseline right" x="55" y="222">LY = 0</text>
-	<rect class="scanline m2" y="210" x="60" width="80"/>
-	<rect class="scanline m3" y="210" x="145" width="172"/>
-	<rect class="scanline m0" y="210" x="322" width="204"/>
-	<text class="middle-baseline darken" y="222" x="150">SHORTEST</text>
-	<text class="middle-baseline darken right" y="222" x="525">LONGEST</text>
-	<text class="middle-baseline small" x="535" y="222">456 dots (= 114 cycles)</text>
-	<text class="middle-baseline right" x="55" y="247">1</text>
-	<rect class="scanline m2" y="235" x="60" width="80"/>
-	<rect class="scanline m3" y="235" x="145" width="289"/>
-	<rect class="scanline m0" y="235" x="439" width="87"/>
-	<text class="middle-baseline darken" y="247" x="150">LONGEST</text>
-	<text class="middle-baseline darken right" y="247" x="525">SHORTEST</text>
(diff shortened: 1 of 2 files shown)


src/pixel_fifo.md
```diff
@@ -2,9 +2,10 @@
 
 ::: tip TERMINOLOGY
 
-All references to a cycle are meant as T-cycles (4.19 MHz) and cycle
-counts are doubled on CGB in double speed mode. When it is stated that a
-certain action *lengthens mode 3* it means that mode 0 (hblank) is
+All references to a cycle or dot are meant as T-cycles (4.19 MHz) and cycle
+counts are doubled on CGB in double speed mode. Dots are the terminology used 
+for cycles in the context of rendering, and are synonymous with a T-cycle.
```
ISSOtm, line 7: Dots are not synonymous, as they stay at 4 MiHz even when double-speed causes T-cycles to be at 8 MiHz. Hence why a different unit is used.

This terminology issue generally overlaps with #194.

src/imgs/ppu_modes_timing.svg
```diff
@@ -1,152 +1,307 @@
...
-	<text class="middle-baseline small" x="540" y="510">4560 dots (= 1140 cycles)</text>
-	<!-- Totals -->
-	<text class="centered" y="20" x="613">One frame:</text>
-	<text class="centered" y="45" x="613">70224 dots</text>
-	<text class="centered" y="70" x="613">(= 17556 cycles)</text>
-	<text class="centered" y="95" x="613">@ 59.7 fps</text>
-</svg>
+<?xml version="1.0" encoding="UTF-8" standalone="no" ?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1" width="638" height="580" viewBox="0 0 638 580" xml:space="preserve">
```
ISSOtm, line 3: Without changing [the `include` directive](https://github.com/gbdev/pandocs/blob/eb4c695db1e67f8a999b243b26267225cf609256/src/pixel_fifo.md#L10), this will embed the `<!DOCTYPE>` tag in the document, which is at best harmless cruft, at worst will break *something*.

https://github.com/gbdev/pandocs/blob/eb4c695db1e67f8a999b243b26267225cf609256/src/pixel_fifo.md#L12

Please refer to [the SVG integration guide](https://github.com/gbdev/pandocs/wiki/SVG-guide#dynamically-styled-svgs) for more info.

src/imgs/ppu_modes_timing.svg
```diff
@@ -1,152 +1,307 @@
...
+<desc>Created with Fabric.js 3.6.3</desc>
```
ISSOtm, line 4: The SVG being redone from scratch causes some issues:

- [ ] Changes are quite hard to trace (`git blame`...); I cannot pinpoint what changes were made to the SVG to improve accuracy. What are they?
- [ ] A lot of properties are now duplicated all over the place, which increases the file size. (I'm not sure what the point of the `<g transform="">`s are, either.)
- [ ] The used fonts are now inconsistent with the rest of the document (Open Sans and Times New Roman vs. Inter):
  [Light-mode render](https://user-images.githubusercontent.com/15271137/133169039-7efd96fe-340d-4886-886a-4601451eecce.png)
- [ ] It also breaks theme integration, most egregiously on the dark "Ayu" mode:
  [Dark-mode render](https://user-images.githubusercontent.com/15271137/133168872-13474d8f-31ea-470b-9957-8e8522104e9b.png)

For these reasons, I think it would be best to port the changes mentioned in bullet 1 to the existing document.
(You don't have to do it yourself, if you don't want to edit the SVG manually I can do so myself.)

If you think it's necessary, I can also add more comments to the SVG, being the original author and all that :P

ISSOtm (changes requested):

src/imgs/ppu_modes_timing.svg
Kim-Dewelski: Ah great! I was gonna see myself on my machine what would happen but I wasn't able to get mdbook onto my machine. The only changes I made was remove the (= cycles) comments scattered around the SVG in favour of explaining that a dot is the same as a T-cycle and so to remove ambiguity, I decided it was best not to include the amounts of cycles passed when specifying dots count for each mode. Because the cycles were in M-cycles when, just before, a box of text said any further usage of "cycle/s" on the page will be in reference to T-cycles. 
I could try to manually edit the SVG myself as well, but I'm not able to locally host the website to try it out so perhaps it's best if you do so.

Kim-Dewelski: I'm rather new to this entire github pull thing so I hope I am not causing any issues! Haha. Anyway, I manually edited the SVG myself. Of course feel free to only apply the changes if you like them! I just found myself confused reading the docs that cycles were specified in M-cycles rather in T-cycles as was specified above. Thank you so much for your work!

aaaaaa123456789: If anything is

aaaaaa123456789: If anything is using T-states, they should be explicitly called T-states, not anything else.

ISSOtm: Well, T-cycles, but I agree.

src/pixel_fifo.md
ISSOtm: Really, then, the word "cycle" should be removed from the page and replaced with "dot", since it is really the unit being used throughout the document (just with a wrong name). Then this box can be simplified to just explaining what a "dot" is.

src/imgs/ppu_modes_timing.svg
aaaaaa123456789: > If anything is using T-states, they should be explicitly called T-states, not anything else.

Isn't "T-state" the official name used in documentation?

src/pixel_fifo.md
Kim-Dewelski: Overlooking the typo
'All references to a cycle are meant as T-cycles (4.19 MHz) and cycle counts are doubled on CGB in double speed mode. On DMG, a dot is equal to a single T-cycle. However, for CGB in double speed a dot is 2 T-cycles as to keep the amount of dots perscanline and frame similar regardless of processor speeds. When it is stated that a certain action lengthens mode 3 it means that mode 0 (hblank) is shortened to make up for the additional time in mode 3, as shown in the following diagram.'
Should the term cycle be omitted from here? Does it cause confusion over the meaning of a 'dot'?

src/imgs/ppu_modes_timing.svg
ISSOtm: Yes, but we don't use official names when they are confusing. Otherwise we'd be talking about "characters", not "tiles".

src/pixel_fifo.md
ISSOtm: T-cycles do not intervene in any of the document. T-cycles and M-cycles are bound to the CPU clock, which speeds up when double-speed mode is enabled.

Dots, however, stem from another clock which is *not* affected by double-speed mode. This is why they are a separate unit: because they are not relative to the CPU speed.

Kim-Dewelski: The problem comes then later down the line when T-cycles are used the measure the duration of, for instance, the the pixel fetcher's different steps.
'The fetcher fetches a row of 8 background or window pixels and queues them up to be mixed with sprite pixels. The pixel fetcher has 5 steps. The first four steps take 2 cycles each and the fifth step is attempted every cycle until it succeeds.'

I could gladly omit the reference to T-cycles and just instead give an overview of "dot", but I personally think it's beneficial to give the insight that, on a normal DMG, a single dot seems to equate to a sinlge T-cycle. Although again I could be very wrong on this and I don't have the specific clock rate for the PPU. I could completely omit the T-cycle parallel and just start explaining, and researching, the frequency of the PPU clock and only explain dots in via the PPU clock rather than any parallels to the CPU, if that's preferred? Either way, originally the diagram used to have M-cycles to explain the timings of the different modes. Or at least it seemed to be M-cycles or what was otherwise meant by the "(= cycles)" comments? I didn't want to omit any original information just clarify what was there.

ISSOtm: As [I explained above](https://github.com/gbdev/pandocs/pull/355#discussion_r707954068), the unit used throughout the document is really dots, *not* T-cycles.

avivace (changes requested):
A part from the mentioned suggestion, LGTM! Thanks @Kim-Dewelski

src/pixel_fifo.md
Kim-Dewelski: Ah my bad. I see now what you mean.

ISSOtm: Great! That's much better, but now there's a needlessly confusing alias. Could you try replacing all occurrences of "cycle" with "dot" in the document?

Kim-Dewelski: For sure! I'm on it!

Kim-Dewelski: should be fixed now! :)

src/pixel_fifo.md
```diff
@@ -2,9 +2,11 @@
...
 ::: tip TERMINOLOGY
 
-All references to a cycle are meant as T-cycles (4.19 MHz) and cycle
-counts are doubled on CGB in double speed mode. When it is stated that a
-certain action *lengthens mode 3* it means that mode 0 (hblank) is
+All references to a cycle are meant as T-cycles (4.19 MHz) and cycle counts are doubled 
+on CGB in double speed mode. On DMG, a dot is equal to a single T-cycle. However, for CGB 
+in double speed a dot is 2 T-cycles as to keep the amount of dots perscanline and frame similar
+regardless of processor speeds.
+When it is stated that a certain action *lengthens mode 3* it means that mode 0 (hblank) is
```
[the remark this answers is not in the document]
ISSOtm: @Kim-Dewelski Once something is taken care of, you can click the "Mark as resolved" button on a conversation ^^

avivace (approved):

ISSOtm:
Small word of update: I realized that some of the document's wording needed changing, and the scope crept into a rewrite of the entire article. Hope you won't mind—the changes already brought in are good, it's just that it's a good time as ever to do this.

avivace:
> Small word of update: I realized that some of the document's wording needed changing, and the scope crept into a rewrite of the entire article. Hope you won't mind—the changes already brought in are good, it's just that it's a good time as ever to do this.

Isn't that feasible in the second moment? Can we start merging this then iterate over later?

ISSOtm:
My usual concerns apply, but if it's like last time, then sure, okay to merge.

PR #356, opened 2021-09-14
README improvements

avivace:
Just a general overhaul of the README, mentioning more pointers for contributing and providing better examples on how to prepare the development environment to run mdbook.

### DIFF
--- a/README.MD
+++ b/README.MD
@@ -12,7 +12,7 @@ Everyone is welcome to contribute opening issues, expressing feedback, adding an
 
 Here are some resources you should take a look at:
 
-- The [Wiki](https://github.com/gbdev/pandocs/wiki), containing various rules and contribuition guidelines you'll need to follow when proposing changes. Some RFCs and discussions detailing the document scope are linked there too.
+- [The wiki](https://github.com/gbdev/pandocs/wiki), containing various rules and contribution guidelines you will have to follow when proposing changes. Some RFCs and discussions detailing the document scope are linked there as well.
 - The project [board](https://github.com/gbdev/pandocs/projects/1), for a general picture of the roadmap and the ongoing efforts.
 - [Issues](https://github.com/gbdev/pandocs/issues), where we discuss about what needs to be worked on and how.
 - The various community channels found [here](https://gbdev.io/chat.html) where we you can chat directly with maintainers and other contributors.
--- a/README.MD
+++ b/README.MD
@@ -34,7 +34,7 @@ To deploy Pan Docs locally:
   # Install python3 (on Debian systems)
   apt install python3
   ```
-2. [Install](https://packaging.python.org/tutorials/installing-packages/#requirements-files) the Python prerequisites. E.g.:
+2. [Install the Python prerequisites](https://packaging.python.org/tutorials/installing-packages/#requirements-files).
 ```sh
 # Create and activate a virtualenv
 python -m venv env
--- a/README.MD
+++ b/README.MD
@@ -14,7 +14,7 @@ Here are some resources you should take a look at:
 
 - [The wiki](https://github.com/gbdev/pandocs/wiki), containing various rules and contribution guidelines you will have to follow when proposing changes. Some RFCs and discussions detailing the document scope are linked there as well.
 - [The project board](https://github.com/gbdev/pandocs/projects/1), for a general picture of the roadmap and the ongoing efforts.
-- [Issues](https://github.com/gbdev/pandocs/issues), where we discuss about what needs to be worked on and how.
+- [The issues](https://github.com/gbdev/pandocs/issues), where we discuss what needs to be worked on, and how.
 - The various community channels found [here](https://gbdev.io/chat.html) where we you can chat directly with maintainers and other contributors.
 
 Once you feel confortable, fork this repository, make your modifications and send a Pull Request.
--- a/README.MD
+++ b/README.MD
@@ -1,28 +1,56 @@
 # Pan Docs
 
-> The single, most comprehensive technical reference to Game Boy available to the public.
+"Pan Docs" is a document started in early 1995, considered the single most comprehensive technical reference to Game Boy available to the public. 
 
-Go to https://gbdev.io/pandocs/ to read the document.
+This git repository hosts a renewed version of it, mantained in the Markdown format and enjoying renewed community attention.
+
+Go to [gbdev.io/pandocs](https://gbdev.io/pandocs) to read the document.
 
 ## Contributing 
 
-[This](https://github.com/gbdev/awesome-gbdev/issues/153) is the RFC proposing the change. You're welcome to discuss the development of this project in the Issues and in our various community channels found [here](https://gbdev.io/chat.html).
+Everyone is welcome to contribute opening issues, expressing feedback, adding and improving content or share new findings.
 
-Contributing is really easy, fork this repo and edit the files in the **src** directory. Then, you can send your PR.
+Here are some places you should take a look to:
+
+- The [Wiki](https://github.com/gbdev/pandocs/wiki), containing various rules and contribuition guidelines you'll need to follow when proposing changes. Some RFCs and discussions detailing the document scope are linked there too.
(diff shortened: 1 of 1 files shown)


README.MD
```diff
@@ -1,28 +1,56 @@
...
+### Deploy
 
 To deploy Pan Docs locally:
 
 1. Install [Rust](https://www.rust-lang.org/tools/install), [mdBook](https://github.com/rust-lang/mdBook#readme), and [Python 3](https://www.python.org/downloads).
-  mdBook powers the book itself, Rust is used for some custom plugins and Python scripts are used to render some images.
-2. [Install](https://packaging.python.org/tutorials/installing-packages/#requirements-files) the Python prerequisites (`pip3 install -r requirements.txt`)
-   If you're doing that in a Python virtual environment, be sure to activate it in the shell that will run the `mdbook` commands.
+  mdBook is the tool rendering the documentation, Rust is used for some custom plugins and Python scripts are used to render some images. E.g.:
+  ```sh
```
ISSOtm, line 28: We shouldn't duplicate 3rd-party install instructions. Linking to them is enough.

README.MD
```diff
@@ -1,28 +1,56 @@
...
+  # Install Rust using rustup
+  curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
+  # Install mdbook using cargo
+  cargo install mdbook
+  # Remember to add cargo bin directory to your path
+  # Install python3 (on Debian systems)
+  apt install python3
+  ```
+2. [Install](https://packaging.python.org/tutorials/installing-packages/#requirements-files) the Python prerequisites. E.g.:
```
ISSOtm, line 37:
```suggestion
2. [Install the Python prerequisites](https://packaging.python.org/tutorials/installing-packages/#requirements-files).
```

README.MD
```diff
@@ -104,13 +132,13 @@ Steps:
 E.g.
 
 ```bash
```
ISSOtm, line 134:
```suggestion
```console
```
And then we'd keep the dollar signs as prompts. Sounds good?

ISSOtm (changes requested):

README.MD
avivace: Why the dollar signs? The idea of these snippets is to lower the barrier to people wanting to contribute, these commands are always the same and just need to be copy pasted (which can't happen that easily if you have to copy paste line by line removing the dollar sign)..

README.MD
avivace: Why? It's 3 lines and helps people to know what to do without having to open 4 pages and look for instuctions in each of them, we are provinding as an example, just ways to quickstart..

README.MD
pinobatch: The dollar signs indicate something to be typed at a prompt (which in Bash usually defaults to something ends in `$ `). Is there a way to use CSS `:before` or similar to produce the dollar signs?

avivace: Not here, as we don't have access to such CSS rules in GitHub Markdown rendered.
Also, the idea of those snippets is to let people quickly copy-paste these commands. We don't have stuff that doesn't need to be typed at a prompt, everything is supposed to.

README.MD
ISSOtm: They will not necessarily stay correct, especially the first one. The second one is indeed likely to not change, but the third one is OS-dependent, too.

I'm against "downstreaming" such instructions because it's an extra thing to pay attention to, that upstream may change without notice (and be in their full rights to), and may not apply to one's setup, Windows especially.

The Python instructions force you to create a virtual environment: why? What does it do? Is it necessary? Does it require any extra steps to be available?

avivace: > I'm against "downstreaming" such instructions because it's an extra thing to pay attention to, that upstream may change without notice (and be in their full rights to), and may not apply to one's setup, Windows especially.

It's 3 lines and they just provide *an example* on how you accomplish the previous indications.. This is not supposed to be set in stone, just a suggestion to get people quickstarted..

It's not a big deal if they get changed upstream (which I seriously doubt, package names are not gonna change). We can provide instructions for Windows too, if you feel like it's in the intended audience (which I seriously doubt - see WSL).

> The Python instructions force you to create a virtual environment: why? What does it do? Is it necessary? Does it require any extra steps to be available?
Installing using a virtual environment is a generally accepted good practise allowing you to not touch your system, and deleting the cloned folder restores the system at the previous state (python-wise)

README.MD
pinobatch: Does this mean [Google developer documentation style guide article "Documenting command-line syntax" section "Command prompt"](https://developers.google.com/style/code-syntax#prompt) is a bad example?

> If your command-line instructions show multiple lines of input, then start each line of input with the `$` prompt symbol.
> [...]
> 👍 **Recommended:**  
> Enter the following code into the terminal:
> 
>     $ adb devices

avivace: > If your command-line instructions show **multiple** lines of input
We only have 3 lines total, and each input is one line only, with clear context and no need to clarify it (we don't use monospaced blocks for anything else).
That is only a guideline. In fact, if you keep reading the very same document, they don't use any `$` at  all.

README.MD
ISSOtm: > It's 3 lines and they just provide *an example* on how you accomplish the previous indications.. This is not supposed to be set in stone, just a suggestion to get people quickstarted..

I commend that goal, but I'm willing to bet that this is not how it will be used, and that it will be treated as gospel by most people, scanning through the README looking for code blocks. I don't think it's a good thing, but if you disagree, I won't make more of a fuss.

> Making them read the documentation of 3 different projects just to understand how to see a change in the text is crazy and will constitute a wall to contributors.

It's not reading three projects' documentation, the links are to the projects' respective install pages. The first one has the command in the middle of the screen, the third one has a big "download" button in the middle, and the second one is arguably the least obvious because it requires either clicking a link or scrolling down a bit (though we could fix that by linking to the `installation` anchor instead). I don't think that's an outlandish claim. Is it?

> It's not a big deal if they get changed upstream (which I seriously doubt, package names are not gonna change).

I'm mostly fine with the `apt` lines, though I disagree with the implicit dismissal of other distros (am I being petty? I hope not); the `curl | sh` one way less so. *That* one is likely to break in several ways, there are ways to install Rustup from one's package manager and then setup Rust, and I'm seriously uncomfortable with recommending people to just pipe a remote-gotten script directly into a shell.

About the Python virtual env: what's wrong with installing packages on the system? I install all of my Python packages through `pacman` instead of `pip`, for example. It's not explained either how to re-activate the virtual env for further development, for example.

avivace: > I don't think it's a good thing, but if you disagree, I won't make more of a fuss.
I disagree. This is not a software repository and it's our responsibility to make as easy as possible for people to contribute to the documentation. Since *we* chose to use a series of tools that require a number of steps to get started, we should provide a way, even if it doesn't provide any insight on the tool themselves (as it's not the focus at all of the contribution), to do so with as little as effort as possible.

Beware, this is true for the setting up steps, but not for the documentation guidelines, which in turn are actually put in place to read before even forking and sending a PR.

>  I don't think that's an outlandish claim. Is it?
Honestly, it is. Do you want people to scrim through other's project documentation just to understand how to deploy *parts* of what they need just to see their modification on a completely unrelated thing live? Do you want contributors to use their time in researching GB quirks or learn how to set up a rust environment?

> About the Python virtual env: 

1. Installing system-wise touches your system configuration and potentially breaks every other package or system-wide binary by messing up its dependencies. It's unrealistic to ask people to change their system-configuration (also requiring privileges) just to edit a row of a document..
2. It's explained how to activate it. Look for the commented line in the snippet of code.
3. Virtual environments 
    1. Are system agnostic
    2. Are Configuration agnostic
    3. Leave the system completely clean
    4. Will consistently lead the environment to be ready to run what we require. The only case ths doesn't happen is if you messed up your Python installation. There's literally no easiest and clean way to let people prepare their environment without wasting time in installing or configuring other tools. It's literally suggested in the Python documentation..

1 and 2 are not valid for your pacman thing and I don't see how your observation is even related when in the other thread you complained about the Debian instructions

PR #363, opened 2021-10-01
Style warning boxes

avivace:
Warning boxes were left out when we migrated to mdbook (#57 )

Mimicking how they were in Vuepress, the left border is a bit bigger. I also ejected the default variables.css from the default theme so I could add a couple of variables to style the titles and the text colors for the boxes. 

Here's how they look in the different themes:

image
image
image
image

Note: do not squash these commits, as there's one with the default version of the variables.css from upstream mdbook

### DIFF
--- a/custom/style.css
+++ b/custom/style.css
@@ -20,13 +20,6 @@ html {
     }
 }
 
-/* Override global variables */
-:root {
-    /* workaround to get hover+active sidebar links on normal links blue */
-    --sidebar-active: var(--links);
-}
-
-
 body {
     font-family: "Inter";
     /* Enable some font features for Inter (https://rsms.me/inter/#features/calt) */
@@ -76,12 +69,13 @@ code {
 .box {
     margin: 20px 0;
     padding: 1.6px 20px;
-    border-left-width: .5rem;
+    border-left-width: .75rem;
     border-left-style: solid;
 }
 
 .box>.box-title {
     font-weight: 600;
+    text-transform: uppercase;
 }
 
 .box.tip {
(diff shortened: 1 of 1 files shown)


ISSOtm:
Isn't it possible to do this by defining our own variables? That way, the rest of `variables.css` would still update automatically, at least.

ISSOtm:
While we're discussing styling of boxes, we could consider the typography of the titles (sse also https://github.com/gbdev/pandocs/pull/365#discussion_r720652755). Do we want to small-caps them?

avivace:
> Isn't it possible to do this by defining our own variables? That way, the rest of `variables.css` would still update automatically, at least.

This is what I'm doing, I didn't modify any of the other existing variables of the themes, but added a few ones

custom/style.css
```diff
@@ -90,12 +91,43 @@ code {
...
+    color: var(--c-warning-title);
 }
 
 .spacer {
     width: 100%;
     height: 3px;
     margin: 25px 0px 25px 0px;
 }
+
+:root {
```
ISSOtm, line 109: Is it better to put those at the top, or the bottom of the file?

avivace: I think bottom is best, I can add a comment describing the scope of that section if it's not clear enough.
E.g. "Global CSS variables", "Theme-relative styling overrides"

ISSOtm: Yeah, that'd be great!

ISSOtm (approved):

custom/style.css
avivace: Okay, I've expanded a bit adding some comments. I've also moved the previous variable variable override in `:root` there, specifying the "override" (vs the other stuff that's added)

PR #334, opened 2021-07-22
Switch to new "bit explanation" format

ISSOtm:
Fixes #318, https://github.com/gbdev/pandocs/issues/293

TODO:
- [ ] Center text in fields, for better readability (e.g. `CGB palette` in the OAM table)

PR #366, opened 2021-10-01
Import GBC Infrared Communication from Dan Docs

avivace:
I've copied the "GBC Infrared Communication" article from @shonumi 's Dan Docs.

Some notes:

1. I've converted the two tables, the first one needs reviewing and could be used as possible standard for the other registry-bit explanation ones
image
2. I have omitted a couple of paragraphs too specific to titles / excessively long lists.
3. The source is specified at the top in a "small" text:
   - How hard it is to put this fragment of text in a "partial" file and import it? So we don't duplicate it in every Dan Docs-imported article?
   - If this style is appreciated, we need to apply it to the similar notice in "Game Boy Printer" 
image
4. We need to decide if we should use GCB or CGB to refer to the Game Boy Color model.

### DIFF
--- a/src/IR.md
+++ b/src/IR.md
@@ -1,6 +1,6 @@
 # GBC Infrared Communication
 
-<small>This section was originally compiled by Shonumi in the "Dan Docs". Upstream source can be found [here](https://shonumi.github.io/dandocs.html).</small>
+{{#include partials/dandocs_notice.md}}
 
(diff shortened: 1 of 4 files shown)


ISSOtm:
This seems to overlap a lot with #327?

For point 1, please not that there's already #334 pending for the tables, with the style that was agreed upon after the attached issue's discussion.

For point 3, this is as simple as [using `{{#include}}`](https://rust-lang.github.io/mdBook/format/mdbook.html#including-files). We'd mostly just need first to figure out where to put the partial files in `src/`.

For point 4, there's probably a convention to be established, possibly using the two different terms in different ways, e.g. `GBC` = all color models (`CGB`, `AGB`, `GBS`), but `CGB` = strictly Color's.

PR #374, opened 2021-10-03
ci: add problemkaputt.de as excluded domain for the linkchecker

avivace:


ISSOtm (approved):
Makes CI pass again, ship it!

PR #379, opened 2021-10-15
Overhaul "Pixel FIFO" article into "Rendering Internals"

### LINKED ISSUE #377: Get Tile step during the FIFO needs correction / improvement
> If the current tile is a window tile, the X coordinate for the window tile is used, otherwise the following formula is used to calculate the X coordinate: ((SCX / 8) + fetcher’s X coordinate) & $1F. Because of this formula, fetcherX can be between 0 and 31.

fetcher's X and fetcherX are too similar and cause confusion. 

Suggestion : rename fetcher's X coordinate to X_tile_number or something similar.
Also provide a range, since it is not inferred from the variable name, i.e. 0 - 31 (fetcher's X advances once every 8 pixel fetch).

> If the current tile is a window tile, the Y coordinate for the window tile is used, otherwise the following formula is used to calculate the Y coordinate: (currentScanline + SCY) & 255. Because of this formula, fetcherY can be between 0 and 159.

currentScanline (same as LY?) ranges between 0 - 143 (visable), SCY ranges between 0 - 255, fetcherY therefore must be between 0 and 255 instead of 159.

Suggestion : change the range of fetcherY

### LINKED ISSUE #408: Contradictions in PPU FIFO size
In the documentation of the PPU FIFO it says the Gameboy has two 16-pixel FIFO's ("Each FIFO can hold up to 16 pixels"). But further down it says the fetcher will never push to FIFO unless the FIFO is *empty*, which means that the FIFO will never hold more than 8 pixels.

ISSOtm:
Also avoid describing SameBoy internals, instead relying on it when otherwise corroborated, or on schematics and/or test ROMs when possible.

Restructure the article to describe behavior more than components, especially in a way that is more friendly to someone not knowing what all the components are about.

Add a diagram, too, and move the mode timing diagram to the STAT article, where it belongs just as well, but where it will be more visible and thus more useful.

Fixes #377, fixes #408.


avivace (changes requested):

ISSOtm:
I rebased the branch and applied some editorial changes; the article is still not finished, and should not be reviewed yet. (At least not before #350 is merged, so we can focus on that.)

src/STAT.md
```diff
@@ -2,7 +2,7 @@
 
 ::: tip TERMINOLOGY
 
-A *dot* is the shortest period over which the PPU can output one pixel: is it equivalent to 1 T-state on DMG or on CGB single-speed mode or 2 T-states on CGB double-speed mode. On each dot during mode 3, either the PPU outputs a pixel or the fetcher is stalling the [FIFOs](<#Pixel FIFO>).
+A *dot* is the shortest period over which the PPU can output one pixel: is it equivalent to 1 T-state on DMG or on CGB single-speed mode or 2 T-states on CGB double-speed mode. On each dot during mode 3, either the PPU outputs a pixel or the fetcher is stalling the [FIFOs](<#Rendering Internals>).
```
avivace, line 5:
```suggestion
A *dot* is the shortest period of time over which the PPU can output one pixel: is it equivalent to 1 T-state on DMG or on CGB single-speed mode or 2 T-states on CGB double-speed mode. On each dot during mode 3, either the PPU outputs a pixel or the fetcher is stalling the [FIFOs](<#Rendering Internals>).
```

src/imgs/ppu_overview.svg
```diff
@@ -0,0 +1,181 @@
...
+  <text x="14" y="405">(X, Y) positions</text>
+  <rect x="5" y="415" class="legend attrs"/>
+  <text x="14" y="419">Attributes</text>
+  <rect x="5" y="429" class="legend tileid"/>
+  <text x="14" y="433">8-bit tile ID</text>
+  <rect x="90" y="415" class="legend tileofs"/>
+  <text x="99" y="419">(9-bit tile ID, Y offset)</text>
+  <rect x="90" y="429" class="legend pxrow"/>
+  <text x="99" y="433">Pixel row (2 bytes)</text>
+  <text x="465" y="405" class="right">(Some arrows have been</text>
```
avivace, line 50: I would put this text way smaller (and maybe in a angle?)

avivace (changes requested):
I just gave it a quick look.

Looks nice so far!

src/Rendering_Internals.md
```diff
@@ -0,0 +1,410 @@
...
+Both appear to be accessed during the same dot, which implies that the PPU can access both VRAM banks at the same time.
+
+#### OBJ fetcher
+
+Since the Pixel Slice Fetcher is normally used continuously for the BG FIFO, the OBJ fetcher waits to take control until two conditions are met:
+- The Pixel Slice Fetcher is attempting to [push pixels](<#Push pixels>)
+- The BG FIFO is not empty
+
+Once both conditions are fulfilled, the OBJ FIFO takes over, discarding the pixels slices already fetched.
+Note that if the BG FIFO is empty, the Pixel Slice Fetcher immediately switches to [Get tile ID](<#Get tile ID>) when refilling it, so the OBJ fetcher will wait for 6 additional dots.
```
calc84maniac, line 162: According to my understanding of the sprite timing, the maximum "penalty" for a sprite occurring too soon in the BG slice fetch sequence is 5 dots (making that sprite contribute at most 11 dots to mode 3). I'm not clear on the technical reason behind this, though.

This understanding is corroborated by https://gbdev.io/pandocs/STAT.html#properties-of-stat-modes and https://www.reddit.com/r/EmuDev/comments/59pawp/gb_mode3_sprite_timing/

src/Rendering_Internals.md
```diff
@@ -0,0 +1,410 @@
...
+[^banks_parallel]:
+Both appear to be accessed during the same dot, which implies that the PPU can access both VRAM banks at the same time.
+
+#### OBJ fetcher
+
+Since the Pixel Slice Fetcher is normally used continuously for the BG FIFO, the OBJ fetcher waits to take control until two conditions are met:
+- The Pixel Slice Fetcher is attempting to [push pixels](<#Push pixels>)
+- The BG FIFO is not empty
+
+Once both conditions are fulfilled, the OBJ FIFO takes over, discarding the pixels slices already fetched.
```
calc84maniac, line 161: What pixel slices are discarded? Are they refetched later, and when is there time for that? I assumed the BG fetching sequence latches the fetched slices somewhere until it's able to successfully push them to the BG FIFO (for which it always has to wait at least 2 dots anyway).

Kapoisu:
I'm going to propose some feedback about why I've had a hard time understanding this document from the point of view of a novice.

Especially in the Sprite part, there are terms like "X coordinate" which I'm not sure what they refer to. From the aspect of implementation, I reset the counter of the fetcher when I enter the window. Hence I can't compare the X coordinate of the sprite to it and it's reasonable to think that these terms refer to a counter which tracks the number of pixels actually shifted out and pushed to the LCD.

The assumption above is valid only if I keep as less states as possible. What if I had an extra counter which tracks the number of pixels already fetched? The timing of the sprite checking would be changed from "**after a pixel is popped**" to "**before a pixel is fetched**." This may not be true (to be honest I'm not that confident) but it's actually a reasonable consequence - because I have to do the sprite checking, I introduce this extra counter.

That is to say, I suppose the term is a little bit too brief because I can consider all of the counters mentioned above sorts of "X coordinate."

In the OBC fetcher part,

> the OBJ fetcher waits to take control until two conditions are met:
> - The Pixel Slice Fetcher is attempting to [push pixels](https://github.com/ISSOtm/pandocs/blob/rendering-internals/src/Rendering_Internals.md#Push%20pixels)
> - The BG FIFO is not empty

Firstly, what is the exact timing of "attempting to push pixels?" Is it the 6th, 7th, or 8th of the fetcher step? From my understanding, this would affect the number of cycles added to mode 3.

I've known that there may be delay up to 11 cycles for each sprite (from somewhere else). There are descriptions such as "shorten/lengthen mode 3 by n dots" spread across the document. My point is, they are not tidied up and summarized in an easy-to-understand way. How/Why these events contribute to the max 11-cycle delay isn't stated clearly.

Use [this page](https://gbdev.io/pandocs/pixel_fifo.html) as an example. 172-289 dots? Readers are not going to figure out 289 = 12 (before shifter is filled) + 160 (pixels per scanline) + 11 (sprite delay) * 10 (number of sprites) from pure imagination (and I don't even know why a possible window restart is not included). 

With all the ambiguities combined, I can't even have a strong guarantee of the correctness of the overall control flow.

src/Rendering_Internals.md
```diff
@@ -0,0 +1,408 @@
...
+- On DMG and on CGB in Non-CGB Mode, the algorithm is simply that only pixels with an ID of 0 (= transparent pixels) are overwritten.
+  Since OBJ pixels are inserted as the OBJs are encountered horizontally, pixels inserted earlier, thus from objects with lower X positions, will have priority.
+- On CGB in CGB Mode, OBJ pixels also store the ID of the OBJ they originated from, and get overwritten by pixels from OBJs with lower IDs.
+
+## Selector
+
+Every time both FIFOs are clocked, the selector decides whether to retain the pixel from the BG or OBJ FIFO.
+
+The selection follows the following rules:
+1. **In CGB Mode**, if [`LCDC` bit 0 (priority enable)](<#CGB Mode: BG and Window master priority>) is reset, pick the BG pixel.
```
calc84maniac, line 251: This rule appears to be simply disabling OBJ, which duplicates the functionality of rule 3 and doesn't seem correct. I believe a correct rule would be to ignore rules 4 and 5.

src/Rendering_Internals.md
```diff
@@ -0,0 +1,408 @@
...
+
+::: tip Wrapping
+
+The two additions in the "BG mode" row are carried out in 8 bits, i.e. modulo 256.
+Due the division by 8, the modulo is essentially 32, i.e. a tilemap's width.
+This is what causes the background to wrap around, both vertically and horizontally.
+
+:::
+
+A byte is read from the computed address, and is forwarded to the Pixel Slice Fetcher as a tile ID.
```
quinnyo, line 138:
```suggestion
A tile ID is read from the computed address, and is forwarded to the Pixel Slice Fetcher.
```

src/Rendering_Internals.md
```diff
@@ -0,0 +1,408 @@
...
+Color models read from the computed address twice: from VRAM bank 0 to get the tile ID, and from VRAM bank 1 to get the attributes.[^banks_parallel]
+
+::: tip Raster effects
+
+Interestingly, unlike e.g. the NES' PPU, great care has been taken to ensure that the BG fetcher re-reads as many registers as possible (`SCY`, `LCDC`, etc.).
+This may have been insight from the former console, on which [proper "raster splits" are quite tricky](https://www.nesdev.org/wiki/PPU_scrolling#Split_X_scroll) due to a lot of internal caching.
+
+:::
```
quinnyo, line 146: This is interesting but might be a bit out of scope?
It sounds like trivia, due to the framing with the NES PPU.

src/Rendering_Internals.md
```diff
@@ -0,0 +1,408 @@
...
+
+This step takes 2 dots, with the VRAM access(es) being performed on the second.
```
quinnyo, line 148: "this step" again -- is this different to the OBJ fetcher step?

quinnyo (changes requested):
I'm a bit of a noob with this part of the Game Boy, which is hopefully a good thing for someone reviewing this!
I've made a bunch of inline comments/suggestions that are all mostly writing/language based because I can't really correct technical points with much confidence.
Instead, I'll cover some general points and my issues with understanding the technical side here:

I can't get past "Pixel Slice Fetcher/Get Tile ID".
- the Get Tile ID + BG + OBJ fetcher nested headings are confusing
- The Pixel Slice Fetcher doesn't even do anything until Get Tile ID has been done?
- we're getting the Tile ID but we're also getting attributes and doing ... something with them?
- *after* the BG and OBJ fetchers have had a go, the Pixel Slice Fetcher does the actual pixel slice fetching -- which is *only* the Color ID part of the the tile?

Some of the points about FIFO refill priority and the related potential delays are repeated several times in slightly different ways. This makes the differences seem like they should be very meaningful but I can't discern what the meaning is.

The BG FIFO, BG Fetcher, BG Fetcher with Window Fetcher hat, OBJ FIFO, OBJ Fetcher terminology makes it really difficult to follow just because they all look/sound like the same words. Sometimes it seems like the two fetchers are *part of* their respective FIFO or vice versa.

I think I want a more clear distinction between procedural, architectural, conceptual information.
To be clear, the information mostly seems to be here, but the presentation sometimes jumps between these modes and I struggle to follow that. I'm sure others have both more and less difficulty with that than I do.

This is a really complicated thing you're trying to explain!

src/Rendering_Internals.md
```diff
@@ -0,0 +1,408 @@
...
+  - **On CGB**, a value between 0 and 7 (for BG, the palette bits from [BG attributes](<#BG Map attributes (CGB Mode only)>); for OBJ, the palette bits from [OAM attributes](<#Byte 3 — Attributes/Flags>)).
+- Source OBJ: only applies to the OBJ FIFO on CGB. This contains the ID of the OBJ the pixel originated from.
+- Priority:
+  - OBJ FIFO: holds the OBJ-to-BG Priority bit from [the OBJ's attributes](<#Byte 3 — Attributes/Flags>).
+  - BG FIFO, **on CGB only**: holds the OBJ-to-BG priority bit from [the tile's attributes](<#BG Map attributes (CGB Mode only)>).
+
+Every scanline, the following occurs in order:
+1. *Mode 2*:
+   1.  OAM is scanned for Y positions in range (based on [`LY`](<#FF44 — LY: LCD Y coordinate \[read-only\]>) and [`LCDC`](<#FF40 — LCDC: LCD control>)); the X coordinate is not checked!
+   2. The first 10 matches get their X and Y coordinates stored in an "OBJ slot".
```
nummacway, line 67: Isn't this different depending on whether the software is running on a DMG (including CGB with OPRI set) or a CGB?

src/Rendering_Internals.md
```diff
@@ -0,0 +1,408 @@
+# Rendering Internals
+
+The Game Boy's PPU is the component responsible for feeding the LCD (= the screen) with pixels.
+This document describes how the PPU renders pixels.
+
+::: tip Terminology
```
Estus-Dev, line 6: It appears that each instance of `tip` and `warning` have a leading space, thus breaking the parser and causing them to display incorrectly.

```suggestion
:::tip Terminology
```

src/Rendering_Internals.md
avivace: I also believe this may be a bit out of scope right there, especially the second "speculation" part.

src/Rendering_Internals.md
```diff
@@ -0,0 +1,408 @@
...
+
+A "dot" is the unit of time within the PPU.
+One "dot" is one 4 MiHz cycle, i.e. a unit of time equal to 1 ∕ 4194304 of a second.
```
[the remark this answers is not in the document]
avivace: Agree here

src/Rendering_Internals.md
calc84maniac: DMG and CGB have identical OAM scan behavior (only the Y coordinate is checked, and the first 10 in OAM order are matched) and OBJ fetch order (lower X first, or lower OAM index for the same X), the actual difference is how OBJ pixels are mixed in the FIFO.

DMG (including CGB with OPRI set) prioritizes all opaque pixels already in the OBJ FIFO when fetching a new OBJ, which means the fetch order (lower X etc.) determines the priority.

CGB prioritizes only the opaque pixels with a smaller OAM index than the new OBJ, which means the FIFO tracks the OAM index per pixel (or possibly the OBJ slot index, since those are in increasing OAM order and would allow cheaper FIFO storage and comparison hardware).

ecopinrox:
> Do you want to rewrite the "Mode 3 Operations" section or do you want to delete it? It seems kinda redundant.

This section can be renamed to something like "Pixel FIFO Operation", since the current title implies that the FIFO can operate in other modes as well. However, I think redistributing this section's information over the rest of the page would be a better course of action. 

I am very new to emulator development, and the biggest problem I have with the pan docs is that many pages reference information that is introduced only in later pages, providing no clear starting point. Related information is also spread across multiple pages and sections, which will surely cause inconsistencies.

src/Rendering_Internals.md
```diff
@@ -0,0 +1,408 @@
+# Rendering Internals
+
+The Game Boy's PPU is the component responsible for feeding the LCD (= the screen) with pixels.
+This document describes how the PPU renders pixels.
+
+:::tip Terminology
+
+A *dot* is the unit of time within the PPU.
+One "dot" is one 4 MiHz cycle, i.e. a unit of time equal to 1 ∕ 4194304 of a second.
+The duration of one "dot" is independent of [CGB double speed](<#FF4D — KEY1 (CGB Mode only): Prepare speed switch>).
```
quinnyo:
```suggestion
The duration of one dot is independent of [CGB double speed](<#FF4D — KEY1 (CGB Mode only): Prepare speed switch>).
```

src/Rendering_Internals.md
nummacway: Shouldn't it pick "OBJ if not transparent, otherwise pick BG" rather than BG in the first place? (If changing the wording, make sure "otherwise" cannot be thought to be related to LCDC bit 0 being reset.)

nummacway (commented):
Review probably not finished, but not much time right now.

src/Rendering_Internals.md
```diff
@@ -0,0 +1,408 @@
...
+   1.  OAM is scanned for Y positions in range (based on [`LY`](<#FF44 — LY: LCD Y coordinate \[read-only\]>) and [`LCDC`](<#FF40 — LCDC: LCD control>)); the X coordinate is not checked!
+   2. The first 10 matches get their X and Y coordinates stored in an "OBJ slot".
+      The fact that there are 10 such slots is why only 10 objects can be displayed per scanline[^more_than_10].
+      <!-- -->
+   This operation takes 2 dots per OBJ, which, multiplied by 40 OBJs in OAM, gives Mode 2's length of 80 dots.
+
+2. *Mode 3*:
+
+   During Mode 3, on each dot, the FIFOs are clocked, and one pixel output to the LCD.
+   The Pixel Slice Fetcher continuously runs in parallel to refill the BG FIFO.
```
quinnyo: "continuously" and "in parallel" is not entirely clear.
- [ ] something more like "runs as required to ensure the BG FIFO is supplied." ?

src/Rendering_Internals.md
```diff
@@ -0,0 +1,408 @@
...
+- Palette:
+  - **On DMG**, this contains the palette bit from [OAM attributes](<#Byte 3 — Attributes/Flags>). Of course, only the OBJ FIFO has this.
+  - **On CGB**, a value between 0 and 7 (for BG, the palette bits from [BG attributes](<#BG Map attributes (CGB Mode only)>); for OBJ, the palette bits from [OAM attributes](<#Byte 3 — Attributes/Flags>)).
+- Source OBJ: only applies to the OBJ FIFO on CGB. This contains the ID of the OBJ the pixel originated from.
+- Priority:
+  - OBJ FIFO: holds the OBJ-to-BG Priority bit from [the OBJ's attributes](<#Byte 3 — Attributes/Flags>).
+  - BG FIFO, **on CGB only**: holds the OBJ-to-BG priority bit from [the tile's attributes](<#BG Map attributes (CGB Mode only)>).
+
+Every scanline, the following occurs in order:
+1. *Mode 2*:
```
quinnyo:
```suggestion
1. *Mode 2 (OAM Scan)*:
```

src/Rendering_Internals.md
```diff
@@ -0,0 +1,408 @@
...
+Both appear to be accessed during the same dot, which implies that the PPU can access both VRAM banks at the same time.
+
+#### OBJ fetcher
+
+Since the Pixel Slice Fetcher is normally used continuously for the BG FIFO, the OBJ fetcher waits to take control until two conditions are met:
+- The Pixel Slice Fetcher is attempting to [push pixels](<#Push pixels>)
+- The BG FIFO is not empty
+
+Once both conditions are fulfilled, the OBJ FIFO takes over, discarding the pixels slices already fetched.
+Note that if the BG FIFO is empty, the Pixel Slice Fetcher immediately switches to [Get tile ID](<#Get tile ID>) when refilling it, so the OBJ fetcher will wait for 6 additional dots.
```
quinnyo: There doesn't seem to be a "base" duration mentioned in this section (OBJ Fetcher). What are these 6 dots in addition to?

quinnyo (changes requested):
Thank you for your efforts!
I've had a solid go at this now. I (re-)read through all the new material, from the top. The overall structure and information presented seems good. It made sense. 👌
Note (again) that I'm not an expert with PPU internals though, so I'm not the best fact checker for this material.

I added a bunch of suggested changes, which are mostly small things, and hopefully pretty self-explanatory.
That gets a bit noisy though, so I'll highlight a couple of more significant issues here:
- [ ] I think the "fetchers" should be renamed.
  - The overlap/ambiguity between "**fetcher**" and "Pixel Slice **Fetcher**" slowed me down a lot.
  - The "fetchers" seem to be tightly coupled to their associated FIFO(s?) and also responsible for a bit more than just retrieving data.
  - They seem a bit less like a "fetcher" and a bit closer to a "queue manager".
  - So my idea of a better name would be along those lines (manage, monitor, wrangle) -- but I'm sure someone that understands the topic better probably can come up with something better (and isn't completely wrong 🤷‍♀️)
- [ ] the term "FIFO" could be replaced with "queue" in many cases
  - It's tiring reading FIFO over and over again.
  - "queue" would be used to refer to the logical/conceptual, FIFO to refer to the physical/electronic.
  - With FIFO being the default term used, it stops being possible to use it to refer to specifically the electronic side of things.
  - May aid in dispelling any confusion about "real FIFOs"
- [ ] The structure of the headings and steps (under Pixel Slice Fetcher) could be improved.
  - [ ] The pair of 'get tile row' steps in particular are presented in a way that makes understanding a bit more difficult than it needs to be.
    - Merging the pair of them as two parts of the same step seems like a reasonable choice. Effectively having 3 steps instead of 4 would look less intimidating.
  - [ ] The step headings could have the step ordinals included, to aid navigation
  - [ ] The 'BG fetcher' and 'OBJ fetcher' subheadings under 'Get tile ID' is a bit awkward.
    - It was easy to lose track of the steps while reading these sections.
    - I don't have a good suggestion for this, unfortunately.

src/Rendering_Internals.md
```diff
@@ -0,0 +1,408 @@
...
+      <!-- -->
+   This operation takes 2 dots per OBJ, which, multiplied by 40 OBJs in OAM, gives Mode 2's length of 80 dots.
+
+2. *Mode 3*:
+
+   During Mode 3, on each dot, the FIFOs are clocked, and one pixel output to the LCD.
+   The Pixel Slice Fetcher continuously runs in parallel to refill the BG FIFO.
+   If the OBJ FIFO needs to be refilled, both FIFOs temporarily stop being clocked while the OBJ FIFO "steals" the Pixel Slice Fetcher to get its pixels.
+
+   Additionally, in the middle of the scanline, the window may be triggered; this is described in further detail [below](#).
```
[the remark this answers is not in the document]
avivace: Same as my previous comment.

src/Rendering_Internals.md
```diff
@@ -0,0 +1,408 @@
...
+The duration of one "dot" is independent of [CGB double speed](<#FF4D — KEY1 (CGB Mode only): Prepare speed switch>).
+
+When it is stated that a certain action *lengthens mode 3*, it implies that mode 0 (HBlank) is shortened to make up for the additional time spent in mode 3, as shown in [this diagram](<#STAT modes>).
+
+:::
+
+::: warning Timings caution
+
+Timings here are not tested by a single test ROM (made especially difficult by their resolution being finer than M-cycles).
+The information here was largely obtained from an emulator that passes `intr_2_mode0*` from [this test suite](https://github.com/wilbertpol/mooneye-gb/tree/b78dd21f0b6d00513bdeab20f7950e897a0379b3/tests/acceptance/gpu), but not all of it has been verified from e.g. [hardware schematics](https://github.com/furrtek/DMG-CPU-Inside).
```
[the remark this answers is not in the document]
avivace: https://github.com/furrtek/DMG-CPU-Inside is redirecting to https://github.com/msinger/dmg-schematics now

src/Rendering_Internals.md
```diff
@@ -0,0 +1,408 @@
...
+
+Since the Pixel Slice Fetcher is shared by both FIFOs, when both of them need to be refilled at the same time, pixels temporarily stop being output until both have been served.
+
+The Pixel Slice Fetcher is told which of the 384 tiles to fetch one slice from, as well as which slice of that tile's 8, by the "fetchers"—again, one for the background and window, another for OBJs.
+These "fetchers" also directly transmit some metadata to the FIFOs[^real_fifo_refilling], such as the palette, priority, etc.
+
+[^real_fifos]:
+Actually, there are more than 2 FIFOs.
+For example, on DMG, there are [two FIFOs for BG pixel indices](https://raw.githubusercontent.com/furrtek/DMG-CPU-Inside/master/Schematics/32_BG_PIXEL_SHIFTER.png), [two for OBJ pixels](https://raw.githubusercontent.com/furrtek/DMG-CPU-Inside/master/Schematics/33_SPRITE_PIXEL_SHIFTER.png), [one for OBJ palette bits](https://raw.githubusercontent.com/furrtek/DMG-CPU-Inside/master/Schematics/34_SPRITE_PALETTE_SHIFTER.png), and [one for OBJ-to-BG priority bits](https://raw.githubusercontent.com/furrtek/DMG-CPU-Inside/master/Schematics/26_BACKGROUND.png).
```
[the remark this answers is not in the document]
avivace: Yes please @quinnyo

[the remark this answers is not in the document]
quinnyo: These ones are not as simple as `sed 's@furrtek/DMG-CPU-Inside@msinger/dmg-schematics@g'`, unfortunately. This repo doesn't have prerendered schematics in it.
So I'm not sure yet what the solution is, but we probably shouldn't be linking to images like this anyway.
I'd say just keep the old ones for now, and work out a solution separately? The more important one was the link to the repo itself, which is fixed. It's not worth blocking the PR over.

PR #384, opened 2021-11-23
fixed memory range

K4stor:
In the current version it reads, "The areas from 0000-7FFF and A000-BFFF address external hardware on the cartridge" but I think the first range should be 4000-7FFF.

### DIFF
--- a/src/Memory_Map.md
+++ b/src/Memory_Map.md
@@ -40,7 +40,7 @@ area are required to be specified correctly.
 
 ## External Memory and Hardware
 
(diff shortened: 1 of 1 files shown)


ISSOtm:
No, the info is correct, it's just that external hardware *typically* doesn't bank 0000-3FFF. (Wisdom Tree is a counter-example.)

K4stor:
> No, the info is correct, it's just that external hardware _typically_ doesn't bank 0000-3FFF. (Wisdom Tree is a counter-example.)

Ah good to know, maybe we should add that info too, for noobs like me 😄

PR #389, opened 2022-01-07
Correction to a formula typo

AaronLip:
Fixes to a minor typo for clarity; the last frequency was repeated twice.


ISSOtm:
No, this is correct: the formula is recursive. The rest of the sentence's wording is probably what makes this confusing, namely mention of `X(0)`.

ISSOtm:
Ah, closing this is okay, but then it's possible others may be confused as you were. Would you like to open an issue about it, or make a PR that clears this up? You would likely do a better job than e.g. myself.

Issue #393, opened 2022-01-10
Set meta description per-page


avivace:
Currently, the description meta tag is set to a site-wide value, on every single exported HTML page:

```
<meta name="description" content="The single, most comprehensive technical reference to Game Boy available to the public.">
```

We need a proper couple of sentences, different for each page, to actually showcase what the page is about.

A couple of possible solutions:
   - Set description manually, per-page. This is the best as we can prepare curated "extracts". I'd like to be able to set it per-markdown file, maybe in a YAML front matter. See https://docs.rs/mdbook-frontmatter/0.0.4/mdbook_frontmatter/ can be useful
   - Unstrip the `{{{ content }}}` value of the HTML, ignore the headers and consider the first N (30, 40?) characters. Maybe even try to detect the first M (2, 3?) sentences.

ISSOtm:
Imo, an automated solution would fall flat too often to be worth the effort. A per-file override would work much better imo.

avivace:
> Imo, an automated solution would fall flat too often to be worth the effort. A per-file override would work much better imo.

Per-file override would be better, yes, but don't underestimate taking the first characters of the content. If you check https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fgbdev.io%2Fpandocs most of the autogenerated descriptions make sense (and they are just the first N words..)

ISSOtm:
I agree they do make sense. However, I was thinking that they don't make the most sense as general descriptions of their pages' contents, rather most of them dive directly into the nitty-gritty. No idea if that's better or worse.

PR #394, opened 2022-01-10
Add basic OG and Twitter meta tags

avivace:
This PR adds meta tags for

- og:site_name
- og:image
- og:title
- twitter:card
- twitter:site
- og:description

effictively making the content more social-media friendly.

A couple of notes:

- I had to eject the `index.hbs` handlebars page template from the default theme and then make modifications.
   - Don't squash this PR when merging so we maintain the first commit as the starting point of the template (upstream status for this version)
   - This detaches us from upstream. Any update on the handlebars template upstream will need to be ported over
- The `og:description` tag is set to the description, which is the same string for the entire site and it's not per-page. It's one of the most important values. I opened https://github.com/gbdev/pandocs/issues/393 to see if we can set it to a more sensible value (and actually related to the page content).


ISSOtm (approved):
As far as detaching from upstream goes, it would be nice if we could instead store a patch file and apply it against the bundled `index.hbs` somehow. But I don't see how to do that in a sane way, so this is the next best thing. We'll just have to keep it in mind when updating mdBook in the future.

PR #402, opened 2022-03-18
Fix formatting in GameShark docs

PikalaxALT:


src/Shark_Cheats.md
```diff
@@ -28,8 +28,8 @@ Check the [Game Genie manual](http://www.digitpress.com/library/manuals/gameboy/
 Game Shark codes consist of eight-digit hex numbers, formatted as
 ABCDEFGH, the meaning of the separate digits is:
 
-` AB    External RAM bank number`
-` CD    New Data`
+` AB    External RAM bank number` \
```
ISSOtm, line 31: Imo, this should be a table instead of pre-formatted text, for accessibility.

ISSOtm (changes requested):

src/Shark_Cheats.md
avivace: I agree, a table would be better here

avivace (approved):
This looks inactive so I'll merge it and open an issue about it (https://github.com/gbdev/pandocs/issues/403)

PR #405, opened 2022-03-24
Add SVG explanation of 2bpp format

sfiera:
I thought it would be nice to include a visual explanation of the 2bpp format inline, rather than linking elsewhere for it.

### DIFF
--- a/src/Tile_Data.md
+++ b/src/Tile_Data.md
@@ -80,7 +80,7 @@ both bytes, bit 7 represents the leftmost pixel, and bit 0 the rightmost. For
 example, the tile data `$3C $7E $42 $42 $42 $42 $42 $42 $7E $5E $7E $0A $7C $56
 $38 $7C` appears as follows:
 
-<figure align="center">
+<figure>
 {{#include imgs/sprite.svg:2:}}
 <figcaption>Sample tile data</figcaption>
 </figure>
--- a/src/Tile_Data.md
+++ b/src/Tile_Data.md
@@ -85,10 +85,11 @@ $38 $7C` appears as follows:
 <figcaption>Sample tile data</figcaption>
 </figure>
 
-In the first row, the values `$3C $7E` in binary are `00111100` and `01111110`.
-The leftmost bits are 0 and 0, thus the color ID is binary `00`, or 0. The next
-bits are 0 and 1, thus the color ID is binary `10`, or 2 (remember to flip the
-order of the bits!). The full eight-pixel row evaluates to 0 2 3 3 3 3 2 0.
+For the first row, the values `$3C $7E` are `00111100` and `01111110` in
+binary. The leftmost bits are 0 and 0, thus the color ID is binary `00`, or 0.
+The next bits are 0 and 1, thus the color ID is binary `10`, or 2 (remember to
+flip the order of the bits!). The full eight-pixel row evaluates to 0 2 3 3 3 3
+2 0.
 
 A tool for viewing tiles can be found
 [here](https://www.huderlem.com/demos/gameboy2bpp.html).
--- a/src/Tile_Data.md
+++ b/src/Tile_Data.md
@@ -80,9 +80,10 @@ both bytes, bit 7 represents the leftmost pixel, and bit 0 the rightmost. For
 example, the tile data `$3C $7E $42 $42 $42 $42 $42 $42 $7E $5E $7E $0A $7C $56
 $38 $7C` appears as follows:
 
-<div align="center">
+<figure align="center">
 {{#include imgs/sprite.svg:2:}}
-</div>
+<figcaption>Sample tile data</figcaption>
+</figure>
 
 In the first row, the values `$3C $7E` in binary are `00111100` and `01111110`.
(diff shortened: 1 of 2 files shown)


avivace:
Looks very nice! Cool!

avivace (approved):

src/Tile_Data.md
```diff
@@ -74,19 +74,22 @@ Byte 2-3  Second Line
...
-To obtain the color ID for the leftmost pixel,
-you take bit 7 of both bytes: 0, and 0. Thus the index is binary 00 = 0. For
-the second pixel, repeat with bit 6: 1, and 0. Thus the index is binary 01 =
-1 (remember to flip the order of the bits!). If you repeat the
-operation you'll find that the IDs for the eight pixels are 0 1 2 3 0 3
-3 1.
+For each line, the first byte specifies the least significant bit of the color
+ID of each pixel, and the second byte specifies the most significant bit. In
+both bytes, bit 7 represents the leftmost pixel, and bit 0 the rightmost. For
+example, the sprite \$3C \$7E \$42 \$42 \$42 \$42 \$42 \$42 \$7E \$5E \$7E \$0A
```
ISSOtm, line 80: Should the raw hex data be put in a monospaced font, without prefixes? That's how it's most often presented.

src/Tile_Data.md
```diff
@@ -74,19 +74,22 @@ Byte 2-3  Second Line
...
+\$7C \$56 \$38 \$7C appears as follows:
 
-A more visual explanation can be found
+<div align="center">
```
ISSOtm, line 83: That doesn't look right. Try a `figure` instead?

src/Tile_Data.md
```diff
@@ -74,19 +74,22 @@ Byte 2-3  Second Line
...
+{{#include imgs/sprite.svg:2:}}
+</div>
+
+In the first row, the values \$3C \$7E in binary are 00111100 and 01111110. The
```
ISSOtm, line 87:
```suggestion
For the first row, the values \$3C \$7E are `00111100` and `01111110` in binary. The
```

src/Tile_Data.md
```diff
@@ -74,19 +74,22 @@ Byte 2-3  Second Line
...
+leftmost bits are 0 and 0, thus the color ID is binary 00, or 0. The next bits
+are 0 and 1, thus the color ID is binary 10, or 2 (remember to flip the order
+of the bits!). The full eight-pixel row evaluates to 0 2 3 3 3 3 2 0.
+
+A tool for viewing tiles can be found
 [here](https://www.huderlem.com/demos/gameboy2bpp.html).
```
ISSOtm, line 93: We should additionally refactor this link to contain a more complete description (if only for accessibility, and also because wider links = more convenient to click)

ISSOtm (changes requested):
Awesome! You even followed the guidelines for themed SVGs, so, thank you *very* much!
I'll check out the render in a bit, though.

ISSOtm:
Here is what the SVG looks like in context:

image

src/Tile_Data.md
sfiera: I added the monospace but kept the $. I see arguments either way on the $: with it, it’s consistent with the rest of the explanation below; without it, it can be pasted directly into the linked tool.

src/Tile_Data.md
sfiera: I switched to <figure>, though I’m not sure what you meant by “doesn’t look right”.

src/Tile_Data.md
Rangi42: Aside: having consistent base prefixes on the numbers themselves ("binary %10, or decimal 2", or even 0b10, not just "binary 10") would make it easier to read the correct value right away.

ISSOtm: I agree, but it's been voted against and is not what the style guide says.

src/Tile_Data.md
ISSOtm: Non-semantic/presentational HTML, more or less.

src/Tile_Data.md
sfiera: Resolving this comment as it doesn’t pertain to a change in this PR.

src/Tile_Data.md
```diff
@@ -74,19 +74,24 @@ Byte 2-3  Second Line
...
-operation you'll find that the IDs for the eight pixels are 0 1 2 3 0 3
-3 1.
+For each line, the first byte specifies the least significant bit of the color
+ID of each pixel, and the second byte specifies the most significant bit. In
+both bytes, bit 7 represents the leftmost pixel, and bit 0 the rightmost. For
+example, the tile data `$3C $7E $42 $42 $42 $42 $42 $42 $7E $5E $7E $0A $7C $56
+$38 $7C` appears as follows:
 
-A more visual explanation can be found
+<figure align="center">
```
ISSOtm, line 83:
```suggestion
<figure>
```

ISSOtm (changes requested):

ISSOtm (approved):

PR #416, opened 2022-06-06
fetcherY can be up to 255, not 159

TheThief:


avivace (approved):

avivace:
@TheThief do we have some source / test rom for this?

TheThief:
> @TheThief do we have some source / test rom for this?

It's a simple documentation error. FetcherY = (scanline +SCY) %256

SCY can be 0-255. Therefore so can fetcherY.

PR #417, opened 2022-06-08
Clarifications to the explanation of interrupts

ihonen:
I noticed that some of the language used in the explanation of interrupts was vague and not technically entirely correct, so I thought I'd contribute a revision.

For example:

- an interrupt is not executed, it is serviced
- the interrupt vector is not the same thing as the interrupt handler
- etc.

Mostly rather insignificant, I know, but I couldn't help my pedantic self.

### DIFF
--- a/src/Interrupts.md
+++ b/src/Interrupts.md
@@ -19,7 +19,7 @@ RETI   ; Enables interrupts and returns (same as the instruction sequence EI, RE
 <ISR>  ; Disables interrupts and calls the interrupt handler
 ```
 
-Where \<ISR\> is the interrupt service routine that is automatically executed
+Where \<ISR\> is the Interrupt Service Routine that is automatically executed
 by the CPU when it services an interrupt request.
 
 The effect of `ei` is delayed by one instruction. This means that `ei`
--- a/src/Interrupts.md
+++ b/src/Interrupts.md
@@ -16,11 +16,11 @@ the following instructions/events only:
 EI     ; Enables interrupts  (that is, IME=1)
 DI     ; Disables interrupts (that is, IME=0)
 RETI   ; Enables interrupts and returns (same as the instruction sequence EI, RET)
-<INT>  ; Disables interrupts and calls interrupt vector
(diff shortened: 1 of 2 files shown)


avivace (approved):
Thanks a lot! I really like the "execution -> servicing" change.

src/Interrupts.md
```diff
@@ -16,14 +16,14 @@ the following instructions/events only:
 EI     ; Enables interrupts  (that is, IME=1)
 DI     ; Disables interrupts (that is, IME=0)
 RETI   ; Enables interrupts and returns (same as the instruction sequence EI, RET)
-<INT>  ; Disables interrupts and calls interrupt vector
+<ISR>  ; Disables interrupts and calls the interrupt handler
 ```
 
-Where \<INT\> means the operation which is automatically executed by the
-CPU when it executes an interrupt.
+Where \<ISR\> is the interrupt service routine that is automatically executed
```
ISSOtm, line 22:
```suggestion
Where \<ISR\> is the Interrupt Service Routine that is automatically executed
```

src/Memory_Map.md
```diff
@@ -25,8 +25,8 @@ The following addresses are supposed to be used as jump vectors:
 -   Interrupts: 0040, 0048, 0050, 0058, 0060
 
 However, this memory area (0000-00FF) may be used for any other purpose in case that your
-program doesn't use any (or only some) RST instructions or interrupts. RST
-is a 1-byte instruction that works similarly to the 3-byte CALL instruction, except
+program doesn't use any (or only some) `rst` instructions or interrupts. `rst`
```
ISSOtm, line 28:
```suggestion
program doesn't use any (or only some) [`rst`](https://rgbds.gbdev.io/docs/v0.5.2/gbz80.7/#RST_vec) instructions or interrupts. `rst`
```

ISSOtm (changes requested):
Thank you! This is much better.

ihonen:
A bit late with my response, but glad you liked it!

PR #428, opened 2022-07-24
Harmonize section titles

ISSOtm:
Currently, section titles are a bit of a mess, with inconsistent formatting from document to document. I wish to harmonize them, which first means agreeing on some rules.

I have proposed some rules, let's see if they can make consensus. (Likely not, obviously. :stuck_out_tongue:) Once something is agreed upon, I will apply the change to all existing headers.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,13 +53,11 @@ References:
 
 ### 2. Units and prefixes:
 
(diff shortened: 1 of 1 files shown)


avivace (approved):

avivace:
Great, thanks @ISSOtm . Sorry for being late on this one

PR #422, opened 2022-06-12
Revise the explanation of the cartridge header

ihonen:
I clarified, reworded and reformatted a bunch of stuff.

PR #429, opened 2022-07-24
Revise the explanation of the cartridge header (continued)

avivace:
Follow up of https://github.com/gbdev/pandocs/pull/422

### DIFF
--- a/src/Power_Up_Sequence.md
+++ b/src/Power_Up_Sequence.md
@@ -206,7 +206,7 @@ Rely on it at your own risk.
 ### Common remarks
 
 The console's WRAM and HRAM are random on power-up.
-[Different models tend to exhibit different patterns](https://twitter.com/CasualPkPlayer/status/1409752977812852736?s=19), but they are random nonetheless, even depending on factors such as the ambient teperature.
+[Different models tend to exhibit different patterns](https://twitter.com/CasualPkPlayer/status/1409752977812852736), but they are random nonetheless, even depending on factors such as the ambient temperature.
 Besides, turning the system off and on again has proven reliable enough [to carry over RAM from one game to another](https://www.youtube.com/watch?v=xayxmTLljr8), so it's not a good idea to rely on it at all.
 
 Emulation of uninitialized RAM is inconsistent: some emulators fill RAM with a constant on startup (typically $00 or $FF), some emulators fully randomize RAM, and others attempt to reproduce the patterns observed on hardware.
--- a/src/The_Cartridge_Header.md
+++ b/src/The_Cartridge_Header.md
@@ -1,70 +1,79 @@
 # The Cartridge Header
 
-An internal information area is located at $0100-014F in each cartridge.
-It contains the following values:
+Each cartridge contains a header, located at the address range `$0100`—`$014F`.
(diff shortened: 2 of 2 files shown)


avivace:
See points in https://github.com/gbdev/pandocs/pull/422#pullrequestreview-1048684969

ihonen:
I incorporated the changes suggested in #422. I also went through the entire document one more time and fixed whatever inconsistencies and such I could find. I squashed all the changes suggested in earlier reviews into a single commit, so there are now 3 commits in this PR: my original commit, the changes suggested in #422 and my latest clean-up commit.

Regarding the supposedly premature merging of the previous PR, I'll point out that I don't think that is a fair assessment as there were in fact no more unaddressed issues _at the time_ @ISSOtm did the merge, so it's hardly his fault.

Regardless, feel free to conduct any further reviewing, but as far as I am concerned this PR is ready for merge.

P.S. I'm not sure if the changes suggested by @ISSOtm in #422 to `src/Power_Up_Sequence.md` are relevant to this PR, so I did not include those (yet).

ihonen:
I incorporated the changes suggested in https://github.com/gbdev/pandocs/pull/422. I also went through the entire document one more time and fixed whatever inconsistencies and such I could find, as well as fixed the broken links. The changes suggested in earlier reviews have now been squashed into a single commit.

Regarding the supposedly premature merging of the previous PR, I'll point out that I don't think that is a fair assessment as there were in fact no more unaddressed issues at the time @ISSOtm did the merge, so it's hardly his fault.

Regardless, feel free to conduct any further reviewing, but as far as I am concerned this PR is ready for merge.

avivace (commented):
Thanks a lot for sticking with this @ihonen !

It looks great now, I guess we just need to revert the `-` -> `—` changes, as they also break links..

ihonen:
> It looks great now, I guess we just need to revert the - -> — changes, as they also break links..

I suppose you mean that we would have to fix every link that references one of the changed headers? Because I did that, at least to the best of my abilities, by searching for every instance of the old title in the source and replacing it with the new one.

It's no big deal to undo the change if you want me to, but I did my best to make the new titles work as well, so whichever you prefer.

avivace:
> > It looks great now, I guess we just need to revert the - -> — changes, as they also break links..
> 
> I suppose you mean that we would have to fix every link that references one of the changed headers? Because I did that, at least to the best of my abilities, by searching for every instance of the old title in the source and replacing it with the new one.
> 
> It's no big deal to undo the dash change if you want me to, but I did my best to make the new titles work as well, so whichever you prefer. I just thought en dashes looked better, and also it's the correct dash to use in ranges (such as address ranges), so I went for it.

The issue here is that changing dashes changes and breaks all the links outside of our control, and it's a change that should be discussed in a different PR. If we decide to switch to another dash, we should apply that globally to the entire document. 

Let's bring back the PR on the scope and open another one to decide about the titles

ihonen:
That's a good point about external links.

Should I undo all other changes in the header titles as well? For example, I changed `Mask ROM Version number` -> `Mask ROM Version Number` (to be fair, I think that's the only non-dash change).

ihonen:
Alright, I undid the title changes. Feel free to merge. :)

PR #433, opened 2022-09-15
Added preliminary HuC-3 documentation.

cuavas:
This adds preliminary documentation for the HuC-3 combination memory controller, real-time clock, tone generator and infrared communication controller.  Most of the information is already collected in the [cited Gameboy Development Forum thread](https://gbdev.gg8.se/forums/viewtopic.php?id=744) as well as various emulator sources.

I also edited the HuC1 page a little after looking at the contributing guidelines:
* Use `$` as prefix for hexadecimal numbers
* Use no prefix for hexadecimal addresses
* Use en dash rather than hyphen to denote ranges
* Hyphenate “read-only” and “write-only”

The MBC5 page change just wraps a line that stood out as excessively long at 72 columns like the rest of the file seems to use.

### DIFF
--- a/src/HuC1.md
+++ b/src/HuC1.md
@@ -1,10 +1,9 @@
 # HuC1
 
-HuC1 is an MBC used by some Game Boy games which, besides doing the
-usual MBC stuff, also provides infrared communication. A lot of sources
-on the internet said that HuC1 was “similar to MBC1”, but they didn’t
-provide any detail. I took a look, and it turns out that HuC1 differs
-from MBC1 quite a lot.
+HuC1 is an MBC developed by Hudson Soft. It implements ROM and RAM
+banking, and also provides infrared communication. Many sources on the
+internet said that HuC1 was “similar to MBC1”, without providing any
+detail. However, the HuC1 differs from MBC1 significantly.
 
 ## Memory Map
 
@@ -18,12 +17,13 @@ Address range | Feature
 
 ### 0000–1FFF — IR Select [write-only]
 
-Most MBCs let you disable the cartridge RAM to prevent accidental
-writes. HuC1 doesn’t do this, instead you use this register to switch
-the A000–BFFF region between “RAM mode” and “IR mode” (described
-below). Write $0E to switch to IR mode, and anything else to switch to
-RAM mode. Nevertheless some HuC1 games attempt to write $0A and $00 to
-this region as if it would enable/disable cart RAM.
+Most MBCs can disable the cartridge RAM to prevent accidental writes.
+HuC1 doesn’t do this. Instead, this register swtiches the $A000–BFFF
+region between “RAM mode” and “IR mode” (described below). Write $0E to
+switch to IR mode, or anything else to switch to RAM mode.
+
+Some HuC1 games still write $0A and $00 to this region as if it would
+enable/disable cart RAM.
 
 ### 2000–3FFF — ROM Bank Number [write-only]
 
@@ -44,9 +44,10 @@ safely ignore these writes.
 When in “IR mode” (wrote $0E to $0000), the IR register is visible
 here. Write to this region to control the IR transmitter. $01 turns it
 on, $00 turns it off. Read from this region to see either $C1 (saw
-light) or $C0 (did not see light). When in “RAM mode” (wrote
-something other than $0E to $0000) this region behaves like normal cart
-RAM.
+light) or $C0 (did not see light).
+
+When in “RAM mode” (wrote something other than $0E to $0000) this region
+behaves like normal cart RAM.
 
 ## External links
 
--- a/src/HuC1.md
+++ b/src/HuC1.md
@@ -1,50 +1,51 @@
 # HuC1
 
-HuC1 is an MBC used by some Game Boy games which besides doing the usual
-MBC stuff, also provides IR comms. A lot of sources on the internet said
-that HuC1 was \"similar to MBC1\", but they didn\'t provide any detail.
-I took a look, and it turns out that HuC1 differs from MBC1 quite a lot.
+HuC1 is an MBC used by some Game Boy games which, besides doing the
+usual MBC stuff, also provides infrared communication. A lot of sources
+on the internet said that HuC1 was \"similar to MBC1\", but they didn\'t
+provide any detail. I took a look, and it turns out that HuC1 differs
+from MBC1 quite a lot.
 
 ## Memory Map
 
 Address range | Feature
 --------------|------------------------------------
-  $0000-1FFF  | RAM/IR select (when writing only)
-  $2000-3FFF  | ROM bank select (when writing only)
-  $4000-5FFF  | RAM bank select (when writing only)
-  $6000-7FFF  | Nothing?
-  $A000-BFFF  | Cart RAM or IR register
+  $0000–1FFF  | RAM/IR select (when writing only)
+  $2000–3FFF  | ROM bank select (when writing only)
+  $4000–5FFF  | RAM bank select (when writing only)
+  $6000–7FFF  | Nothing?
+  $A000–BFFF  | Cart RAM or IR register
 
-### 0000-1FFF IR Select (Write Only)
+### 0000–1FFF IR Select (Write-Only)
 
 Most MBCs let you disable the cartridge RAM to prevent accidental
 writes. HuC1 doesn\'t do this, instead you use this register to switch
-the A000-BFFF region between \"RAM mode\" and \"IR mode\" (described
-below). Write 0x0E to switch to IR mode, and anything else to switch to
-RAM mode. Nevertheless some HuC1 games attempt to write 0x0A and 0x00 to
+the A000–BFFF region between \"RAM mode\" and \"IR mode\" (described
+below). Write $0E to switch to IR mode, and anything else to switch to
+RAM mode. Nevertheless some HuC1 games attempt to write $0A and $00 to
 this region as if it would enable/disable cart RAM.
 
-### 2000-3FFF ROM Bank Number (Write Only)
+### 2000–3FFF ROM Bank Number (Write-Only)
 
 HuC1 can accept a bank number of at least 6 bits here.
 
-### 4000-5FFF RAM Bank Select (Write Only)
+### 4000–5FFF RAM Bank Select (Write-Only)
 
 HuC1 can accept a bank number of at least 2 bits here.
 
-### 6000-7FFF Nothing? (Write Only)
+### 6000–7FFF Nothing? (Write-Only)
 
 Writes to this region seem to have no effect. Even so, some games do
 write to this region, as if it had the same effect as on MBC1. You may
 safely ignore these writes.
 
-### A000-BFFF Cart RAM or IR register (Read/Write)
+### A000–BFFF Cart RAM or IR register (Read/Write)
 
-When in \"IR mode\" (wrote 0x0E to 0x0000), the IR register is visible
-here. Write to this region to control the IR transmitter. 0x01 turns it
-on, 0x00 turns it off. Read from this region to see either 0xC1 (saw
-light) or 0xC0 (did not see light). When in \"RAM mode\" (wrote
-something other than 0x0E to 0x000) this region behaves like normal cart
+When in \"IR mode\" (wrote $0E to 0000), the IR register is visible
+here. Write to this region to control the IR transmitter. $01 turns it
+on, $00 turns it off. Read from this region to see either $C1 (saw
+light) or $C0 (did not see light). When in \"RAM mode\" (wrote
+something other than $0E to 0000) this region behaves like normal cart
 RAM.
 
 ## External links
--- a/src/HuC1.md
+++ b/src/HuC1.md
@@ -1,9 +1,8 @@
 # HuC1
 
 HuC1 is an MBC developed by Hudson Soft. It implements ROM and RAM
-banking, and also provides infrared communication. Many sources on the
-internet said that HuC1 was “similar to MBC1”, without providing any
-detail. However, the HuC1 differs from MBC1 significantly.
+banking, and also provides infrared communication.
+Despite many sources on the internet claiming that HuC1 is “similar to MBC1”, it actually differs from MBC1 significantly.
 
 ## Memory Map
 
--- a/src/HuC1.md
+++ b/src/HuC1.md
@@ -2,7 +2,7 @@
 
 HuC1 is an MBC used by some Game Boy games which, besides doing the
 usual MBC stuff, also provides infrared communication. A lot of sources
-on the internet said that HuC1 was \"similar to MBC1\", but they didn\'t
+on the internet said that HuC1 was “similar to MBC1”, but they didn’t
 provide any detail. I took a look, and it turns out that HuC1 differs
 from MBC1 quite a lot.
 
@@ -16,36 +16,36 @@ Address range | Feature
   $6000–7FFF  | Nothing?
   $A000–BFFF  | Cart RAM or IR register
 
-### 0000–1FFF IR Select (Write-Only)
+### 0000–1FFF — IR Select [write-only]
 
 Most MBCs let you disable the cartridge RAM to prevent accidental
-writes. HuC1 doesn\'t do this, instead you use this register to switch
-the A000–BFFF region between \"RAM mode\" and \"IR mode\" (described
+writes. HuC1 doesn’t do this, instead you use this register to switch
+the A000–BFFF region between “RAM mode” and “IR mode” (described
 below). Write $0E to switch to IR mode, and anything else to switch to
 RAM mode. Nevertheless some HuC1 games attempt to write $0A and $00 to
 this region as if it would enable/disable cart RAM.
 
-### 2000–3FFF ROM Bank Number (Write-Only)
+### 2000–3FFF — ROM Bank Number [write-only]
 
 HuC1 can accept a bank number of at least 6 bits here.
 
-### 4000–5FFF RAM Bank Select (Write-Only)
+### 4000–5FFF — RAM Bank Select [write-only]
 
 HuC1 can accept a bank number of at least 2 bits here.
 
-### 6000–7FFF Nothing? (Write-Only)
+### 6000–7FFF — Nothing? [write-only]
 
 Writes to this region seem to have no effect. Even so, some games do
 write to this region, as if it had the same effect as on MBC1. You may
 safely ignore these writes.
 
-### A000–BFFF Cart RAM or IR register (Read/Write)
+### A000–BFFF — Cart RAM or IR register [read/write]
 
-When in \"IR mode\" (wrote $0E to 0000), the IR register is visible
+When in “IR mode” (wrote $0E to $0000), the IR register is visible
 here. Write to this region to control the IR transmitter. $01 turns it
 on, $00 turns it off. Read from this region to see either $C1 (saw
-light) or $C0 (did not see light). When in \"RAM mode\" (wrote
-something other than $0E to 0000) this region behaves like normal cart
+light) or $C0 (did not see light). When in “RAM mode” (wrote
+something other than $0E to $0000) this region behaves like normal cart
 RAM.
 
 ## External links
--- a/src/HuC3.md
+++ b/src/HuC3.md
@@ -1,0 +1,171 @@
+# HuC-3
+
+HuC-3 is an MBC used by some Game Boy games which, besides doing the
+usual MBC stuff, also provides a real-time clock, speaker, and infrared
+communication. The CR2025 coin cell is user-replaceable. It is a
+successor to the [HuC1](./HuC1.md).
+
+The HuC-3 is poorly understood. Observed behavior suggests the
+real-time clock and tone generator are implemented using a 4-bit
+microcontroller core with internal program ROM.
+
+
+## Memory
+
+### 0000–3FFF – ROM Bank 00 (Read-Only)
+
+Contains the first 16 KiB of the ROM.
+
+### 4000–7FFF – ROM Bank 00–7F (Read-Only)
+
+This area may contain any of the further 16 KiB banks of the ROM. It is
+unknown whether bank $00 can be selected here.
+
+### A000–BFFF – RAM Bank 00–03, or RTC/IR register (Read/Write)
+
+Depending on the current register selection and RAM Bank Number (see
+below), this memory space is used to access an 8 KiB external RAM Bank,
+or a single I/O Register.
+
+
+## Memory Control Registers
+
+### 0000–1FFF – RAM/RTC/IR Select (Write-Only)
+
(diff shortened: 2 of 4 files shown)


src/HuC1.md
```diff
@@ -1,50 +1,51 @@
...
 
-### A000-BFFF Cart RAM or IR register (Read/Write)
+### A000–BFFF Cart RAM or IR register (Read/Write)
 
-When in \"IR mode\" (wrote 0x0E to 0x0000), the IR register is visible
-here. Write to this region to control the IR transmitter. 0x01 turns it
-on, 0x00 turns it off. Read from this region to see either 0xC1 (saw
-light) or 0xC0 (did not see light). When in \"RAM mode\" (wrote
-something other than 0x0E to 0x000) this region behaves like normal cart
+When in \"IR mode\" (wrote $0E to 0000), the IR register is visible
```
ISSOtm, line 44:
```suggestion
When in \"IR mode\" (wrote $0E to $0000), the IR register is visible
```

src/HuC1.md
```diff
@@ -1,50 +1,51 @@
...
+here. Write to this region to control the IR transmitter. $01 turns it
+on, $00 turns it off. Read from this region to see either $C1 (saw
+light) or $C0 (did not see light). When in \"RAM mode\" (wrote
+something other than $0E to 0000) this region behaves like normal cart
```
ISSOtm, line 48:
```suggestion
something other than $0E to $0000) this region behaves like normal cart
```

src/HuC3.md
```diff
@@ -0,0 +1,171 @@
...
+### 4000–5FFF – RAM Bank Select (Write-Only)
+
+HuC-3 can accept a bank number of at least 2 bits here.
+
+### 6000–7FFF – Nothing? (Write-Only)
+
+Games write $01 here on startup, but it doesn\'t seem to have any
+observable effect.
+
+
```
ISSOtm, line 63:
```suggestion
```

src/HuC3.md
```diff
@@ -0,0 +1,171 @@
...
+---------|------------------------------------------------------------
+  00–06  | Output or input space for reading or writing current time
+  10–12  | Minute of day counter (rolls over at 1440)
+  13–15  | Day counter
+  26     | Tone selection in two least significant bits
+  27     | Tone generator enabled if value is 1 (all bits checked)
+  58–5A  | Event time minutes
+  5B–5D  | Event time days
+
+
```
ISSOtm, line 168:
```suggestion
```

src/MBC5.md
```diff
@@ -29,8 +29,10 @@ Same as for MBC1, except that RAM sizes are 8 KiB, 32 KiB and 128 KiB.
 Mostly the same as for MBC1. Writing $0A will enable reading and
 writing to external RAM. Writing $00 will disable it.
 
-Actual MBCs actually enable RAM when writing any value whose bottom 4 bits equal $A (so $0A, $1A, and so on), and disable it when writing anything else.
-Relying on this behavior is not recommended for compatibility reasons.
+Actual MBCs actually enable RAM when writing any value whose bottom 4
+bits equal $A (so $0A, $1A, and so on), and disable it when writing
+anything else. Relying on this behavior is not recommended for
+compatibility reasons.
```
ISSOtm, line 35:
```suggestion
Actual MBCs actually enable RAM when writing any value whose bottom 4 bits equal $A (so $0A, $1A, and so on), and disable it when writing anything else.
Relying on this behavior is not recommended for compatibility reasons.
```
Text should not be wrapped at 80 columns (this is a hold-over that we haven't done a pass on); one line per sentence is preferred, as it keeps diffs clearer: a change does not reflow all text following it, muddying `git blame`.

src/HuC1.md
```diff
@@ -1,50 +1,51 @@
...
-  $6000-7FFF  | Nothing?
-  $A000-BFFF  | Cart RAM or IR register
+  $0000–1FFF  | RAM/IR select (when writing only)
+  $2000–3FFF  | ROM bank select (when writing only)
+  $4000–5FFF  | RAM bank select (when writing only)
+  $6000–7FFF  | Nothing?
+  $A000–BFFF  | Cart RAM or IR register
 
-### 0000-1FFF IR Select (Write Only)
+### 0000–1FFF IR Select (Write-Only)
```
ISSOtm, line 19: I appreciate the effort, but the section title does not follow [the style guide](https://github.com/gbdev/pandocs/blob/master/CONTRIBUTING.md#8-section-titles) anyway, so I am currently preparing a PR to fix that. (If you can follow it in the new docs, that would be very appreciated!)

src/HuC3.md
```diff
@@ -0,0 +1,171 @@
...
+HuC-3 is an MBC used by some Game Boy games which, besides doing the
+usual MBC stuff, also provides a real-time clock, speaker, and infrared
+communication. The CR2025 coin cell is user-replaceable. It is a
+successor to the [HuC1](./HuC1.md).
+
+The HuC-3 is poorly understood. Observed behavior suggests the
+real-time clock and tone generator are implemented using a 4-bit
+microcontroller core with internal program ROM.
+
+
```
ISSOtm, line 12:
```suggestion
```

src/HuC3.md
```diff
@@ -0,0 +1,171 @@
...
+This area may contain any of the further 16 KiB banks of the ROM. It is
+unknown whether bank $00 can be selected here.
+
+### A000–BFFF – RAM Bank 00–03, or RTC/IR register (Read/Write)
+
+Depending on the current register selection and RAM Bank Number (see
+below), this memory space is used to access an 8 KiB external RAM Bank,
+or a single I/O Register.
+
+
```
ISSOtm, line 30:
```suggestion
```

src/HuC3.md
```diff
@@ -0,0 +1,171 @@
...
+### 6000–7FFF – Nothing? (Write-Only)
+
+Games write $01 here on startup, but it doesn\'t seem to have any
+observable effect.
+
+
+## I/O Registers
+
+These registers are accessed by reading or writing int the A000–BFFF
```
ISSOtm, line 66:
```suggestion
These registers are accessed by reading or writing in the $A000–BFFF
```

src/HuC3.md
```diff
@@ -0,0 +1,171 @@
...
+ready to receive a command, or low when the RTC MCU is busy.
+
+Writing with the least significant bit clear requests that the RTC MCU
+execute the last command written to register $B.
+
+### E – IR (Read/Write)
+
+Similar to the equivalent register of the HuC-1. The least significant
+bit is used for infrared communication.
+
```
ISSOtm, line 100:
```suggestion
```

src/HuC3.md
```diff
@@ -0,0 +1,171 @@
...
+
+## RTC Communication Protocol
+
+The HuC-3 chip provides read and write access to a 256-nybble window
+into the RTC MCU\'s internal memory. Multi-nybble values are stored
```
ISSOtm, line 105: Is the backslash necessary?

src/HuC3.md
```diff
@@ -0,0 +1,171 @@
...
+
+* Write $0D to 0000 (select RTC Semaphore register)
+* Poll A000 until least significant bit is set (wait until ready)
+* Write $0B to 0000 (select RTC Command/Argument register)
+* Write command and argument to A000
+* Write $0D to 0000 (select RTC semaphore register)
+* Write $FE to A000 (clear semaphore, requesting MCU execute command)
+* Poll A000 until least significant bit is set (wait for completion)
+* Write $0C to 0000 (select RTC Command/Response register)
+* Read A000 and use value from 4 least significant bits
```
ISSOtm, line 120:
```suggestion
* Write $0D to $0000 (select RTC Semaphore register)
* Poll $A000 until least significant bit is set (wait until ready)
* Write $0B to $0000 (select RTC Command/Argument register)
* Write command and argument to $A000
* Write $0D to $0000 (select RTC semaphore register)
* Write $FE to $A000 (clear semaphore, requesting MCU execute command)
* Poll $A000 until least significant bit is set (wait for completion)
* Write $0C to $0000 (select RTC Command/Response register)
* Read $A000 and use value from 4 least significant bits
```

ISSOtm (changes requested):
Thanks for the doc! Considering this is called "preliminary", are you planning to expand this doc further?

src/HuC1.md
cuavas: Well, I took the following from CONTRIBUTING.md to heart:

> To prevent clutter, don't use a prefix for hex numbers when it's clear from the context that a number is hexadecimal. For example, addresses and lists of opcodes. In those cases, zero-pad, even for smaller numbers: `0000-3FFF` instead of `0-3FFF`.

I padded addresses to four hexadecimal digits and didn’t add a prefix, as addresses are implied to be hexadecimal.  Should this rule be updated/clarified?

src/HuC3.md
cuavas: I’ve intentionally left two blank lines before every level 2 heading (`##` prefix) besides the first to make the sections visible when editing the source in a text editor.

src/HuC3.md
cuavas: Once again, I was following the rule I quoted from CONTRIBUTING.md that says to omit the radix prefix for addresses.

cuavas:
> Considering this is called "preliminary", are you planning to expand this doc further?

Depends on how motivated I am to try and reverse-engineer the behaviour of a microcontroller embedded in a chip with very little to go on.  The fact is, knowledge about the HuC-3 is very limited at the moment.

src/HuC1.md
ISSOtm: Fair that you're following the rules; I think that addresses in running text like this should get a hex prefix at least for consistency ("0E" is arguably unambiguous), and the prefix helps it stand out from other words.

src/HuC3.md
cuavas: I saw it in use in other markdown documents in your repository (e.g. HuC1.md).  Presumably it’s to prevent automatic conversion of foot/inch marks to typographical quotes.  Should I just use typographical quote characters?

ISSOtm:
That's perfectly fine, I was asking only to know whether to mark this PR as a draft :)

src/HuC3.md
ISSOtm: Yeah, those should be more appropriate here. I think we have some backslashes left over from the MediaWiki conversion, but in general, I'd rather not have them if they aren't necessary.

src/MBC5.md
cuavas: > Text should not be wrapped at 80 columns (this is a hold-over that we haven't done a pass on); one line per sentence is preferred, as it keeps diffs clearer: a change does not reflow all text following it, muddying git blame.

Interesting.  We’ve made the opposite decision for MAME’s documentation.  We found that diffs were borderline impossible to deal with when paragraphs weren’t hard wrapped in the source.  I’ll back out this part of the change.

src/HuC3.md
ISSOtm: AFAIK, that is not valid Markdown (and I have seen parsers act incorrectly in that case), so I'd rather avoid it.

src/HuC3.md
cuavas: OK, I’ll fix it.

src/MBC5.md
ISSOtm: If paragraphs aren't wrapped at all, yes, it's a nightmare. This is why I think that "one line per sentence" (which I somewhat borrowed from man pages, where periods at EOL are treated as sentence end markers) is a good compromise.

src/HuC3.md
cuavas: You know, I really hate markdown because every single implementation of it is different – it’s the opposite of a standard.  I’ll remove them here.

cuavas:
OK:
* Changed headings to use em dash separator between address and description, square brackets around read/write usage information, and lowercase for read/write usage information.
* Removed double blank lines.
* Fixed the stray “t”.
* Added radix prefix to addresses in prose.
* Backed out MBC5.md change.
* Replaced backslash-escaped foot/inch marks with typographical quotes/apostrophes.

I still don’t know exactly what this is actually going to look like in the end, as every flavour of markdown is handled differently.  I’m mostly familiar with Sphinx RST (as used in MAME’s documentation), and the GitHub and Reddit flavours.

src/HuC3.md
ISSOtm: Well, that's what [CommonMark](https://commonmark.org) attempts to fix, but bugs are legion.

cuavas: > Well, that's what [CommonMark](https://commonmark.org/) attempts to fix, but bugs are legion.

I’m not even sure it’s “bugs” as such – the different dialects are evolving in different directions for whatever reason.

For example GitHub markdown doesn’t require a blank line to introduce a list, treats unadorned newlines as hard line breaks, and so in.  It seems to be an intentional break with the other flavours.

RST is also evolving in its own direction, with its own syntax for links, anchors, invoking macros, etc.  It seems to be more influenced by the TeX/LaTeX family of document markup languages.  We end up needing to lean pretty heavily on some of its features for the MAME documentation.

Anyway, that’s all kind of beside the point.  All the differences just make me nervous about writing anything in markdown when it’s going to be rendered by an unfamiliar engine.

cuavas:
> Are you unable to set up mdBook locally? I can upload a render myself if that's the case.

It’s a matter of being thin on the ground.  I have a constant backlog of MAME work, as well as real life™, and I’m severely sleep deprived.  I simply don’t have the energy to set up mdBook locally.

The GitHub workflows seemed to want to upload to your main site, so I didn’t try enabling them on my fork, because I don’t want to hit you with security alerts when it fails to push out the updated.

ISSOtm:
The GitHub workflow has a safeguard so that it only updates the website when ran as a push, to `master`, on `gbdev/pandocs`. Don't worry, we already had a few issues with *my* fork way back when. :P

Anyway, [I manually rendered the current state here](http://eldred.fr/pandocs/HuC3.html).

ISSOtm:
> I still don’t know exactly what this is actually going to look like in the end, as every flavour of markdown is handled differently. I’m mostly familiar with Sphinx RST (as used in MAME’s documentation), and the GitHub and Reddit flavours.

Note that [GH-flavored MarkDown](https://github.github.com/gfm/) is a couple of extensions over CommonMark, which [mdBook's parser](https://lib.rs/pulldown-cmark) implements, plus a few GitHub extensions.

I'll review tomorrow, I must ensure I don't become sleep-deprived myself ^^'

ISSOtm (approved):

src/HuC1.md
```diff
@@ -1,50 +1,51 @@
 # HuC1
 
-HuC1 is an MBC used by some Game Boy games which besides doing the usual
-MBC stuff, also provides IR comms. A lot of sources on the internet said
-that HuC1 was \"similar to MBC1\", but they didn\'t provide any detail.
-I took a look, and it turns out that HuC1 differs from MBC1 quite a lot.
+HuC1 is an MBC used by some Game Boy games which, besides doing the
+usual MBC stuff, also provides infrared communication. A lot of sources
+on the internet said that HuC1 was “similar to MBC1”, but they didn’t
+provide any detail. I took a look, and it turns out that HuC1 differs
```
avivace, line 6:
```suggestion
provide any detail. The HuC1 differs
```

Since we're here already, those kind of comments/remarks should be removed

src/HuC3.md
```diff
@@ -0,0 +1,166 @@
+# HuC-3
+
+HuC-3 is an MBC used by some Game Boy games which, besides doing the
```
avivace, line 3:
```suggestion
HuC-3 is an MBC used by some Game Boy games. Besides doing the
```

src/HuC3.md
```diff
@@ -0,0 +1,166 @@
...
+usual MBC stuff, also provides a real-time clock, speaker, and infrared
```
avivace, line 4:
```suggestion
usual MBC stuff, it also provides a real-time clock, speaker, and infrared
```

src/HuC3.md
```diff
@@ -0,0 +1,166 @@
...
+Value | Feature
+------|-----------------------------
+  $0  | Cart RAM (read-only)
+  $A  | Cart RAM (read/write)
+  $B  | RTC command/argument (write)
+  $C  | RTC command/response (read)
+  $D  | RTC semaphore (read/write)
+  $E  | IR (read/write)
+
+If any other value is written, reads from $A000–BFFF return open bus
```
avivace, line 45: Do we use "open bus values" anywhere else in the document? Or do we refer to them as rubbish etc? @ISSOtm

src/HuC3.md
```diff
@@ -0,0 +1,166 @@
...
+
+When reading, the least significant bit is high when the RTC MCU is
+ready to receive a command, or low when the RTC MCU is busy.
+
+Writing with the least significant bit clear requests that the RTC MCU
+execute the last command written to register $B.
+
+### $E — IR [read/write]
+
+Similar to the equivalent register of the HuC1. The least significant
```
avivace, line 95: Maybe we can link to the "equivalent register of HuC1" here?

src/SUMMARY.md
```diff
@@ -68,6 +68,7 @@
   - [MMM01](./MMM01.md)
   - [M161](./M161.md)
   - [HuC1](./HuC1.md)
+  - [HuC-3](./HuC3.md)
```
avivace, line 71: Maybe we could have a similar styling as HuC 1?
```suggestion
  - [HuC3](./HuC3.md)
```

avivace (changes requested):
Wow, thanks a lot @cuavas ! This looks already mostly merge-able, just left some comments

src/HuC1.md
cuavas: This page is basically copied from a jrra.zone blog post (hopefully with permission): http://jrra.zone/blog/huc1.html

I wasn’t editing it for style, I don’t know whether the license of the blog post would even allow that.  There’s no clear copyright notice/license on the original blog post, so Berne Convention automatic copyright/all rights reserved applies.

src/HuC3.md
cuavas: Why be less explicit?  “Rubbish” could be interpreted as meaning the chip drives the bus with an unpredictable value.  “Open bus” describes what actually happens with the hardware.

src/HuC3.md
cuavas: There’s already a link the the HuC1 page at the top.  It isn’t described very well there, anyway.

src/SUMMARY.md
cuavas: The silkscreen on the HuC1 says “HuC1A” while the silkscreen on the HuC-3 says “HuC-3”.  The difference in punctuation is intentional.

src/HuC1.md
ISSOtm: All content under Pan Docs has been licensed to the public domain, so that should be fine.

src/HuC3.md
ISSOtm: I'm not sure we use the term anywhere else, but yes, it's the correct term to use.

src/HuC1.md
cuavas: I tightened up the wording in a few places. here.

src/HuC3.md
cuavas: I edited this paragraph to read better.

ISSOtm (approved):

src/HuC1.md
```diff
@@ -1,51 +1,53 @@
 # HuC1
 
-HuC1 is an MBC used by some Game Boy games which besides doing the usual
-MBC stuff, also provides IR comms. A lot of sources on the internet said
-that HuC1 was \"similar to MBC1\", but they didn\'t provide any detail.
-I took a look, and it turns out that HuC1 differs from MBC1 quite a lot.
+HuC1 is an MBC developed by Hudson Soft. It implements ROM and RAM
+banking, and also provides infrared communication. Many sources on the
+internet said that HuC1 was “similar to MBC1”, without providing any
```
avivace, line 5: I would replace this with (starting with "Many sources on the", from the previous line)

```
Unlike generally believed, the HuC1 differs from MBC1 significantly.
```

avivace (changes requested):

src/HuC1.md
cuavas: But that reads poorly in English – if you really want to rework it in that direction, it should be “Contrary to popular belief…”  But then you’re not explaining where the popular belief came from (i.e. these repeated claims on various web sites).

Anyway, if you want to endlessly argue about wording like this, count me out.  I have more important things to do.  I just wanted to get write up my findings on the HuC-3 while it was fresh in my mind after doing reverse-engineering and emulation work on it.

ISSOtm: I don't think the "many sources on the Internet" part belongs in Pan Docs at all, which would solve the problem.

> Anyway, if you want to endlessly argue about wording like this, count me out. I have more important things to do. I just wanted to get write up my findings on the HuC-3 while it was fresh in my mind after doing reverse-engineering and emulation work on it.

Considering that all of the technical tidbits have been approved by both of us, I'd entirely understand if you decline to participate in the grammar/presentation bikeshedding; we can make phrasing changes ourselves. Regardless, thank you *very much* for your research, and for caring enough to reshare it. Know that it's appreciated.

src/HuC1.md
```diff
@@ -1,51 +1,53 @@
 # HuC1
 
-HuC1 is an MBC used by some Game Boy games which besides doing the usual
-MBC stuff, also provides IR comms. A lot of sources on the internet said
-that HuC1 was \"similar to MBC1\", but they didn\'t provide any detail.
-I took a look, and it turns out that HuC1 differs from MBC1 quite a lot.
+HuC1 is an MBC developed by Hudson Soft. It implements ROM and RAM
+banking, and also provides infrared communication. Many sources on the
```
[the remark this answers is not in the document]
ISSOtm: I disagree, the two consecutive "and"s would be confusing.

src/HuC1.md
cuavas: > I don't think the "many sources on the Internet" part belongs in Pan Docs at all, which would solve the problem.

Well, it gives historical context for JRRA’s comments, and it’s a sad fact that there are still many outdated websites that misleadingly say the HuC1 and HuC-3 are “similar to MBC1” or “similar to MBC5”.  It’s another thing carried over from the original blog post.

src/HuC1.md
```diff
@@ -1,51 +1,53 @@
...
+internet said that HuC1 was “similar to MBC1”, without providing any
+detail. However, the HuC1 differs from MBC1 significantly.
```
ISSOtm, line 6:
```suggestion
banking, and also provides infrared communication.
Despite many sources on the internet claiming that HuC1 is “similar to MBC1”, it actually differs from MBC1 significantly.
```
Is this a reasonable compromise?

cuavas: Yeah, that reads OK.

avivace (approved):

PR #436, opened 2022-09-17
Apply new section name style

ISSOtm:
Enforce #428


avivace (approved):
Looks good to me! I didn't check any anchor link, but if those are good we can merge.

ISSOtm:
I spent a lot of time eliminating them, the only remaining warnings need a tooling upgrade to be fixed, so merging then.

Issue #322, opened 2021-07-15
Allow SVG to be affected by theme


avivace:
https://stackoverflow.com/questions/18434094/how-to-style-svg-with-external-css
https://rust-lang.github.io/mdBook/format/mdbook.html#including-files

avivace:
@ISSOtm this should be documented in the wiki or at least somewhere to help contributors in building SVGs affect-able by the theme..

ISSOtm:
Sounds related to #277.

ISSOtm:
I wrote up something [on the wiki](https://github.com/gbdev/pandocs/wiki/SVG-guide); feel free to add any information you deem important or necessary, and ping me if I missed anything.

avivace:
> I wrote up something [on the wiki](https://github.com/gbdev/pandocs/wiki/SVG-guide); feel free to add any information you deem important or necessary, and ping me if I missed anything.

Thank you! Looks very nice

LIJI32:
While we're at it, [the use of ClearType here](https://github.com/gbdev/pandocs/blob/master/src/imgs/VGA_versus_CGB.png) is not good for my health.

Issue #449, opened 2022-11-08
Migrate from pygal 


avivace:
We need to migrate from pygal to another python plotting library. We currently use it to render two simple graphs from csv files: https://github.com/gbdev/pandocs/blob/c73a4421b0e7008ebc57bcda3e8754c53e7fdd03/src/imgs/src/graph_render.py

avivace:
This is not "mandatory" anymore, as we moved to pygal 3.0.0 which supports Python 3.10

PR #452, opened 2022-11-10
Plot generation: move to matplotlib, make output SVGs affected by CSS theme

avivace:
It may seem a bit convoluted, but it works:

https://user-images.githubusercontent.com/14352721/201120995-a3bf61d5-e9f6-472f-ba4a-b150bfb8c9dd.mp4

Some more stuff is needed to merge this but we should be already at a pretty good point:

- [x] move from pygal to matplotlib https://github.com/gbdev/pandocs/issues/449
- [x] make the output SVG use the CSS variables instead of hardcoded colros
- [ ] update requirements.txt
- [ ] move the call to the plot generation from the `renderer` to the `preproc` step of mdbook, otherwise the output SVG files cannot be inline injected
	- We should consider just calling a python script, instead of having the single calls for every plot like it is now [here](https://github.com/gbdev/pandocs/blob/master/book.toml#L16). @ISSOtm opinions?
- [ ] replace the two img tags with #import in the MBC5 article so the svg is actually injected in the HTML

Fixes https://github.com/gbdev/pandocs/issues/322

### DIFF
--- a/src/imgs/src/README.md
+++ b/src/imgs/src/README.md
@@ -1,0 +1,7 @@
+# imgs src
+
+This folder contains the source code and the datasets used to generate some plots.
+
+Those plots are included in Pan Docs as normal SVG figures, but their generation is part of pipeline.
+
+The scripts are traditional matplotlib executions but you may notice some additional manipulation on the resulting SVGs to set the lines and text to CSS variables instead of hardcoded colors.
--- a/src/imgs/src/graph_render.py
+++ b/src/imgs/src/graph_render.py
@@ -1,78 +1,97 @@
-import pygal
-from pygal.style import Style
-import math
 from sys import argv, stderr
 
-# ------------------------------------------------------------------------------------------------
-# Configuration Constants
-# ------------------------------------------------------------------------------------------------
+import lxml
+import matplotlib.pyplot as plt
+import pandas as pd
+from bs4 import BeautifulSoup
 
-# Rotation of X-Axis labels in degrees
-x_label_rotation = 0.01
-
-# Number of labels to be displayed on the X-Axis
-x_label_count = 10
-
-# ------------------------------------------------------------------------------------------------
-# The first line of the file must contain the X- and Y-Axis labels seperated by commas.
-# The following lines are expected to contain the graph data in a comma-separated format
-#  and in the same order as the Axis labels.
-# ------------------------------------------------------------------------------------------------
 
 def gen_graph(in_path, title):
-    custom_style = Style(
-        font_family="Inter",
-        label_font_size=12,
-        major_label_font_size=12,
-        title_font_size=16
+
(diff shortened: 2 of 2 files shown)


alloncm:
I might be missing something but why not use a Rust plotting library instead of calling a python script from the Rust renderer?
This way you can remove Python from the dependencies and maybe make the deployment easier.

The only reason I can think of is that Rust simply don't have a good plotting library with support for SVG's (but this looks like a good one - https://github.com/plotters-rs/plotters)

ISSOtm:
I swear I have seen a comment somewhere about using a Rust plotter instead to cut on the dependencies, suggesting https://github.com/plotters-rs/plotters, but I can't find it anymore.

ISSOtm:
The rationale for moving to a Rust library instead is to remove a contribution barrier, by using simpler setup instructions (no virtualenv) that work on all systems.

As for matplotlib itself, while you have a point that it is more popular, I can't say it's more standard—its documentation is very vague and confusing, and I have never seen two StackOverflow answers agreeing on a strategy to using it... :/

Besides, someone can always contribute a different setup later if they don't like the one we end up adopting, and it can use Python if they're willing to test their instructions. I don't think a move would be a blocker for that?

avivace:
> The rationale for moving to a Rust library instead is to remove a contribution barrier, by using simpler setup instructions (no virtualenv) that work on all systems.

Python is not mandatory to contribute, one can simply skip the plot generations. If someone desires can provide a plot generated with a rust program instead, too.

> As for matplotlib itself, while you have a point that it is more popular, I can't say it's more standard—its documentation is very vague and confusing, and I have never seen two StackOverflow answers agreeing on a strategy to using it... :/

I don't see how this is a negative points, it's okay if you have a lot of ways to do something.

> Besides, someone can always contribute a different setup later if they don't like the one we end up adopting, and it can use Python if they're willing to test their instructions. I don't think a move would be a blocker for that?

Did you check this PR? Imagine translating all the logic I added to parse SVGs to make them affected by CSS translated in Rust. I (personally) am not going to do that or ask anyone to go through that to contribute a plot. Python is super declarative and the matplotlib part is all self-declarative.

ISSOtm:
Spring has ended a few weeks ago, but pollen levels are still at a record high so it's still spring cleaning time if you ask me. Rebased (to check for any merge conflicts). Re-reading the arguments, sure, let's go with Python. We can consider moving to a Rust lib if at all later, that bikeshedding should not block this improvement.

We can't use `{{#include}}` tags for this without adding a new preproc step, because the `"links"` preproc is what processes these tags. Our custom preproc runs after it so that it can process transcluded fragments.

A possibility is making our custom preproc process e.g. `{{#plot}}` tags, which would invoke the Python script and replace the tag with the script's output. Then I think we can have the best of both worlds!

Issue #448, opened 2022-11-08
Unable to deploy with python version 3.10 and above


alloncm:
Hi, I was trying to to deploy the site locally on my Windows 10 machine.
After following the instructions in the DEPLOY.md file I was greeted with this error for `mdbook serve` command:
```sh
2022-11-08 21:57:18 [INFO] (mdbook::book): Book building has started
    Finished release [optimized] target(s) in 0.28s
     Running `target\release\pandocs-preproc.exe supports html`
    Finished release [optimized] target(s) in 0.26s
     Running `target\release\pandocs-preproc.exe`
warning: Referencing multiply-defined section "Palettes" (using chapter "Palettes")
warning: Referencing multiply-defined section "Palettes" (using chapter "Palettes")
warning: Referencing multiply-defined section "Window" (using chapter "Scrolling")
warning: Referencing multiply-defined section "Window" (using chapter "Scrolling")
warning: The Cartridge Header: Internal anchor link "#0144–0145 — New licensee code" not found, keeping as-is
warning: The Cartridge Header: Internal anchor link "#0144–0145 — New licensee code" not found, keeping as-is
warning: Referencing multiply-defined section "0000-1FFF - RAM Enable (Write Only)" (using chapter "Gameboy_Camera")
warning: Referencing multiply-defined section "0000-1FFF - RAM Enable (Write Only)" (using chapter "Gameboy_Camera")
2022-11-08 21:57:21 [INFO] (mdbook::book): Running the html backend
    Finished release [optimized] target(s) in 0.27s
     Running `target\release\pandocs-preproc.exe supports linkcheck`
    Finished release [optimized] target(s) in 0.29s
     Running `target\release\pandocs-preproc.exe`
warning: Referencing multiply-defined section "Palettes" (using chapter "Palettes")
warning: Referencing multiply-defined section "Palettes" (using chapter "Palettes")
warning: Referencing multiply-defined section "Window" (using chapter "Scrolling")
warning: Referencing multiply-defined section "Window" (using chapter "Scrolling")
warning: The Cartridge Header: Internal anchor link "#0144–0145 — New licensee code" not found, keeping as-is
warning: The Cartridge Header: Internal anchor link "#0144–0145 — New licensee code" not found, keeping as-is
warning: Referencing multiply-defined section "0000-1FFF - RAM Enable (Write Only)" (using chapter "Gameboy_Camera")
[...]

PR #453, opened 2022-11-11
Fix deployment on Windows

alloncm:
On windows the default python3 executable added to Path is not py3 and it depends on the installer.
Microsoft store adds both python and python3 and the official installer adds only python

for more details see #448

### DIFF
--- a/DEPLOY.md
+++ b/DEPLOY.md
@@ -2,7 +2,7 @@
 
 This document will explain you how to run a local copy of Pan Docs. 
 
-1. Install [Rust](https://www.rust-lang.org/tools/install), [mdBook](https://github.com/rust-lang/mdBook#readme), and [Python 3](https://www.python.org/downloads).
+1. Install [Rust](https://www.rust-lang.org/tools/install), [mdBook](https://github.com/rust-lang/mdBook#readme), and [Python 3](https://www.python.org/downloads) (3.9 or an earlier version).
   mdBook is the tool rendering the documentation, Rust is used for some custom plugins and Python scripts are used to render some images. E.g.:
   ```sh
   # Install Rust using rustup
@@ -10,7 +10,7 @@ This document will explain you how to run a local copy of Pan Docs.
   # Install mdbook using cargo
   cargo install mdbook
   # Remember to add cargo bin directory to your path
-  # Install Python 3.9 or earlier; for example, on APT-based systems (Ubuntu, Debian...):
+  # Install Python; e.g. on Debian-based systems:
   apt install python3
   ```
 2. [Install the Python prerequisites](https://packaging.python.org/tutorials/installing-packages/#requirements-files).
--- a/DEPLOY.md
+++ b/DEPLOY.md
@@ -10,7 +10,7 @@ This document will explain you how to run a local copy of Pan Docs.
   # Install mdbook using cargo
   cargo install mdbook
   # Remember to add cargo bin directory to your path
-  # Install python3 (on Debian systems)
+  # Install Python 3.9 or earlier; for example, on APT-based systems (Ubuntu, Debian...):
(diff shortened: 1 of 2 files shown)


DEPLOY.md
```diff
@@ -10,14 +10,19 @@ This document will explain you how to run a local copy of Pan Docs.
...
+  # On windows version 3.9 can be installed from the Microsoft Store
   ```
 2. [Install the Python prerequisites](https://packaging.python.org/tutorials/installing-packages/#requirements-files).
 ```sh
 # Create and activate a virtualenv
 python -m venv env
+# bash
 source env/bin/activate
+# Windows
+. .\env\Scripts\Activate.bat # On powershell repalce bat with ps1
```
avivace, line 25: I don't think there's any active maintainer on Windows, so I would strongly advice against adding windows-specific instructions nobody has the time nor the tools to keep maintained.

avivace (changes requested):
I don't agree with the changes in the documentation, but I do think we should check further on the executable we're pointing to

DEPLOY.md
alloncm: Fair point, the Rust code checking for Windows got me think it might be a wanted change.
Since Windows is not supported I think it might be good to add that in the DEPLOY.md and simply remove this if in the Rust code.

DEPLOY.md
```diff
@@ -2,6 +2,8 @@
 
 This document will explain you how to run a local copy of Pan Docs. 
 
+Be aware that Windows is not supported.
```
ISSOtm, line 5: Is it not? It's not tested, but I believe we do support it.

DEPLOY.md
```diff
@@ -10,7 +12,7 @@ This document will explain you how to run a local copy of Pan Docs.
   # Install mdbook using cargo
   cargo install mdbook
   # Remember to add cargo bin directory to your path
-  # Install python3 (on Debian systems)
+  # Install python3 (on Debian systems) make sure to install version 3.9
```
ISSOtm, line 15:
```suggestion
  # Install Python 3.9 or earlier; for example, on APT-based systems (Ubuntu, Debian...):
```

ISSOtm (changes requested):

DEPLOY.md
```diff
@@ -10,7 +10,7 @@ This document will explain you how to run a local copy of Pan Docs.
   # Install mdbook using cargo
   cargo install mdbook
   # Remember to add cargo bin directory to your path
-  # Install python3 (on Debian systems)
+  # Install Python 3.9 or earlier; for example, on APT-based systems (Ubuntu, Debian...):
```
avivace, line 13:
```suggestion
  # Install Python 3.9 or earlier; for example, on debian-based systems:
```

A system is not apt-based. A package manager can be apt-based.

alloncm: I accepted a [suggestion](https://github.com/gbdev/pandocs/pull/453#discussion_r1020931875) from @ISSOtm so I believe we need his comment here too

avivace:
@alloncm can you allow maintainers to edit your feature branch? it should be a flag somewhere (something like "write access to maintainers?")

DEPLOY.md
ISSOtm: I would say a system can be called "APT-based" if `apt` is configured for a `/usr` prefix instead of a `/usr/local` or other prefix. While it's theoretically possible to run multiple concurrent package managers, *usually* only one is the "system" package manager, and systems for which APT is that, I would call "APT-based".

And I believe that people running several concurrent package managers will be savvy enough to forgive the inaccuracy :stuck_out_tongue:

ISSOtm (approved):

avivace (approved):

PR #454, opened 2022-11-11
Document BG to OBJ priority in CGB mode

### LINKED ISSUE #443: Document the CGB OAM vs BG priority
Hi, 
when adding support for CGB to my GameBoy emulator, I noticed in the development of the new PPU that in the CGB mode the priority between the background layer and the sprites layer is not documented well and I couldn't figure out exactly the correct behavior and besides [cgb-acid2](https://github.com/mattcurrie/cgb-acid2) which really helped me there were no resources (at least that I found) so I decided to try and test it myself.

I wrote a [test rom](https://github.com/alloncm/MagenTests) for this and documented the expected behavior in the readme of the project but I thought maybe it will better to upstream this to the Pandocs as well.

Wanted to verify that this indeed is a requested feature before I actually add this and open a PR.

Thanks

alloncm:
Closes #443

### DIFF
--- a/src/LCDC.md
+++ b/src/LCDC.md
@@ -110,6 +110,8 @@ When Bit 0 is cleared, the background and window lose their priority -
 the sprites will be always displayed on top of background and window,
 independently of the priority flags in OAM and BG Map attributes.
 
+When Bit 0 is set [see background and window to obj priority](Tile_Maps.md#bg-to-obj-priority-in-cgb-mode).
+
 ## Using LCDC
 
 LCDC is a powerful tool: each bit controls a lot of behavior, and can be
--- a/src/LCDC.md
+++ b/src/LCDC.md
@@ -110,7 +110,7 @@ When Bit 0 is cleared, the background and window lose their priority -
 the sprites will be always displayed on top of background and window,
 independently of the priority flags in OAM and BG Map attributes.
 
-When Bit 0 is set, pixel priority is resolved [as described here](Tile_Maps.md#bg-to-obj-priority-in-cgb-mode).
+When Bit 0 is set, pixel priority is resolved [as described here](<#BG-to-OBJ Priority in CGB Mode>).
 
 ## Using LCDC
 
--- a/src/LCDC.md
+++ b/src/LCDC.md
@@ -110,7 +110,7 @@ When Bit 0 is cleared, the background and window lose their priority -
 the sprites will be always displayed on top of background and window,
 independently of the priority flags in OAM and BG Map attributes.
 
-When Bit 0 is set [see background and window to obj priority](Tile_Maps.md#bg-to-obj-priority-in-cgb-mode).
+When Bit 0 is set, pixel priority is resolved [as described here](Tile_Maps.md#bg-to-obj-priority-in-cgb-mode).
 
 ## Using LCDC
 
--- a/src/OAM.md
+++ b/src/OAM.md
@@ -122,4 +122,6 @@ This can be exploited to only hide parts of an object behind the background
 ([video demonstration](https://youtu.be/B8sJGgCVvnk)).
 A similar behaviour [can be seen on the NES](https://forums.nesdev.org/viewtopic.php?f=10&t=16861).
 
+**In CGB Mode**, BG vs. OBJ priority is declared in more than one register, [please see this page](Tile_Maps.md#bg-to-obj-priority-in-cgb-mode) for more details.
+
 :::
--- a/src/OAM.md
+++ b/src/OAM.md
@@ -122,6 +122,6 @@ This can be exploited to only hide parts of an object behind the background
 ([video demonstration](https://youtu.be/B8sJGgCVvnk)).
 A similar behaviour [can be seen on the NES](https://forums.nesdev.org/viewtopic.php?f=10&t=16861).
 
-**In CGB Mode**, BG vs. OBJ priority is declared in more than one register, [please see this page](Tile_Maps.md#bg-to-obj-priority-in-cgb-mode) for more details.
+**In CGB Mode**, BG vs. OBJ priority is declared in more than one register, [please see this page](<#BG-to-OBJ Priority in CGB Mode>) for more details.
 
 :::
--- a/src/Tile_Maps.md
+++ b/src/Tile_Maps.md
@@ -33,14 +33,42 @@ Bit 3    Tile VRAM Bank number      (0=Bank 0, 1=Bank 1)
 Bit 2-0  Background Palette number  (BGP0-7)
 ```
 
-When Bit 7 is set, the corresponding BG tile will have priority above
-all OBJs (regardless of the priority bits in OAM memory). There's also
-a Master Priority flag in LCDC register Bit 0 which overrides all other
-priority bits when cleared.
-
 Note that, if the map entry at `0:9800` is tile \$2A, the attribute at
 `1:9800` doesn't define properties for ALL tiles \$2A on-screen, but only
 the one at `0:9800`!
+
+### BG-to-OBJ Priority in CGB Mode
+
+In CGB Mode, the priority between the BG (and window) layer and the OBJ layer is declared in three different places:
+1. [BG Map Attribute bit 7](#bg-map-attributes-cgb-mode-only) - BG-to-OAM Priority (CGB Mode)
+2. [LCDC bit 0](LCDC.md#lcdc0--bg-and-window-enablepriority) - BG and Window Priority (CGB Mode)
+3. [OAM Attributes bit 7](OAM.md#byte-3--attributesflags) - BG and Window over OBJ
+
+We can infer the following rules from the table below:
+* When the BG color is 0 the OBJ will always have priority (ignoring the flags)
+* When LCDC bit 0 is clear OBJ will always have priority (ignoring the rest of the flags)
+* In order to grant the BG priority (color 1-3) LCDC bit 0 must be set and if OAM attributes bit 7 is clear BG Map attributes bit 7 must be set to override it
+
+::: tip NOTE
+
+OAM Attributes bit 7 will grant OBJ priority when clear and not when set
+
+:::
+
+The following table visualize the relations between the 3 flags
+
+LCDC bit 0 | OAM attr bit 7 | BG attr bit 7 | Priority
+:---------:|:--------------:|:-------------:|---------
+0          | 0              | 0             | OBJ
+0          | 0              | 1             | OBJ
+0          | 1              | 0             | OBJ
+0          | 1              | 1             | OBJ
+1          | 0              | 0             | OBJ
+1          | 0              | 1             | BG color 1–3, otherwise OBJ
+1          | 1              | 0             | BG color 1–3, otherwise OBJ
+1          | 1              | 1             | BG color 1–3, otherwise OBJ
+
+[This test ROM](https://github.com/alloncm/MagenTests) can be used to observe the above.
 
 ## Background (BG)
 
--- a/src/Tile_Maps.md
+++ b/src/Tile_Maps.md
@@ -45,9 +45,16 @@ In CGB Mode, the priority between the BG (and window) layer and the OBJ layer is
 - [OAM Attributes bit 7](<#[Byte 3 — Attributes/Flags](https://gbdev.io/pandocs/OAM.html#byte-3--attributesflags)>)
 
 We can infer the following rules from the table below:
-* When the BG color is 0 the OBJ will always have priority (ignoring the flags)
-* When LCDC bit 0 is clear OBJ will always have priority (ignoring the rest of the flags)
-* In order to grant the BG priority (color 1-3) LCDC bit 0 must be set and if OAM attributes bit 7 is clear BG Map attributes bit 7 must be set to override it
+* If the BG color index is 0, the OBJ will always have priority;
+* Otherwise, if LCDC bit 0 is clear, the OBJ will always have priority;
+* Otherwise, if both the BG Attributes and the OAM Attributes bit 7 is clear, the OBJ will have priority;
+* Otherwise, BG will have priority.
+
+::: tip
+
+BG Map attributes bit 7 is used to override OAM attributes bit 7
+
+:::
 
 The following table visualize the relations between the 3 flags
 
--- a/src/Tile_Maps.md
+++ b/src/Tile_Maps.md
@@ -49,12 +49,6 @@ We can infer the following rules from the table below:
 * When LCDC bit 0 is clear OBJ will always have priority (ignoring the rest of the flags)
 * In order to grant the BG priority (color 1-3) LCDC bit 0 must be set and if OAM attributes bit 7 is clear BG Map attributes bit 7 must be set to override it
 
-::: tip NOTE
-
-OAM Attributes bit 7 will grant OBJ priority when *clear*, not when *set*.
-
-:::
-
 The following table visualize the relations between the 3 flags
 
 LCDC bit 0 | OAM attr bit 7 | BG attr bit 7 | Priority
@@ -69,6 +63,15 @@ LCDC bit 0 | OAM attr bit 7 | BG attr bit 7 | Priority
 1          | 1              | 1             | BG color 1–3, otherwise OBJ
 
 [This test ROM](https://github.com/alloncm/MagenTests) can be used to observe the above.
+
+::: warning
+
+Keep in mind that:
+* OAM Attributes bit 7 will grant OBJ priority when **clear**, not when **set**.
+* priority between all OBJs is resolved **before** priority with the BG layer is considered.
+Please refer [to this page](<#Drawing priority>) for more details.
+
+:::
 
 ## Background (BG)
 
--- a/src/Tile_Maps.md
+++ b/src/Tile_Maps.md
@@ -50,12 +50,6 @@ We can infer the following rules from the table below:
 * Otherwise, if both the BG Attributes and the OAM Attributes have bit 7 clear, the OBJ will have priority;
 * Otherwise, BG will have priority.
 
-::: tip
-
-BG Map attributes bit 7 is used to override OAM attributes bit 7
-
-:::
-
 The following table shows the relations between the 3 flags:
 
 LCDC bit 0 | OAM attr bit 7 | BG attr bit 7 | Priority
--- a/src/Tile_Maps.md
+++ b/src/Tile_Maps.md
@@ -40,9 +40,9 @@ the one at `0:9800`!
 ### BG-to-OBJ Priority in CGB Mode
 
 In CGB Mode, the priority between the BG (and window) layer and the OBJ layer is declared in three different places:
-1. [BG Map Attribute bit 7](#bg-map-attributes-cgb-mode-only) - BG-to-OAM Priority (CGB Mode)
-2. [LCDC bit 0](LCDC.md#lcdc0--bg-and-window-enablepriority) - BG and Window Priority (CGB Mode)
-3. [OAM Attributes bit 7](OAM.md#byte-3--attributesflags) - BG and Window over OBJ
+- [BG Map Attribute bit 7](<#BG Map Attributes (CGB Mode only)>)
+- [LCDC bit 0](<#LCDC.0 — BG and Window enable/priority>)
+- [OAM Attributes bit 7](<#[Byte 3 — Attributes/Flags](https://gbdev.io/pandocs/OAM.html#byte-3--attributesflags)>)
 
 We can infer the following rules from the table below:
 * When the BG color is 0 the OBJ will always have priority (ignoring the flags)
@@ -51,7 +51,7 @@ We can infer the following rules from the table below:
 
 ::: tip NOTE
 
-OAM Attributes bit 7 will grant OBJ priority when clear and not when set
+OAM Attributes bit 7 will grant OBJ priority when *clear*, not when *set*.
 
 :::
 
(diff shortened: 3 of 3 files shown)


src/OAM.md
```diff
@@ -122,4 +122,6 @@ This can be exploited to only hide parts of an object behind the background
 ([video demonstration](https://youtu.be/B8sJGgCVvnk)).
 A similar behaviour [can be seen on the NES](https://forums.nesdev.org/viewtopic.php?f=10&t=16861).
 
+>**Notice:** in CGB mode interaction between BG and OBJ is declared in more than one register, [See background and window to obj priority](Tile_Maps.md#bg-to-obj-priority-in-cgb-mode)
```
avivace, line 125: We are not used (see the rest of the document) to add "Notices" titling them in bold and putting them in block quotes. This may be better going in a warning/note box?

src/Tile_Maps.md
```diff
@@ -33,15 +33,39 @@ Bit 3    Tile VRAM Bank number      (0=Bank 0, 1=Bank 1)
...
+### BG-to-OBJ Priority in CGB Mode
+
+In the CGB mode there are 3 different places where the priority between the BG (and window) layer and the OBJ layer is declared:
+1. [BG Map Attribute bit 7](#bg-map-attributes-cgb-mode-only) - BG-to-OAM Priority (CGB Mode)
+2. [LCDC bit 0](LCDC.md#lcdc0--bg-and-window-enablepriority) - BG and Window Priority (CGB Mode)
+3. [OAM Attributes bit 7](OAM.md#byte-3--attributesflags) - BG and Window over OBJ
+
+
+The following table visualize the relations between the 3 flags
+> **Notice:** OAM Attributes bit 7 will grant OBJ priority when clear and not when set
```
avivace, line 49: Same as my previous observation in this thread

src/Tile_Maps.md
```diff
@@ -33,15 +33,39 @@ Bit 3    Tile VRAM Bank number      (0=Bank 0, 1=Bank 1)
...
+| OAM Attributes bit 7 | BG Map Attribute bit 7 | LCDC bit 0 | Priority |
+| -------------------- | ---------------------- | ---------- | -------- |
+| [_] | [X] | [X] | BG color 0 - OBJ, otherwise BG |
+| [_] | [X] | [_] | OBJ |
+| [X] | [_] | [X] | BG color 0 - OBJ, otherwise BG |
+| [X] | [X] | [X] | BG color 0 - OBJ, otherwise BG |
+| [_] | [_] | [X] | OBJ |
+| [_] | [_] | [_] | OBJ |
+| [X] | [X] | [_] | OBJ |
+| [X] | [_] | [_] | OBJ |
```
avivace, line 60: Same here, I think this kind of "condition" table could be rendered in another way/more compliant to the rest of the document, especially because "[_]" and "[X]" are very unusual ways to render a Yes/No 1/0

src/LCDC.md
```diff
@@ -110,6 +110,8 @@ When Bit 0 is cleared, the background and window lose their priority -
 the sprites will be always displayed on top of background and window,
 independently of the priority flags in OAM and BG Map attributes.
 
+[See background and window to obj priority](Tile_Maps.md#bg-to-obj-priority-in-cgb-mode)
```
avivace, line 113: It's not very clear to me why the pointer is here. E.g.: how should the reader decide to click here? What kind of additional information do we need to convery that requires adding a pointer?
E.g. *the behaviour needs to be expanded because of an edge case that...*

avivace (changes requested):
Thanks @alloncm ! It looks good to me, but the modifications really need to be made more compliant to the style/formatting of the rest of the documents.

src/Tile_Maps.md
alloncm: Replaced with tip note

src/Tile_Maps.md
alloncm: Replaced with 1/0

src/LCDC.md
alloncm: Good point, I hope its clearer now

src/LCDC.md
```diff
@@ -110,6 +110,8 @@ When Bit 0 is cleared, the background and window lose their priority -
 the sprites will be always displayed on top of background and window,
 independently of the priority flags in OAM and BG Map attributes.
 
+When Bit 0 is set [see background and window to obj priority](Tile_Maps.md#bg-to-obj-priority-in-cgb-mode).
```
ISSOtm, line 113: Considering the behaviour is simply that OBJs always get priority (which is what's described just above), is it necessary to link to there?

src/OAM.md
```diff
@@ -122,4 +122,6 @@ This can be exploited to only hide parts of an object behind the background
 ([video demonstration](https://youtu.be/B8sJGgCVvnk)).
 A similar behaviour [can be seen on the NES](https://forums.nesdev.org/viewtopic.php?f=10&t=16861).
 
+In CGB mode interaction between BG and OBJ is declared in more than one register, [See background and window to obj priority](Tile_Maps.md#bg-to-obj-priority-in-cgb-mode)
```
ISSOtm, line 125:
```suggestion
**In CGB Mode**, BG vs. OBJ priority is declared in more than one register, [please see this page](Tile_Maps.md#bg-to-obj-priority-in-cgb-mode) for more details.
```

Maybe the tip box should be renamed "priority over BG", too?

src/Tile_Maps.md
```diff
@@ -33,15 +33,44 @@ Bit 3    Tile VRAM Bank number      (0=Bank 0, 1=Bank 1)
...
-a Master Priority flag in LCDC register Bit 0 which overrides all other
-priority bits when cleared.
-
 Note that, if the map entry at `0:9800` is tile \$2A, the attribute at
 `1:9800` doesn't define properties for ALL tiles \$2A on-screen, but only
 the one at `0:9800`!
 
+### BG-to-OBJ Priority in CGB Mode
+
+In the CGB mode there are 3 different places where the priority between the BG (and window) layer and the OBJ layer is declared:
```
ISSOtm, line 42:
```suggestion
In CGB Mode, the priority between the BG (and window) layer and the OBJ layer is declared in three different places:
```

src/Tile_Maps.md
```diff
@@ -33,15 +33,44 @@ Bit 3    Tile VRAM Bank number      (0=Bank 0, 1=Bank 1)
...
+1. [BG Map Attribute bit 7](#bg-map-attributes-cgb-mode-only) - BG-to-OAM Priority (CGB Mode)
+2. [LCDC bit 0](LCDC.md#lcdc0--bg-and-window-enablepriority) - BG and Window Priority (CGB Mode)
+3. [OAM Attributes bit 7](OAM.md#byte-3--attributesflags) - BG and Window over OBJ
```
ISSOtm, line 45:
```suggestion
- [BG Map Attribute bit 7](#bg-map-attributes-cgb-mode-only)
- [LCDC bit 0](LCDC.md#lcdc0--bg-and-window-enablepriority)
- [OAM Attributes bit 7](OAM.md#byte-3--attributesflags)
```

src/Tile_Maps.md
```diff
@@ -33,15 +33,44 @@ Bit 3    Tile VRAM Bank number      (0=Bank 0, 1=Bank 1)
...
+| 0 | 0 | 0 | OBJ |
+| 1 | 1 | 0 | OBJ |
+| 1 | 0 | 0 | OBJ |
+
+From the table we can infer the following:
+* When the BG color is 0 the OBJ will always have priority (ignoring the flags)
+* When LCDC bit 0 is clear OBJ will always have priority (ignoring the rest of the flags)
+* In order to grant the BG priority (color 1-3) LCDC bit 0 must be set and if OAM attributes bit 7 is clear BG Map attributes bit 7 must be set to override it
+
+A test ROM for this behavior can be found [here](https://github.com/alloncm/MagenTests)
```
ISSOtm, line 72:
```suggestion
[This test ROM](https://github.com/alloncm/MagenTests) can be used to observe the above.
```

src/Tile_Maps.md
```diff
@@ -33,15 +33,44 @@ Bit 3    Tile VRAM Bank number      (0=Bank 0, 1=Bank 1)
...
+| OAM Attributes bit 7 | BG Map Attribute bit 7 | LCDC bit 0 | Priority |
+| :--------------------: | :----------------------: | :----------: | -------- |
+| 0 | 1 | 1 | BG color 0 - OBJ, otherwise BG |
+| 0 | 1 | 0 | OBJ |
+| 1 | 0 | 1 | BG color 0 - OBJ, otherwise BG |
+| 1 | 1 | 1 | BG color 0 - OBJ, otherwise BG |
+| 0 | 0 | 1 | OBJ |
+| 0 | 0 | 0 | OBJ |
+| 1 | 1 | 0 | OBJ |
+| 1 | 0 | 0 | OBJ |
```
ISSOtm, line 65:
```suggestion
LCDC bit 0 | OAM attr bit 7 | BG attr bit 7 | Priority
:---------:|:--------------:|:-------------:|---------
0          | 0              | 0             | OBJ
0          | 0              | 1             | OBJ
0          | 1              | 0             | OBJ
0          | 1              | 1             | OBJ
1          | 0              | 0             | OBJ
1          | 0              | 1             | BG color 1–3, otherwise OBJ
1          | 1              | 0             | BG color 1–3, otherwise OBJ
1          | 1              | 1             | BG color 1–3, otherwise OBJ
```
I think organising the table like this makes the relationships clearer.

src/Tile_Maps.md
```diff
@@ -33,15 +33,44 @@ Bit 3    Tile VRAM Bank number      (0=Bank 0, 1=Bank 1)
...
+
+From the table we can infer the following:
+* When the BG color is 0 the OBJ will always have priority (ignoring the flags)
+* When LCDC bit 0 is clear OBJ will always have priority (ignoring the rest of the flags)
+* In order to grant the BG priority (color 1-3) LCDC bit 0 must be set and if OAM attributes bit 7 is clear BG Map attributes bit 7 must be set to override it
```
ISSOtm, line 70: I'd put this above the table (and rephrase accordingly), since the rules are really simple.

ISSOtm (changes requested):
Thanks!

src/LCDC.md
alloncm: When bit 0 is clear the OBJs always get priority, but when bit 0 is set the background and window don't always get priority and that's why I link there

src/Tile_Maps.md
alloncm: Why remove the names of the flags?

src/OAM.md
alloncm: I'm not really sure, most of this tip is about the internal priority between objects

avivace:
@alloncm please update the threads or close them when changing the related lines to allow us to better resume/restart our reviewing

avivace (approved):

src/LCDC.md
ISSOtm: Alright, that makes sense. How about this?

```suggestion
When Bit 0 is set, pixel priority is resolved [as described here](Tile_Maps.md#bg-to-obj-priority-in-cgb-mode).
```

src/Tile_Maps.md
```diff
@@ -33,15 +33,43 @@ Bit 3    Tile VRAM Bank number      (0=Bank 0, 1=Bank 1)
...
+
+In CGB Mode, the priority between the BG (and window) layer and the OBJ layer is declared in three different places:
+1. [BG Map Attribute bit 7](#bg-map-attributes-cgb-mode-only) - BG-to-OAM Priority (CGB Mode)
+2. [LCDC bit 0](LCDC.md#lcdc0--bg-and-window-enablepriority) - BG and Window Priority (CGB Mode)
+3. [OAM Attributes bit 7](OAM.md#byte-3--attributesflags) - BG and Window over OBJ
+
+We can infer the following rules from the table below:
+* When the BG color is 0 the OBJ will always have priority (ignoring the flags)
+* When LCDC bit 0 is clear OBJ will always have priority (ignoring the rest of the flags)
+* In order to grant the BG priority (color 1-3) LCDC bit 0 must be set and if OAM attributes bit 7 is clear BG Map attributes bit 7 must be set to override it
```
ISSOtm, line 50: I think this formulation is clearer:

```suggestion
* If the BG color index is 0, the OBJ will always have priority;
* Otherwise, if LCDC bit 0 is clear, the OBJ will always have priority;
* Otherwise, if neither the BG Attributes nor the OBJ Attributes have bit 7 set, the OBJ will have priority;
* Otherwise, BG will have priority.
```

ISSOtm (changes requested):
Good! I like this change, I only have some formatting/styling suggestions left.

src/Tile_Maps.md
```diff
@@ -33,15 +33,43 @@ Bit 3    Tile VRAM Bank number      (0=Bank 0, 1=Bank 1)
...
-all OBJs (regardless of the priority bits in OAM memory). There's also
-a Master Priority flag in LCDC register Bit 0 which overrides all other
-priority bits when cleared.
-
 Note that, if the map entry at `0:9800` is tile \$2A, the attribute at
 `1:9800` doesn't define properties for ALL tiles \$2A on-screen, but only
 the one at `0:9800`!
 
+### BG-to-OBJ Priority in CGB Mode
+
```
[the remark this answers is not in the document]
alloncm: Added this below the explanation and merged with the tip note about the oam bit 7

src/Tile_Maps.md
alloncm: I added this and reworded the third point a bit (as a non native English speaker it was a bit hard to understand at first) and added a tip about the purpose of the BG attr bit 7 (overriding the OAM attr bit 7)

avivace (approved):

src/LCDC.md
```diff
@@ -110,6 +110,8 @@ When Bit 0 is cleared, the background and window lose their priority -
 the sprites will be always displayed on top of background and window,
 independently of the priority flags in OAM and BG Map attributes.
 
+When Bit 0 is set, pixel priority is resolved [as described here](Tile_Maps.md#bg-to-obj-priority-in-cgb-mode).
```
ISSOtm, line 113:
```suggestion
When Bit 0 is set, pixel priority is resolved [as described here](<#BG-to-OBJ Priority in CGB Mode>).
```

src/OAM.md
```diff
@@ -122,4 +122,6 @@ This can be exploited to only hide parts of an object behind the background
 ([video demonstration](https://youtu.be/B8sJGgCVvnk)).
 A similar behaviour [can be seen on the NES](https://forums.nesdev.org/viewtopic.php?f=10&t=16861).
 
+**In CGB Mode**, BG vs. OBJ priority is declared in more than one register, [please see this page](Tile_Maps.md#bg-to-obj-priority-in-cgb-mode) for more details.
```
ISSOtm, line 125:
```suggestion
**In CGB Mode**, BG vs. OBJ priority is declared in more than one register, [please see this page](<#BG-to-OBJ Priority in CGB Mode>) for more details.
```

src/Tile_Maps.md
```diff
@@ -33,15 +33,53 @@ Bit 3    Tile VRAM Bank number      (0=Bank 0, 1=Bank 1)
...
+In CGB Mode, the priority between the BG (and window) layer and the OBJ layer is declared in three different places:
+- [BG Map Attribute bit 7](<#BG Map Attributes (CGB Mode only)>)
+- [LCDC bit 0](<#LCDC.0 — BG and Window enable/priority>)
+- [OAM Attributes bit 7](<#[Byte 3 — Attributes/Flags](https://gbdev.io/pandocs/OAM.html#byte-3--attributesflags)>)
```
ISSOtm, line 45:
```suggestion
- [OAM Attributes bit 7](<#Byte 3 — Attributes/Flags>)
```

src/Tile_Maps.md
```diff
@@ -33,15 +33,53 @@ Bit 3    Tile VRAM Bank number      (0=Bank 0, 1=Bank 1)
...
+
+We can infer the following rules from the table below:
+* If the BG color index is 0, the OBJ will always have priority;
+* Otherwise, if LCDC bit 0 is clear, the OBJ will always have priority;
+* Otherwise, if both the BG Attributes and the OAM Attributes bit 7 is clear, the OBJ will have priority;
```
ISSOtm, line 50:
```suggestion
* Otherwise, if both the BG Attributes and the OAM Attributes have bit 7 clear, the OBJ will have priority;
```

src/Tile_Maps.md
```diff
@@ -33,15 +33,53 @@ Bit 3    Tile VRAM Bank number      (0=Bank 0, 1=Bank 1)
...
+* Otherwise, BG will have priority.
+
+::: tip
+
+BG Map attributes bit 7 is used to override OAM attributes bit 7
```
ISSOtm, line 55: What is this saying?

src/Tile_Maps.md
```diff
@@ -33,15 +33,53 @@ Bit 3    Tile VRAM Bank number      (0=Bank 0, 1=Bank 1)
...
+
+:::
+
+The following table visualize the relations between the 3 flags
```
ISSOtm, line 59:
```suggestion
The following table shows the relations between the 3 flags:
```

src/Tile_Maps.md
```diff
@@ -33,15 +33,53 @@ Bit 3    Tile VRAM Bank number      (0=Bank 0, 1=Bank 1)
...
+1          | 1              | 1             | BG color 1–3, otherwise OBJ
+
+[This test ROM](https://github.com/alloncm/MagenTests) can be used to observe the above.
+
+::: warning
+
+Keep in mind that:
+* OAM Attributes bit 7 will grant OBJ priority when **clear**, not when **set**.
+* priority between all OBJs is resolved **before** priority with the BG layer is considered.
+Please refer [to this page](<#Drawing priority>) for more details.
```
ISSOtm, line 79:
```suggestion
* Priority between all OBJs is resolved **before** priority with the BG layer is considered.
  Please refer [to this page](<#Drawing priority>) for more details.
```

ISSOtm (changes requested):
Alright! Almost done.

src/Tile_Maps.md
alloncm: I'm not sure this is the best way to say it but I'm trying to explain the logical purpose of the BG Map attributes bit 7.
This is not a crucial detail but rather an interesting observation about the relation between those flags.

ISSOtm: I don't understand how it "overrides" the other bit, since their roles are symmetrical.

alloncm: I don't think their roles are symmetrical, in order to grant the BG layer priority over the OBJ layer you first need LCDC bit 0 to be set.
Only if the OBJ attributes bit 7 is clear you also need to set BG attributes bit 7 but if OBJ attributes bit 7 is set BG attributes bit 7 is not relevant and can be ignored (it's state doesn't matter and the BG will have priority).

This was rather not really intuitive to me at first so I thought mentioning this in a tip box could help the reader.

ISSOtm: I agree that LCDC bit 0 has a different role, but the role of BG and OBJ attributes' bit 7 is identical (besides the latter's inverted meaning).

src/Tile_Maps.md
```diff
@@ -33,15 +33,53 @@ Bit 3    Tile VRAM Bank number      (0=Bank 0, 1=Bank 1)
...
+LCDC bit 0 | OAM attr bit 7 | BG attr bit 7 | Priority
+:---------:|:--------------:|:-------------:|---------
+0          | 0              | 0             | OBJ
+0          | 0              | 1             | OBJ
+0          | 1              | 0             | OBJ
+0          | 1              | 1             | OBJ
+1          | 0              | 0             | OBJ
+1          | 0              | 1             | BG color 1–3, otherwise OBJ
+1          | 1              | 0             | BG color 1–3, otherwise OBJ
+1          | 1              | 1             | BG color 1–3, otherwise OBJ
```
ISSOtm, line 70:
```suggestion
0          | 1              | 0             | OBJ
0          | 1              | 1             | OBJ
0          | 0              | 0             | OBJ
0          | 0              | 1             | OBJ
1          | 1              | 0             | OBJ
1          | 1              | 1             | BG color 1–3, otherwise OBJ
1          | 0              | 0             | BG color 1–3, otherwise OBJ
1          | 0              | 1             | BG color 1–3, otherwise OBJ
```
Looks like we forgot that OBJ attributes bit 7 is inverted... right?

ISSOtm (changes requested):

src/Tile_Maps.md
alloncm: The original is the actual value of the bit, not sure if you want to invert it but I think it will more confusing seeing in this table inverted values for this bit (there is also a warning below stating that OBJ get priority upon clear and not set).

ISSOtm: When the bit is 0, the OBJ loses priority, so  `1 0 0` shows BG, not the OBJ. The original appears not to be `[$FExx] >> 7`.

ISSOtm: No, it is correct, I'm just being confused.

src/Tile_Maps.md
alloncm: After more thinking I realized you are correct

ISSOtm (approved):

avivace (approved):

PR #457, opened 2022-11-22
Add address of NR2x registers

### LINKED ISSUE #455: Addresses for NR2x registers not mentioned in "Audio Registers"
Every other audio/hardware register has its corresponding address listed, and I'm surprised that the addresses for the NR2x registers were left out with the recent-ish overhaul of the audio section. I understand their function is left out due to it being identical to the NR1x registers, but their addresses in relation to the register names do not have the same scheme as the NR1x registers (NR10-NR14 = FF10-FF14, but NR21-NR24 = FF16-FF19, not FF21-FF24) and may be confusing to those new to the Gameboy's audio hardware. My memory isn't the best, so I frequently refer to the Pandocs when I want to use the hex address values instead of the name of the register. IMO the omission should be dealt with sooner rather than later.

ISSOtm:
Fixes #455


ISSOtm:
cc @jbshelton

avivace (approved):

PR #458, opened 2022-11-27
Loop 41 times in the compact HDMA procedure

tertu-m:
If it loops 40 times, this procedure will begin to return before the OAM DMA transfer is complete, corrupting the return address (in SameBoy at least). I have not tested this behavior on hardware, but could do so.


pinobatch:
Can it instead loop 20 times and return conditionally?

ISSOtm:
It should be able to, yes.

ISSOtm (approved):
Rebased, and switched to the hardware-verified `ret z` approach.

avivace (approved):

PR #459, opened 2022-12-25
src/HuC3.md: Fixed a bit position.

cuavas:
I was really sleep-deprived when I originally wrote this – the 4-bit field is bits 3-0, obviously.


ISSOtm:
Today is Christmas, and thus you are forgiven. Thank you! :)

Issue #437, opened 2022-09-18
Provide a docker image for local development


avivace:
To make (first-time?) contributors' life easier, it may be nice to provide a docker image that sets up the development environment and directly serves rendered pandocs.

PR #462, opened 2023-01-15
Add Docker image for local development

leinacc:
draft because:
1. ~~review on if mount should include more directories~~ Added the 4 directories used by mdbook's preprocessors
2. review on whether an ubuntu container with 3.10 already installed is a good base, or if we want an alternative OS, with a different python install. Due to complexity of python/pip/venv-related installs, im leaning towards a python-based image.
3. ~~`build-essential` could be overkill. Need to check the actual minimum required for `cargo`~~ Replaced with just `gcc`
4. ~~`mdbook watch` won't let you ctrl-c out of `docker run`, so may need an alternative~~ `-it` flags added to `docker run` to allow this

Fixes https://github.com/gbdev/pandocs/issues/437


avivace:
Wow, this is awesome! I've been meaning to do this since a while. Thanks @leinacc !

Dockerfile
```diff
@@ -0,0 +1,23 @@
+# Get a python 3.10 image with curl
+FROM revolutionsystems/python:3.10.0-wee-optimized-lto
```
avivace, line 2: What about basing it on the official Rust image? https://hub.docker.com/_/rust

leinacc: I'll give it an experiment, might make the build process faster/lightweight

leinacc: I'm getting a number of issues using the rust image as a base. So far, before `COPY`, it's this:
```dockerfile
# Get a rust image with python 3.9
FROM rust:latest
SHELL ["bash", "-lc"]
# have to reference cargo like this
RUN /usr/local/cargo/bin/cargo install mdbook

# Upgrade to python 3.10
RUN echo "deb http://http.us.debian.org/debian/ testing non-free contrib main" >> /etc/apt/sources.list
RUN apt update
RUN apt upgrade python3 -y; exit 0
RUN apt install python3-pip -y; exit 0
```

Those `exit 0`s are due to apt throwing the following, causing `apt` commands to fail finishing
```bash
Setting up usrmerge (35) ...
Warning: overlayfs detected, /usr/lib/usrmerge/convert-usrmerge will not
be run automatically. See #1008202 for details.

If this is a container then it can be converted by unpacking the image,
entering it with chroot(8), installing usrmerge and then repacking the
image again. at /usr/lib/usrmerge/convert-usrmerge line 402.
E: usrmerge failed.
dpkg: error processing package usrmerge (--configure):
 installed usrmerge package post-installation script subprocess returned error exit status 1
```

`python3 -m venv env` also fails with a mysterious error `Error: name 'cmd' is not defined` I can't find solutions for

The above, coupled with python:3.10.0-wee-optimized-lto being a lightweight container makes me lean towards keeping it

leinacc:
Applied all my draft message changes now. Given how much easier it is to install rust over python3/pip3/venv (plus OS dependencies and fixes), I think we use the python3.10 image as a base

avivace (commented):
> Applied all my draft message changes now. Given how much easier it is to install rust over python3/pip3/venv (plus OS dependencies and fixes), I think we should use the python3.10 image as a base

Okay, I tried it and it seems pretty nice! Really eases the development setup phase.

Two things:
- The shell is unresponsive to keyboard interrupts. Sending CTRL+C or CTRL+Z doesn't seem to have any effect.. I don't really like having to send a docker kill command to stop it
- I'm still super fond of using this base image. In the end we really just need Python3.10 + curl, I would prefer to use something more official (e.g. debian or alpine base ones?) so they provide a more solid/maintained base 


In the meantime I'm preparing another PR to add a step to the CI/CD to automatically build and push the docker image to the container registry of this repository, so users could directly pull the image and not have to build it (https://github.com/gbdev/pandocs/issues/463)

leinacc:
Agreed with the above 2. They're fixed now. The new docker run flags should let you stop the container with ctrl-c

PR #464, opened 2023-01-16
fix typo in LCD mode timing description

iximeow:
the microsecond estimates for mode timings didn't line up. presumably HBlank is intended to pad times for each line at a consistent 376 dots, and adding minimum mode 3 + maximum mode 0 equals maximum mode 3 + minimum mode 0 - 376  dots either way. but the microsecond estimates didn't match: 60 + 20 = 80µs vs 40 + 49 = 89µs.

presumably the correct µs estimates for mode 3 are "(40 to 69µs)":
```
>> 291 / (2 ** 22) * 1000 * 1000
> 69.38 (microseconds)
```

correcting this value makes the min/max sums of mode 3/mode 0 times consistently 89µs, which is almost the same as the time for 376 dots:
```
>> 376 / (2 ** 22) * 1000 * 1000
> 89.65
```

(i also spot-checked the other numbers in this table and everything else lines up)

((also thank you all very much for these documents!!))

### DIFF
--- a/src/STAT.md
+++ b/src/STAT.md
@@ -51,10 +51,10 @@ to the CPU: writes are ignored, and reads return garbage values (usually $FF).
 
 Mode | Action                                                      | Duration                                                           | Accessible video memory
 -----|------------------------------------------------------------------|--------------------------------------------------------------------|-------------------------
-  2  | Searching OAM for OBJs whose Y coordinate overlap this line | 80 dots                                                    | VRAM, CGB palettes
+  2  | Searching OAM for OBJs whose Y coordinate overlap this line | 80 dots                                               | VRAM, CGB palettes
   3  | Reading OAM and VRAM to generate the picture                | 168 to 291 dots, depending on sprite count            | None
-  0  | Nothing (HBlank)                               | 85 to 208 dots, depending on previous mode 3 duration | VRAM, OAM, CGB palettes
-  1  | Nothing (VBlank)                                 | 4560 dots (10 scanlines)                                  | VRAM, OAM, CGB palettes
+  0  | Nothing (HBlank)                                            | 85 to 208 dots, depending on previous mode 3 duration | VRAM, OAM, CGB palettes
+  1  | Nothing (VBlank)                                            | 4560 dots (10 scanlines)                              | VRAM, OAM, CGB palettes
 
 ## Properties of STAT modes
 
--- a/src/STAT.md
+++ b/src/STAT.md
@@ -52,7 +52,7 @@ to the CPU: writes are ignored, and reads return garbage values (usually $FF).
 Mode | Action                                                      | Duration                                                           | Accessible video memory
 -----|------------------------------------------------------------------|--------------------------------------------------------------------|-------------------------
   2  | Searching OAM for OBJs whose Y coordinate overlap this line | 80 dots (19 µs)                                                    | VRAM, CGB palettes
-  3  | Reading OAM and VRAM to generate the picture                | 168 to 291 dots (40 to 60 µs) depending on sprite count            | None
+  3  | Reading OAM and VRAM to generate the picture                | 168 to 291 dots (40 to 69 µs) depending on sprite count            | None
   0  | Nothing (HBlank)                               | 85 to 208 dots (20 to 49 µs) depending on previous mode 3 duration | VRAM, OAM, CGB palettes
   1  | Nothing (VBlank)                                 | 4560 dots (1087 µs, 10 scanlines)                                  | VRAM, OAM, CGB palettes
 
--- a/src/STAT.md
+++ b/src/STAT.md
@@ -51,10 +51,10 @@ to the CPU: writes are ignored, and reads return garbage values (usually $FF).
 
 Mode | Action                                                      | Duration                                                           | Accessible video memory
 -----|------------------------------------------------------------------|--------------------------------------------------------------------|-------------------------
-  2  | Searching OAM for OBJs whose Y coordinate overlap this line | 80 dots (19 µs)                                                    | VRAM, CGB palettes
-  3  | Reading OAM and VRAM to generate the picture                | 168 to 291 dots (40 to 69 µs) depending on sprite count            | None
-  0  | Nothing (HBlank)                               | 85 to 208 dots (20 to 49 µs) depending on previous mode 3 duration | VRAM, OAM, CGB palettes
-  1  | Nothing (VBlank)                                 | 4560 dots (1087 µs, 10 scanlines)                                  | VRAM, OAM, CGB palettes
+  2  | Searching OAM for OBJs whose Y coordinate overlap this line | 80 dots                                                    | VRAM, CGB palettes
+  3  | Reading OAM and VRAM to generate the picture                | 168 to 291 dots, depending on sprite count            | None
+  0  | Nothing (HBlank)                               | 85 to 208 dots, depending on previous mode 3 duration | VRAM, OAM, CGB palettes
+  1  | Nothing (VBlank)                                 | 4560 dots (10 scanlines)                                  | VRAM, OAM, CGB palettes
 
 ## Properties of STAT modes
 


src/STAT.md
```diff
@@ -52,7 +52,7 @@ to the CPU: writes are ignored, and reads return garbage values (usually $FF).
 Mode | Action                                                      | Duration                                                           | Accessible video memory
 -----|------------------------------------------------------------------|--------------------------------------------------------------------|-------------------------
   2  | Searching OAM for OBJs whose Y coordinate overlap this line | 80 dots (19 µs)                                                    | VRAM, CGB palettes
```
ISSOtm, line 54:
```suggestion
  2  | Searching OAM for OBJs whose Y coordinate overlap this line | 80 dots (20 µs)                                                    | VRAM, CGB palettes
```

src/STAT.md
```diff
@@ -52,7 +52,7 @@ to the CPU: writes are ignored, and reads return garbage values (usually $FF).
...
-  3  | Reading OAM and VRAM to generate the picture                | 168 to 291 dots (40 to 60 µs) depending on sprite count            | None
+  3  | Reading OAM and VRAM to generate the picture                | 168 to 291 dots (40 to 69 µs) depending on sprite count            | None
   0  | Nothing (HBlank)                               | 85 to 208 dots (20 to 49 µs) depending on previous mode 3 duration | VRAM, OAM, CGB palettes
```
ISSOtm, line 56:
```suggestion
  3  | Reading OAM and VRAM to generate the picture                | 168 to 291 dots (42 to 73 µs) depending on sprite count            | None
  0  | Nothing (HBlank)                               | 85 to 208 dots (21 to 52 µs) depending on previous mode 3 duration | VRAM, OAM, CGB palettes
```

ISSOtm (changes requested):
Going by the definition at the top of [this page](https://gbdev.io/pandocs/pixel_fifo.html), one "dot" is 1/4 µs. So we should compute the estimate as `291 / 4` = 72.75, which I'd round up to 73 because it's an upper bound.

But your method is also correct! It's just that all other µs numbers are wrong :D

((Thank you for your contribution!))

avivace (approved):

iximeow:
> Going by the definition at the top of [this page](https://gbdev.io/pandocs/pixel_fifo.html), one "dot" is 1/4 µs.

@ISSOtm i'm not sure where to see that? i do see

> All references to a dot are meant as dots (4.19 MHz).

which i think itself is an estimate (the clock _is_ `2^22hz` in normal mode right?). so rather than calculating from the `1/4µs` estimate i think it makes more sense to whatever is stating a dot is a firm 1/4µs! going by the 4.19MHz clock a dot would _actually_ be 0.2384µs..

it's aesthetically displeasing that the 376 dots to a scan line comes out to 89.645 microseconds, since that rounds (by quite a bit, in context) up to 90.  `208 dots` happens to round to 50 microseconds and "works out" (rounded time sums to 90µs when mode 3 is `168 dots -> 40µs`) but sweating the off-by-one-ness of microseconds doesn't seem like a big deal here :D

ISSOtm:
Ok, yes, I got owned by SI versus binary units, a dot is not 1/4 µs, it's 1/4... [whatever the binary equivalent of micro is]second. Re-calculating appropriately, I match your correction, and that 208 cycles (49.591... µs) rounds to 50 µs. We would want to round up for the range's upper bound anyway.

> the clock is 2^22hz in normal mode right?

The dot clock is *always* 4 MiHz, even in double-speed mode; this is why it's introduced as a unit separate from cycles, which are either at 4 MiHz or 8 MiHz.

> it's aesthetically displeasing that the 376 dots to a scan line comes out to 89.645 microseconds, since that rounds (by quite a bit, in context) up to 90.

If that helps you feel better, a scanline actually includes Mode 2's 80 dots for a total of 456 dots/scanline (which is a tenth of VBlank's length); that amounts to `1000000/4194304* 456` = 108.7189 µs, which is rounds slightly closer 😛.

Unfortunately, the conversion ratio means that nothing will round nicely, unless we worked with numbers close to 2^16... so either we live with that, or we remove the estimates outright.

---

**Removing the µs numbers** actually seems to me more and more like the better solution, since they seem much less useful info than cycle counts, considering we're primarily working from the CPU's PoV. It's fairly easy to do the conversion, too, if you end up needing it.

aaaaaa123456789:
I'd be happy removing the wall time measurements. The dot clock has a well-known rate, so anyone who is in the very unlikely scenario of needing those values can trivially calculate them, and they're just confusing otherwise, because most people mentally assume that 1us = one single-speed CPU cycle.

iximeow:
> since they seem much less useful info than cycle counts, considering we're primarily working from the CPU's PoV. It's fairly easy to do the conversion, too, if you end up needing it.

i'm (easily) sold! i've pushed another commit that removes the microsecond timings from this table. i left the `16.74 ms` figure above that describes the entire frame because it seems like a good way to introduce that these timings really are _an entire frame_ - it's only off by 0.016%, too.

aaaaaa123456789:
> i left the `16.74 ms` figure above that describes the entire frame because it seems like a good way to introduce that these timings really are _an entire frame_ - it's only off by 0.016%, too.

Doesn't seem to be off at all to me:

image

src/STAT.md
```diff
@@ -51,10 +51,10 @@ to the CPU: writes are ignored, and reads return garbage values (usually $FF).
...
-  2  | Searching OAM for OBJs whose Y coordinate overlap this line | 80 dots (19 µs)                                                    | VRAM, CGB palettes
-  3  | Reading OAM and VRAM to generate the picture                | 168 to 291 dots (40 to 69 µs) depending on sprite count            | None
-  0  | Nothing (HBlank)                               | 85 to 208 dots (20 to 49 µs) depending on previous mode 3 duration | VRAM, OAM, CGB palettes
-  1  | Nothing (VBlank)                                 | 4560 dots (1087 µs, 10 scanlines)                                  | VRAM, OAM, CGB palettes
+  2  | Searching OAM for OBJs whose Y coordinate overlap this line | 80 dots                                                    | VRAM, CGB palettes
+  3  | Reading OAM and VRAM to generate the picture                | 168 to 291 dots, depending on sprite count            | None
+  0  | Nothing (HBlank)                               | 85 to 208 dots, depending on previous mode 3 duration | VRAM, OAM, CGB palettes
+  1  | Nothing (VBlank)                                 | 4560 dots (10 scanlines)                                  | VRAM, OAM, CGB palettes
```
ISSOtm, line 57:
```suggestion
  2  | Searching OAM for OBJs whose Y coordinate overlap this line | 80 dots                                               | VRAM, CGB palettes
  3  | Reading OAM and VRAM to generate the picture                | 168 to 291 dots, depending on sprite count            | None
  0  | Nothing (HBlank)                                            | 85 to 208 dots, depending on previous mode 3 duration | VRAM, OAM, CGB palettes
  1  | Nothing (VBlank)                                            | 4560 dots (10 scanlines)                              | VRAM, OAM, CGB palettes
```

ISSOtm (changes requested):
Yep, looks good to me then, just with neater table alignment.

iximeow:
it's off by 0.002706298828125 milliseconds :sunglasses:
```
; 100 - (16.74 / (70224 / (2 ** 22) * 1000)) * 100
	~0.01616
```
is how i computed the 0.016% error. close enough to hopefully not matter.

aaaaaa123456789:
> it's off by 0.002706298828125 milliseconds sunglasses
> 
> ```
> ; 100 - (16.74 / (70224 / (2 ** 22) * 1000)) * 100
> 	~0.01616
> ```
> 
> is how i computed the 0.016% error. close enough to hopefully not matter.

Error below half a unit in the last place is just a matter of significant digits, not a true error.

ISSOtm:
Yes, though it should still be pointed out that rounding did occur.

ISSOtm (approved):

avivace (approved):

PR #465, opened 2023-01-17
Document SGB command $19 - PAL_PRI

leinacc:
SGB command $19 points to the following (ROM $00c9e7 in SGB2):

```
CMD_PAL_PRI:
	lda wSGBPacketsData+1
	and #$01
	sta $02f4
	rts 
```

Example test:
Repeat the steps below with $0002f4 set to 0, then 1
* Find a game that sends a non-PAL_TRN palette command (for example: https://hh.gbdev.io/game/snek-gbc)
* Change the user's palette in the SGB menus (either 1st menu - predefined choices, or 4th menu - custom colors)
* Have the PAL command sent (so for Snek, that's transition to and from the title screen)
* If $0002f4 is 0, the user's palette choice should still affect the game, otherwise the game's palette choice will

Hardware test:
Use the steps above, running Space Invaders SGB to test $0002f4=0, and Donkey Kong 1994 for =1


src/SGB_Command_Undocumented.md
```diff
@@ -4,16 +4,10 @@ The following information has been extracted from disassembling a SGB1v2
...
 
 ## Stubbed commands
 
 Commands $1A to $1F (inclusive)'s handlers are stubs (only contain an
 `RTS`). This is interesting, since the command-processing function
 explicitly ignores commands $1E and $1F.
-
-## SGB command $19
-
-The game _Donkey Kong_ (1994) appears to send this command, and it appears
```
avivace, line 17: Looks good to me! The only thing I would do is to keep a reference to Donkey Kong as an example of game that does it

avivace (changes requested):

src/SGB_Command_Palettes.md
```diff
@@ -89,6 +89,8 @@ This command can set a flag that will let the game clear and prevent use of
 the user's SGB palette choice, when it sends 1 of the above commands, except 
 PAL_TRN.
 
+_Donkey Kong_ (1994) is 1 known game that appears to use this.
```
avivace, line 92:
```suggestion
_Donkey Kong_ (1994) is one known game that appears to use this.
```

avivace (changes requested):

ISSOtm:
I have hardware to test this, but I'm not sure which BIOS revision it is running, nor exactly what the test steps are.

leinacc:
> I have hardware to test this, but I'm not sure which BIOS revision it is running, nor exactly what the test steps are.

I've got the following revisions that are all affected:
* version 0: md5: 6a702a9c6c110863ace946321b871632
* version 1: md5: a4d4d35c827aedbda3eece1af0b57902
* version 2: md5: b15ddb15721c657d82c5bab6db982ee9

Best way to show the effect is by playing Donkey Kong 94, and any other SGB game and for each:
* Set user's palettes somewhere in the 1st or 4th sgb menu option
* Transition to a state in which the other game would have sent another PAL command (pressing Start on the title screens of DK and Space Invaders)
* For Donkey Kong, the palettes would revert to the game's choice, and the user's palette option would have reset if you go back to the menus
* For the other game, you will still keep the user's palettes

Alternatively, I can try getting a test rom up where you can toggle the relevant SNES ram var

ISSOtm:
Yep, that matches what I got.

ISSOtm:
(Unrelatedly, it would be nice to have a `which.sgb` that tells you which SGB BIOS revision you are running.)

avivace (approved):

src/SGB_Command_Palettes.md
```diff
@@ -82,3 +82,20 @@ The palette data is sent by VRAM-Transfer (4 KBytes).
 
 Each Palette consists of four 16-bit color definitions (8 bytes). Note:
 The data is stored at 3000h-3FFFh in SNES memory.
+
+## SGB Command $19 — PAL_PRI
+
+This command can set a flag that will let the game clear and prevent use of 
+the user's SGB palette choice, when it sends 1 of the above commands, except 
+PAL_TRN.
```
ISSOtm, line 90:
```suggestion
If the player overrides the active palette set (a pre-defined or the custom one), it stays in effect until the smiley face is selected again, or the player presses the X button on their SNES controller.

However, if `PAL_PRI` is enabled, then changing the palette set (via any of the above commands except `PAL_TRN`) will switch back to the game's newly-modified palette set, if it wasn't already active.
```
The check if for the flag to be zero or not, right?

Also, can you reach a hybrid state e.g. by using all four palettes, and then only sending e.g. a `PAL01` packet? Or would palettes 2 and 3 also revert to the game-defined ones?

ISSOtm (changes requested):
I didn't find the behaviour's description very clear, so here's my proposal.

src/SGB_Command_Palettes.md
leinacc: I like it. I will check the hybrid state later

leinacc: The check is for 0 yes, but due to the `and #$01` mask, it's really `bit 0 set: system pals priority`. I'll change that

The behavior observed (in emulator and bios code, I'll need to write a test rom for hardware) is that there is a separate palette buffer area for system-defined palettes and user-defined palettes:
```
System: $000400-$0005ff (shadow CGRAM)
User: $7ecc30-$7ecc37 (4 colors copied to CGDATA for all of BG0-BG3 palettes)
```
If PAL_PRI is in effect, PAL commands will trigger an update from the system-defined palettes, so palettes 2 and 3 will revert to last palettes set for them (or SGB's default BG palettes). So no hybrid state.

If PAL_PRI is not in effect, they will still be saved, but will only reveal themselves if, for example, the user clicks the 'face' thing that clears the user's palette choice, or if they send PAL_PRI with system priority, then a PAL command after.

leinacc: Definitely, just keeping it thorough :) Yes, the other bits are completely cleared before saving to wram

ISSOtm (approved):

src/SGB_Command_Palettes.md
```diff
@@ -82,3 +82,19 @@ The palette data is sent by VRAM-Transfer (4 KBytes).
 
 Each Palette consists of four 16-bit color definitions (8 bytes). Note:
 The data is stored at 3000h-3FFFh in SNES memory.
+
+## SGB Command $19 — PAL_PRI
+
+If the player overrides the active palette set (a pre-defined or the custom one), it stays in effect until the smiley face is selected again, or the player presses the X button on their SNES controller.
```
avivace, line 88: _A pre-defined or THE custom one_ sounds a bit unclear here... Maybe "the pre-defined or the the custom one"? or "a pre-defined or a custom one"?

src/SGB_Command_Palettes.md
```diff
@@ -82,3 +82,19 @@ The palette data is sent by VRAM-Transfer (4 KBytes).
...
+
+However, if `PAL_PRI` is enabled, then changing the palette set (via any of the above commands except `PAL_TRN`) will switch back to the game's newly-modified palette set, if it wasn't already active.
```
avivace, line 90:
```suggestion
However, if `PAL_PRI` is enabled then changing the palette set (via any of the above commands except `PAL_TRN`) will switch back to the game's newly-modified palette set, if it wasn't already active.
```

avivace (changes requested):
LGTM, just left a couple of minor comments

src/SGB_Command_Palettes.md
ISSOtm: There are several pre-defined palettes, but only a single custom one.

avivace: And those are the ones getting "overridden" by the player?

leinacc: This is the single custom one:
Screenshot_20230118_211531

These are the predefined ones (4 pages of 8):
Screenshot_20230118_211604

These are user-selectable palettes, which can be overridden when PAL_PRI's flag is set

avivace: Okay, I was just trying to understand what information we need to convey by specifying that the active one can be either the custom or one of the default ones, which I guess is just the standard behaviour, hence it doesn't add any information to the "active palette set" so it could be removed?

avivace, line 88:
```suggestion
If the player overrides the active palette set, it stays in effect until the smiley face is selected again, or the player presses the X button on their SNES controller.
```

avivace (changes requested):

src/SGB_Command_Palettes.md
leinacc: Would it be worth describing system palettes and user palettes above this? Then the initial part of the sentence would be

If the player overrides the system palettes (with a pre-defined or the custom user palette),

avivace: Yes, but I don't want to hold PR for this, so I'll merge this in the the current status and we can take care of improving this later.

avivace (approved):

Issue #439, opened 2022-09-19
Move LY and LYC elsewhere


ISSOtm:
They don't really pertain to scrolling (where they currently are, mostly because that's where they have always been), but where else should they go?

pinobatch:
If anywhere, `LY` and `LYC` belong near `STAT`, as they reflect the status of the PPU and control one of the STAT IRQ sources.

PR #469, opened 2023-01-24
Move ly lyc to stat page

leinacc:
Closes https://github.com/gbdev/pandocs/issues/439

I decided on putting these on top of STAT as they're mentioned in sentence 2 of STAT's description, so it's fresh info

Maybe title needs changing? Though LCD Status Registers feels wrong

### DIFF
--- a/src/STAT.md
+++ b/src/STAT.md
@@ -14,7 +14,7 @@ values from 144 to 153 indicating the VBlank period.
 
 ## FF45 — LYC: LY compare
 
-The Game Boy permanently compares the value of the LYC and LY registers.
+The Game Boy constantly compares the value of the LYC and LY registers.
 When both values are identical, the "LYC=LY" flag in the STAT register
 is set, and (if enabled) a STAT interrupt is requested.
 
--- a/src/STAT.md
+++ b/src/STAT.md
@@ -1,4 +1,4 @@
-# LCD Status Register
+# LCD Status Registers
 
 ::: tip TERMINOLOGY
 
--- a/src/STAT.md
+++ b/src/STAT.md
@@ -40,6 +40,8 @@ It is constantly updated.
 
 Bits 3-6 select which sources are used for [the STAT interrupt](<#INT $48 — STAT interrupt>).
 
+## STAT modes
+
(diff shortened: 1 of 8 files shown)


src/STAT.md
```diff
@@ -1,5 +1,17 @@
 # LCD Status Register
 
+## FF44 — LY: LCD Y coordinate \[read-only\]
```
avivace, line 3: I think in any case they should be after the terminology box, as it's meant to be an introductory block or text to satisfy so me knowledge pre-requirements

src/STAT.md
```diff
@@ -6,6 +6,18 @@ A *dot* is the shortest period over which the PPU can output one pixel: is it eq
...
+
+LY indicates the current horizontal line, which might be about to be drawn,
+being drawn, or just been drawn. LY can hold any value from 0 to 153, with
+values from 144 to 153 indicating the VBlank period.
+
+## FF45 — LYC: LY compare
+
+The Game Boy permanently compares the value of the LYC and LY registers.
```
ISSOtm, line 17:
```suggestion
The Game Boy constantly compares the value of the LYC and LY registers.
```

ISSOtm (changes requested):
I dunno, "LCD status registers" seems fine to me, since they are multiple registers that report the PPU's drawing status.

That said, the name in `SUMMARY.md` should also be changed, then.

ISSOtm:
That last commit is indeed a good idea, but then I'm thinking we may want to actually introduce and link to a new section at the bottom to specifically the explanation of the LCD modes. What do you think?

leinacc:
> That last commit is indeed a good idea, but then I'm thinking we may want to actually introduce and link to a new section at the bottom to specifically the explanation of the LCD modes. What do you think?

It sounds like a good idea just seeing those other pages link to STAT by mentioning modes.
I'll place the header above `The LCD controller operates on a 2^22 Hz = 4.194 MHz dot clock.` where it starts to describe them

avivace (approved):

ISSOtm (approved):

ISSOtm:
Thank you!

PR #470, opened 2023-02-21
Fix a typo and an inconsistency in MBC1

Egggggg:
I'm not entirely sure about the typo, there might be some bank number magic that I don't know about, so I can remove that commit if I need to


ISSOtm (approved):

ISSOtm:
Thank you!

PR #474, opened 2023-03-23
Border has an extra scanline, visible in some cases

pinobatch:
One scanline immediately below the SGB border flickers when SGB is busy.  Explain why, and how to hide it

Related Border Crossing commit:
https://github.com/pinobatch/little-things-gb/commit/a3df90fec2ee37001e5f326e7e3fab92b61ec775

### DIFF
--- a/src/SGB_Command_Border.md
+++ b/src/SGB_Command_Border.md
@@ -48,7 +48,8 @@ The map data is sent by VRAM-Transfer (4 KBytes).
 
 ```
  000-6FF  BG Map 32x28 Entries of 16 bits each (1792 bytes)
- 700-7FF  Not used, don't care
+ 700-73F  BG Map 1x28 extra row, 32 entries of 16 bits each (64 bytes)
+ 740-7FF  Not used, don't care
  800-85F  BG Palette Data (Palettes 4-6, 16 little-endian RGB555 colors each)
  860-FFF  Not used, don't care
 ```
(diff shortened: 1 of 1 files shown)


src/SGB_Command_Border.md
```diff
@@ -83,8 +84,18 @@ border defining all unique tiles would have to define this many tiles:
 
 Because the CHR RAM allocated by SGB for border holds only 256 tiles, a full-screen border must repeat at least 281 tiles and a letterboxed border at least 90.
 
+It turns out that 29 rows of the border tilemap sent through PCT_TRN are at least partly visible in some situations.
+The SGB system software sets the border layer's vertical scroll position (BG1VOFS) to 0.
+Because the S-PPU normally displays lines BGxVOFS+1 through BGxVOFS+224 of each layer, this hides the first scanline of the top row of tiles and adds one scanline of the nominally invisible 29th row at the bottom.
+Most of the time, SGB hides this extra line with forced blanking (writing $80 to INIDISP at address $012100).
+While SGB is busy doing some things, such as fading out the border's palette, it neglects to force blanking, making the line flicker on some TVs.
```
avivace, line 91: "busy doing some things" doesn't really feel right here. Are we sure we need such phrasing?

src/SGB_Command_Border.md
```diff
@@ -83,8 +84,18 @@ border defining all unique tiles would have to define this many tiles:
...
+
+To fully eliminate flicker, write a row of all-black tilemap entries after the bottom row of the border ($8700-$873F in VRAM in a PCT_TRN), or at least a row of tiles whose top row of pixels is blank.
+If that is not convenient, such as if a border data format doesn't guarantee an all-black tile ID, you can make the flicker less objectionable by repeating the last scanline.
```
avivace, line 94:
```suggestion
If that is not convenient, such as if a border data format doesn't guarantee an all-black tile ID, you can make the flicker less noticeable by repeating the last scanline.
```

avivace (commented):
Thanks a lot @pinobatch !

src/SGB_Command_Border.md
pinobatch: I don't know the official name of the "some things" that cause SGB to skip copying the GB buffer to VRAM.

pinobatch: In SNESdev, zlago suggested "busy processing (some) game or user commands". oziphanto called it "impossible to define what a word would be unless we actually know what 'some things' explicitly are", and finding what does and doesn't cause the system software to show a busy line would be a separate research project in itself.

pinobatch: The latest commit has a different wording for this.

avivace (approved):

ISSOtm (approved):

PR #475, opened 2023-03-27
add footnote about real DMG clock speed

jkotlinski:
Add footnote that DMG typically runs 50-70 PPM slower than specified.

### DIFF
--- a/src/Specifications.md
+++ b/src/Specifications.md
@@ -20,13 +20,13 @@ td:first-child {
       <td>CPU</td><td colspan="4">8-bit 8080-like Sharp CPU (speculated to be a SM83 core)</td>
     </tr>
     <tr>
-      <td>CPU freq</td><td colspan="2">4.194304&nbsp;MHz</td><td>Depends on revision<sup class="footnote-reference"><a href="#sgb_clk">1</a></sup></td><td>Up to 8.388608&nbsp;MHz</td>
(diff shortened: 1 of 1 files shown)


avivace (approved):

ISSOtm (changes requested):
Hmm, this should be checked on MGB, maybe the drift is different there. Or at least, it should be mentioned that their status is unknown?

jkotlinski:
Unfortunately, I do not have an MGB. I updated the text accordingly.

src/Specifications.md
```diff
@@ -58,6 +58,9 @@ td:first-child {
   </tbody>
 </table>
 
+[^dmg_clk]:
+Real DMG units tend to run about 50-70 PPM slow. Accuracy of other models is unknown. [See this page](https://github.com/jkotlinski/gbchrono) for more details.
```
avivace, line 62:
```suggestion
Real DMG units tend to run about 50-70 PPM slower. Accuracy of other models is unknown. [See this page](https://github.com/jkotlinski/gbchrono) for more details.
```

avivace (approved):

jkotlinski:
Not sure I follow this edit, it is normal language to say that a watch is
running slow.

ons 29 mars 2023 kl. 23:59 skrev Antonio Vivace ***@***.***>:

PR #476, opened 2023-04-03
STAT: fix typo on VBlank duration

deltabeard:
The duration of VBlank is written as 4560 dots which is different to the 456 dots mentioned on line 48.


CasualPokePlayer:
It is 4560 dots as VBlank comprises of 10 scanlines, each scanline is 456 dots. Line 48 is talking about the length of scanline, not the length of VBlank.

PR #485, opened 2023-06-14
Update The_Cartridge_Header.md

Guernouille:
9H - Bottom Up (used in Japan-only game Snowboard Champion)


avivace:
> What do you think about adding a footnote mentioning that extra fact?
> 
> Also, I considered linking to the game's entry on Dat-O-Matic or some similar database, but couldn't find it. If you have such a link, it'd be nice to add, but no biggie if not.

@Guernouille did you see this? :D

PR #486, opened 2023-06-15
audio: change "wavelength" nomenclature to "period value"

pinobatch:
Long ago, the name of the 11-bit value in NR13-NR14, NR23-NR24, and NR33-NR34 was changed from "frequency", which is wrong, to "wavelength", which is less wrong.  It stuck because nobody objected (consensus by crickets).  Some time later, jvsTSX complained that "wavelength" is confusing because the word is more often understood as the reciprocal of spatial frequency, not temporal frequency.  It was pointed out that documents for other PSGs using a frequency divider also use "period" as the name of the parameter.

Add some worked examples to help readers understand the relationship among period value, period, sample rate, and tone frequency.

### DIFF
--- a/src/Audio.md
+++ b/src/Audio.md
@@ -75,11 +75,11 @@ When the length timer reaches 64, the channel is turned off.
 ### Frequency
 
 Music notes and audio waves are typically manipulated in terms of **frequency**[^pitch], i.e. how often the signal repeats per second.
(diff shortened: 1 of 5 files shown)


ISSOtm (approved):
Thank you for this and the accompanying editorial cleanups!

ISSOtm:
Cc @jvsTSX to check if these changes are satisfactory

ISSOtm:
(LGTM'd by jvsTSX on IM.)

PR #487, opened 2023-06-21
ci: switch to "deploy-page" action and stop relying on additional branches

ISSOtm:
This should be more robust and faster (no more branch manipulation), as well as avoiding notifications from pushes to that branch.


avivace (approved):
Amazing, thanks @ISSOtm !

PR #489, opened 2023-06-21
Switch Twitter link to some Nitter instance

ISSOtm:
This kind of reverts 3f1c43e, since now we'll know if that link goes down. Twitter cannot be used, since like that commit mentions, Twitter now blocks automation (thanks to a certain Elongated Muskrat), which includes mdbook-linkcheck.

The choice of Nitter instance is motivated by a few points. First, a list of instances is available from Nitter's GitHub wiki (https://github.com/zedeus/nitter/wiki/Instances).

The first parameter in the choice is the region; there are three (3) "anycast" instances, which may (1) reduce latency for readers of Pan Docs, and (2) reduce the load of Pan Docs' link-checking itself (by distributing across several servers behind the anycast). I think, anyway.

One of these instances is not updated, which is not a good sign for reliability either.

Then, it turns out several instances block HTTP HEAD requests like Twitter (but not GET, suggesting it is intentional); of the remaining two, only the chosen one doesn't.

### DIFF
--- a/src/Power_Up_Sequence.md
+++ b/src/Power_Up_Sequence.md
@@ -207,7 +207,7 @@ Rely on it at your own risk.
 ### Common remarks
 
 The console's WRAM and HRAM are random on power-up.
-[Different models tend to exhibit different patterns](https://notabird.site/CasualPkPlayer/status/1409752977812852736), but they are random nonetheless, even depending on factors such as the ambient temperature.
+[Different models tend to exhibit different patterns](https://web.archive.org/web/20220131221108/https://twitter.com/CasualPkPlayer/status/1409752977812852736), but they are random nonetheless, even depending on factors such as the ambient temperature.
 Besides, turning the system off and on again has proven reliable enough [to carry over RAM from one game to another](https://www.youtube.com/watch?v=xayxmTLljr8), so it's not a good idea to rely on it at all.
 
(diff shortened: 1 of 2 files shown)


ISSOtm:
Right, I forgot to mention this: I also considered https://twiiit.com/, but unfortunately, some of the instances I got in my testing had badly configured TLS, and others blocked HEAD requests; hence picking a known-good one instead.

avivace:
What about https://web.archive.org/web/20220131221108/https://twitter.com/CasualPkPlayer/status/1409752977812852736 ?

src/Power_Up_Sequence.md
```diff
@@ -207,7 +207,7 @@ Rely on it at your own risk.
 ### Common remarks
 
 The console's WRAM and HRAM are random on power-up.
-[Different models tend to exhibit different patterns](https://twitter.com/CasualPkPlayer/status/1409752977812852736), but they are random nonetheless, even depending on factors such as the ambient temperature.
+[Different models tend to exhibit different patterns](https://notabird.site/CasualPkPlayer/status/1409752977812852736), but they are random nonetheless, even depending on factors such as the ambient temperature.
```
avivace, line 210:
```suggestion
[Different models tend to exhibit different patterns](https://web.archive.org/web/20220131221108/https://twitter.com/CasualPkPlayer/status/1409752977812852736), but they are random nonetheless, even depending on factors such as the ambient temperature.
```

avivace (changes requested):

PR #490, opened 2023-06-23
fix times signs

pinobatch:
Change times signs between digits from x to × (U+00D7) and change hexadecimal values in GB context from 0x to $


ISSOtm (changes requested):
You're doing God's work here. Thank you so much.

avivace (approved):

ISSOtm (approved):

PR #501, opened 2023-08-17
Add "rendering overview" document

ISSOtm:
Related but separate from #379.

The existing "rendering" section is renamed to "graphics" to make it more obvious, and so that the new overview document can be called "rendering", as it's intended to describe the actual rendering process.

The motivation for making this page separate, is to gather a lot of the more general info about the rendering process, which is currently scattered across other pages (e.g. STAT) and thus harder to come by.

It placement at the end of the list is because it is not necessary to understand this process to get most work done; understanding of the rendering process only starts getting useful when getting into raster effects, I think.

I am publishing this fairly early so feedback can be gathered on the principle itself; I do intend to keep that PR's scope small, so I can actually carry it to the finish line... heh.


avivace (approved):

ISSOtm:
There might be value in moving the "LCD Status" regs to after this page, since they pertain exclusively to info explained in this overview page.

src/Rendering.md
```diff
@@ -1,85 +1,72 @@
...
+### OBJ penalty algorithm
 
-The _objects_ layer is designed to fill this gap: _objects_ are made of 1 or 2 stacked tiles (8×8 or 8×16 pixels)
-and can be displayed anywhere on the screen.
+Only the OBJ's leftmost pixel matters here, transparent or not; it is designated as "The Pixel" in the following.
 
-::: tip NOTE
+1. Determine the tile (background or window) that The Pixel is within. (This is affected by horizontal scrolling and/or the window!)
+2. If that tile has **not** been considered by a previous OBJ yet:
+   1. Count how many of that tile's pixels are to the right of The Pixel.
```
calc84maniac, line 61: The wording "to the right" sounds exclusive, but I believe the logic here is inclusive.

src/Rendering.md
```diff
@@ -1,85 +1,72 @@
...
-it's impractical for objects that need to move separately, such as the player.
+### OBJ penalty algorithm
 
-The _objects_ layer is designed to fill this gap: _objects_ are made of 1 or 2 stacked tiles (8×8 or 8×16 pixels)
-and can be displayed anywhere on the screen.
+Only the OBJ's leftmost pixel matters here, transparent or not; it is designated as "The Pixel" in the following.
 
-::: tip NOTE
+1. Determine the tile (background or window) that The Pixel is within. (This is affected by horizontal scrolling and/or the window!)
+2. If that tile has **not** been considered by a previous OBJ yet:
```
calc84maniac, line 60: I don't believe the OBJ rendering order has been discussed at this point, so "previous OBJ" may be confusing.

src/Rendering.md
ISSOtm: The subtraction below should have been adjusted to account for that.

src/Rendering.md
ISSOtm: Good catch!

avivace (approved):

src/Audio.md
```diff
@@ -36,7 +36,7 @@ The speaker merges back the two channels, losing the stereo aspect entirely.
 
 The Game Boy's sound chip is called the <abbr title="Audio Processing Unit">APU</abbr>.
 
-The APU runs off the same master clock as the rest of the Game Boy, which is to say, it is fully synced with the CPU and [PPU](<#Rendering Overview>).
+The APU runs off the same master clock as the rest of the Game Boy, which is to say, it is fully synced with the CPU and [PPU](<#Graphics Overview>).
```
ISSOtm, line 39: Actually, "Rendering Overview" is correct, as that is the page that defines the term "PPU".

src/Rendering.md
```diff
@@ -1,85 +1,73 @@
...
+  0  | Waiting until the end of the scanline      | 376 - mode 3's duration              | VRAM, OAM, CGB palettes
+  1  | Waiting until the next frame               | 4560 dots (10 scanlines)             | VRAM, OAM, CGB palettes
 
-### Window
+## Mode 3 length
 
-The window is sort of a second background layer on top of the background.
-It is fairly limited: it has no transparency, it's always a
-rectangle and only the position of the top-left pixel can be controlled.
+During Mode 3, by default the PPU outputs one pixel to the screen per dot; the screen is 160 pixels wide, so the minimum Mode 3 length is 160 + 12[^first12] = 172 dots.
```
ISSOtm, line 43:
```suggestion
During Mode 3, by default the PPU outputs one pixel to the screen per dot, from left to right; the screen is 160 pixels wide, so the minimum Mode 3 length is 160 + 12[^first12] = 172 dots.
```

src/Rendering.md
ISSOtm: Maybe a footnote attached to "previous"?

> Since pixels are emitted from left to right, OBJs overlapping the scanline are considered from leftmost <link to OAM page, "X position" anchor> to rightmost, with ties broken by the OAM order (lowest address first).

src/Rendering.md
```diff
@@ -1,85 +1,73 @@
...
+1. Determine the tile (background or window) that The Pixel is within. (This is affected by horizontal scrolling and/or the window!)
+2. If that tile has **not** been considered by a previous OBJ yet:
+   1. Count how many of that tile's pixels are to the right of The Pixel.
+   2. Subtract 3.
+   3. Incur this many dots of penalty, or zero if negative (from waiting for the BG fetch to finish).
+3. Incur a flat, 6-dot penalty (from fetching the OBJ's tile).
+
+**Exception**: an OBJ with an OAM X position of 0 (thus, completely off the left side of the screen) always incurs a 11-cycle penalty, regardless of `SCX`.
+
+TODO: a diagram of some examples would probably help this be much clearer!
```
ISSOtm, line 69: I think we should merge this PR as-is, and open an issue to get this resolved separately.

ISSOtm: Obviously, if we agree on that.

```suggestion
```

src/Audio.md
avivace: Okay, let's switch back to Rendering then?

src/Rendering.md
ISSOtm: I guess it's not, because then the maximum is 7 pixels; minus 3 = 4; plus the "flat" 6 = 40, but the maximum stated above is 11. So probably that the line below should be "Subtract 2." instead?

PR #506, opened 2023-09-30
Add wrap behavior to the SCX and SCY registers

### LINKED ISSUE #502: lcd-position-and-scrolling: wrap behavior
It's my understanding that when the value of SCX/SCY are large so that the coordinate would exceed 256, it will wrap over to 0.

alloncm:
Closes #502 

I'm not sure whether the pseudocode is helpful or just clatters the document, Ill be happy to hear what you think.


src/Scrolling.md
```diff
@@ -4,11 +4,19 @@
 These registers can be accessed even during Mode 3, but modifications may not take
 effect immediately (see further below).
 
-## FF42–FF43 — SCY, SCX: Viewport Y position, X position
+## FF42–FF43 — SCY, SCX: Background viewport Y position, X position
```
ISSOtm, line 7: Indeed, I agree this name is better!
Note, however, that changing the section's name created [some broken links](https://github.com/gbdev/pandocs/pull/506/checks#step:7:211), which must be updated as well.

src/Scrolling.md
```diff
@@ -4,11 +4,19 @@
...
 
-Those specify the top-left coordinates of the visible 160×144 pixel area within the
+Specify the top-left coordinates of the visible 160×144 pixel area within the
```
ISSOtm, line 9: Maybe "these two registers", or "this pair of registers"? I don't think imperative works very well here.

src/Scrolling.md
```diff
@@ -4,11 +4,19 @@
...
 256×256 pixels BG map. Values in the range 0–255 may be used.
 
+In case the calculated bottom-right coordinates are larger than 255 it will wrap around towards the top-left corner.
+
+Pseudocode for the calculated bottom-right coordinates would be:
+```
+bottom  := (SCY + 144) & 255
+right   := (SCX + 160) & 255
+```
```
ISSOtm, line 18: I feel like a discussion of the bottom and right edge of the screen with the two formulas would flow better.

Additionally, mind the fencepost error: only 143 and 159 must be added, or the coordinates are of the pixel *one past* both edges.

ISSOtm (changes requested):
I agree there's worth in mentioning the wrap-around behaviour explicitly.

src/Scrolling.md
avivace: I think "These two registers" sounds better, yeah

avivace, line 9:
```suggestion
These two registers specify the top-left coordinates of the visible 160×144 pixel area within the
```

avivace (changes requested):

src/Scrolling.md
alloncm: I rephrased This section into a discussion and fixed the formulas

src/Scrolling.md
```diff
@@ -4,14 +4,17 @@
...
 
-## FF42–FF43 — SCY, SCX: Viewport Y position, X position
+## FF42–FF43 — SCY, SCX: Background viewport Y position, X position
 
-Those specify the top-left coordinates of the visible 160×144 pixel area within the
+These two registers specify the top-left coordinates of the visible 160×144 pixel area within the
 256×256 pixels BG map. Values in the range 0–255 may be used.
 
+The PPU calculates the bottom-right coordinates of the viewport with those formulas: `bottom := (SCY + 143) & 255` and `right := (SCX + 159) & 255`.
+As we can infer, in case the values are larger than 255 they will wrap around towards the top-left corner of the screen.
```
ISSOtm, line 13:
```suggestion
The PPU calculates the bottom-right coordinates of the viewport with those formulas: `bottom := (SCY + 143) & 255` and `right := (SCX + 159) & 255`.
As we can infer, in case the values are larger than 255 they will wrap around towards the top-left corner of the tilemap.
```

...though the wording here sounds a little confusing/confused to me. May I suggest the following:

```suggestion
The coordinates of the bottom-right on-screen pixel are `Y = (SCY + 143) % 256` and `X = (SCX + 159) % 256`.
As suggested by the modulos, the tilemaps effectively "wrap around" both horizontally and vertically.
```

A diagram (something like the OAM chapter's OBJ positions, or screenshot(s?) of any VRAM viewer) may also convey the point more succinctly. What do you think?

src/Scrolling.md
```diff
@@ -4,14 +4,17 @@
...
+
 ## FF4A–FF4B — WY, WX: Window Y position, X position plus 7
 
-Specify the top-left coordinates of [the Window](#Window).
+These two registers specify the top-left coordinates of [the Window](#Window).
```
ISSOtm, line 17: Good catch that the change applied to here as well!

```suggestion
These two registers specify the on-screen coordinates of [the Window](#Window)'s top-left pixel.
```

I think this wording is clearer with regards to the coordinate space, and what the coordinates refer to?

ISSOtm (changes requested):
Yes, this already looks much better!

src/Scrolling.md
```diff
@@ -4,14 +4,20 @@
...
+## FF42–FF43 — SCY, SCX: Background viewport Y position, X position
 
-Those specify the top-left coordinates of the visible 160×144 pixel area within the
+These two registers specify the top-left coordinates of the visible 160×144 pixel area within the
 256×256 pixels BG map. Values in the range 0–255 may be used.
 
+The PPU calculates the bottom-right coordinates of the viewport with those formulas: `bottom := (SCY + 143) % 256` and `right := (SCX + 159) % 256`.
+As suggested by the modulo operations, in case the values are larger than 255 they will "wrap around" towards the top-left corner of the tilemap.
+
+### Visual example:
```
ISSOtm, line 15:
```suggestion
Example from the homebrew game *Mindy's Hike*:

```

I think it's more useful to provide context for the screenshot (though a link the game is not necessary), and a blank line ensures we have correct markup.

I would even encourage using `<figure>` for this:
```md
<figure><figcaption>

Example from the homebrew game *Mindy's Hike*:

</figcaption>

...image...

</figure>
```

ISSOtm (changes requested):
Great! Only one minor comment left, and I'd greenlight this.

Also, the image's file name has a typo; no biggie, but I think it's worth fixing while we're at it.

ISSOtm (approved):
Thank you!

PR #508, opened 2023-10-23
Document custom markup in CONTRIBUTING instead

ISSOtm:
As discussed in https://github.com/gbdev/pandocs/pull/334#discussion_r1344805255


avivace (approved):
Thanks Eldred!

PR #511, opened 2023-11-25
Switch "CPU instruction set" to document encoding instead

### LINKED ISSUE #456: CPU Instruction Set Encodings
Currently, it seems that many of the instructions are "generalized" to avoid the need to write individual instructions for each register. However, there are some encodings for instructions that just say "xx", which doesn't really help with decoding those instructions or even specifying what valid values would be for that specific instruction.

For example, the `ld r, nn` instruction says the encoding is `xx nn`. But valid values for "xx" in this case are `0x06`, `0x16`, `0x26`, `0x0E`, `0x1E`, `0x2E`, and `0x3E`, which aren't really mentioned outside of the opcodes table which is buried in the References section.

I'm not sure what the best style for expanding on this would be, but as it is, the Encoding column seems to not be very useful unless it has the entire encoding already. Maybe a link to the opcodes table might be better served at the top of this page?

ISSOtm:
Fixes #456.

[As mentioned there](https://github.com/gbdev/pandocs/issues/456#issuecomment-1386463531), optables provides a very compact view of instructions, but may not help discerning patterns; whereas gbz80(7) describes the instructions themselves, but does not work well as a cheat sheet.

This PR attempts to document what neither of those two resources do: the instructions' encoding, in a manner closer to how the CPU decodes them. The patterns can be gleaned from optables, especially in the octal view, but not everyone may connect the dots.

Cycle counts, behaviour descriptions, etc. have been removed, since that information can be found in gbz80(7) and optables; and removing that data makes this page more compact, and thus I think easier to navigate and grasp.

Instructions have been grouped thematically, and those groups roughly ordered—though blocks 0 and 3 are quite resistant to any consensus ordering.

Anyone and everyone is welcome to share their feedback on the decisions made about the content and the presentation. Thank you!


ISSOtm:
cc people involved in #456: @Kenjugs @iximeow @54UL. What do you think?

avivace (approved):

src/CPU_Instruction_Set.md
```diff
@@ -1,151 +1,170 @@
...
+{{#bits 8 <
+	"r8"      0:"<code>b</code>" 1:"<code>c</code>" 2:"<code>d</code>" 3:"<code>e</code>" 4:"<code>h</code>" 5:"<code>l</code>" 6:"<code>[hl]</code>" 7:"<code>a</code>" ;
+	"r16"     0:"<code>bc</code>" 1:"<code>de</code>" 2:"<code>hl</code>" 3:"<code>sp</code>" ;
+	"r16stk"  0:"<code>bc</code>" 1:"<code>de</code>" 2:"<code>hl</code>" 3:"<code>af</code>" ;
+	"r16mem"  0:"<code>bc</code>" 1:"<code>de</code>" 2:"<code>hl+</code>" 3:"<code>hl-</code>" ;
+	"cond"    0:"<code>nz</code>" 1:"<code>z</code>" 2:"<code>nc</code>" 3:"<code>c</code>" ;
+	"b3"      0-7:"A 3-bit bit index" ;
+	"imm8"    0-7:"The following byte" ;
+	"imm16"   0-7:"The following two bytes, in little-endian order" ;
+}}
```
iximeow, line 16: nit: `vec3` isn't described here, but it's the operand type written out for `rst`. clearly rst doesn't take a `b3` or anything like that, so that makes sense. but it'd be helpful to have some words like `"vec3" 0-4: "The reset vector number. Multiply by 8 for the vector's address"` in this table.

iximeow:
broadly this seems really nice! i have the one nit, otherwise :+1: . you have a TODO about links to optables, are you thinking a paragraph of preface about those other resources? or something more deeply connected to the revised docs?

ISSOtm:
I wanted to write a paragraph explaining the choice of what that page documents and what it doesn't, as well as giving links to other complementary resources.

Glad you like it! ^^

src/CPU_Instruction_Set.md
ISSOtm: Yeah, that's indeed an accidental omission. I would suggest the following wording, though: `"rst's target address, divided by 8"`?

ISSOtm:
The paragraph linking to other resources was easy to write, but an attempt at a key proved tremendously more difficult; I can't help but think I've fumbled it. But at least it's something that can be discussed.

ISSOtm:
I'm planning to merge this next week without further activity.

avivace (approved):

PR #517, opened 2023-12-13
APU channel 3 docs should refer to channel 3 registers

iximeow:
just a simple copypaste error :)


iximeow:
oh actually the trigger description reference the wrong register too. i'll edit just a bit more..

ISSOtm (approved):
Can you tell reformatting the audio docs got really long and tiring? :')
Thanks for the fix. ^^

PR #518, opened 2023-12-14
APU: clarify which bits of NR41 are actually used

iximeow:
the docs that were present gave the misleading impression that NR41 sets an 8-bit timer similar to NR31, and that channel 4's length is counted up to 256 like NR31. but channel 4 counts length up to 64 more like channels 1 and 2, with an initial length set from the low six bits of this register.

(spotted this by going through the above logic, noticing that channel 4 seemed to run too long, and comparing against a few other emulators. i don't actually know what happens with the upper two bits here, if they're ignored, or what.)


ISSOtm (commented):
That looks like a copy-paste mistake from NR31, the upper two bits are indeed unused. Thanks!

PR #519, opened 2023-12-14
Remove confusing and redundant wording

tobiasvl:
- Remove confusing wording that seemed to imply that timer interrupts are executed immediately (regardless of priority)
- Remove redundant information about the IF register, which is not present in the descriptions of other interrupts

### DIFF
--- a/src/Interrupt_Sources.md
+++ b/src/Interrupt_Sources.md
@@ -52,10 +52,8 @@ hidden by the text box.
 
 ## INT $50 — Timer interrupt
 
-Every time that the timer overflows (that is, when [TIMA](<#FF05 — TIMA: Timer counter>) exceeds $FF),
-an interrupt is requested by setting bit 2 in the IF register
-($FF0F). As soon as that interrupt is enabled, the CPU will execute it by
-calling the timer interrupt vector at $0050.
+The timer interrupt is requested every time that the timer overflows
+(that is, when [TIMA](<#FF05 — TIMA: Timer counter>) exceeds $FF).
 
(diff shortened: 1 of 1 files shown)


src/Interrupt_Sources.md
```diff
@@ -52,10 +52,8 @@ hidden by the text box.
 
 ## INT $50 — Timer interrupt
 
-Every time that the timer overflows (that is, when [TIMA](<#FF05 — TIMA: Timer counter>) exceeds $FF),
-an interrupt is requested by setting bit 2 in the IF register
-($FF0F). As soon as that interrupt is enabled, the CPU will execute it by
-calling the timer interrupt vector at $0050.
+The timer interrupt is requested every time that the timer overflows
```
ISSOtm, line 55: I think making the words "is requested" into a link to either `IF`'s documentation or a more general description of the IRQ process (is there one?) would be appropriate in fulfilling the role that the rightfully removed text was. What do you say?

avivace: Agree!

tobiasvl:
Then maybe the link should be added to every mention of an interrupt being requested? It seems a bit arbitrary to explain what it means for a timer interrupt to be requested, when the VBlank and STAT interrupts have already been discussed above with no explanation of what an IRQ entails.

ISSOtm:
Oh, yes, sure! I only focused on the PR's changes, but if you think the change should be made elsewhere as well, then by all means go ahead!

ISSOtm (approved):
Looks good to me now!

avivace (approved):

PR #523, opened 2023-12-17
Accomodate out-of-repo builds

### LINKED ISSUE #514: Allow building out-of-tree
This was requested by @mid-kid, who packages Pan Docs as a Gentoo package for offline viewing. Currently, this chokes by default because we expect to be able to call `git show` to retrieve the commit info for the front page.

To do:

- [ ] Figure out what to display on the front page when building out-of-tree
- [ ] Decide on an interface to enable out-of-tree builds
- [ ] Figure out a good place to document that last item

For the second bullet, I think we can rely on a key in mdBook's `output.pandocs` table, which would disable Git scanning if set to `true`. This can even be configured via environment variables e.g. `MDBOOK_output__pandocs__out_of_tree=true mdbook build` to avoid patching the sources.

For the last item, I think `DEPLOY.md` is the best place?

ISSOtm:
Fix #514, cc @mid-kid for the interface/behaviour.

### DIFF
--- a/preproc/src/main.rs
+++ b/preproc/src/main.rs
@@ -9,9 +9,10 @@
 
 use std::collections::HashMap;
 use std::io;
+use std::path::Path;
 use std::process;
 
-use anyhow::Context;
+use anyhow::{anyhow, Context};
 use clap::{App, Arg, ArgMatches, SubCommand};
 use mdbook::book::Book;
 use mdbook::errors::Error;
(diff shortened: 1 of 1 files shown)


avivace (approved):

mid-kid:
Sounds good to me! What would the correct invocation of mdbook build be?

ISSOtm:
Hopefully you're able to figure that out on your own? (This is not a taunt, but rather I tried to make this reasonably easy for other packagers who may not be able to ask us for help, and I want to check that.)

mid-kid:
If I got this right, it's `MDBOOK_preprocessor__pandocs__out_of_repo=true mdbook build`?

ISSOtm:
Yep, that matches my understanding. Hopefully both of us are correct :D

mid-kid:
Yeah, :+1:

PR #527, opened 2024-01-02
Update CI to stop using Node 12 actions

ISSOtm:
See https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/


ISSOtm:
I created the PR to ensure that CI would be green, so I'm merging without review.

PR #530, opened 2024-01-03
Update requirements.txt file for changes in #452

### LINKED ISSUE #322: Allow SVG to be affected by theme

glinesbdev:
Supersedes and closes #452
Original OP follows:

---

It may seem a bit convoluted, but it works:

https://user-images.githubusercontent.com/14352721/201120995-a3bf61d5-e9f6-472f-ba4a-b150bfb8c9dd.mp4

Some more stuff is needed to merge this but we should be already at a pretty good point:

- [x] move from pygal to matplotlib #449
- [x] make the output SVG use the CSS variables instead of hardcoded colross
- [ ] update requirements.txt
- [ ] move the call to the plot generation from the `renderer` to the `preproc` step of mdbook, otherwise the output SVG files cannot be inline injected
	- We should consider just calling a python script, instead of having the single calls for every plot like it is now [here](https://github.com/gbdev/pandocs/blob/master/book.toml#L16). @ISSOtm opinions?
- [ ] replace the two img tags with #import in the MBC5 article so the svg is actually injected in the HTML

Fixes #322

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 /docs/
 /target/
 /env/
+/generated/
+__pycache__/
--- a/book.toml
+++ b/book.toml
@@ -10,15 +10,19 @@ build-dir = "docs"
 create-missing = true # This is kept for convenience, but CI sets it to false
 use-default-preprocessors = false
 
-# Custom preprocessor for internal links processing
-[preprocessor.pandocs]
-command = "cargo run -p pandocs-preproc --locked --release --"
-after = [ "links" ]
+[preprocessor.graph_gen]
+command = "src/imgs/src/preproc.py"
+before = [ "links" ] # This generates some of the files that get `{{#include}}`d
 
 # `{{#include }}` etc. resolution
 [preprocessor.links]
 
-# Custom back-end for our custom markup
+# Custom preprocessor for internal links processing and other custom markup
+[preprocessor.pandocs]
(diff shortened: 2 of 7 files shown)


ISSOtm:
Alright, I have edited the OP to replicate the original PR's. That means what's left is only mdBook "wiring up", which falls under my responsibility... so I'll have to do that then. :P

Thank you @glinesbdev!

avivace (dismissed):
This is amazing, @glinesbdev . Putting some sense into what I wrote and making this modular will be super useful also when not generating graphs but when we'll want to make other svgs "theme-able". Thanks a lot!

I'm trying the command line usage but when I run:

```
python graph_render.py MBC5_Rumble_Mild.csv "MBC5_Rumble_Mild"
```

The svg is printed out in the stdout and then two svg are actually generate, a "test.svg" and an "output.svg"..

Is this expected?

ISSOtm (approved):
Thank you @glinesbdev! My apologies for taking this long to do the mdBook part of this PR.

avivace (approved):
Amazing work. Thanks for cleaning up my initial work and hooking it up @glinesbdev and @ISSOtm !

Issue #529, opened 2024-01-03
Make bits-description table separations more visible


vizigr0u:
The tables showing the bit layout can be a bit hard to read sometimes because of the low contrast and very thin lines separating each field.
e.g. 
![image](https://github.com/gbdev/pandocs/assets/1981001/40c908a2-79ed-4b14-940e-bb86f2fa5d8e)
![image](https://github.com/gbdev/pandocs/assets/1981001/636b741e-6dc7-46b9-ab4a-5e05183ff3e9)

(Issue exists for every theme)

Example of what a higher contrast would look like on the navy skin:
![image](https://github.com/gbdev/pandocs/assets/1981001/018be8d4-2340-4724-917e-6d42b5e4c4e5)

 (changing `--table-border-color` to `hsl(210, 25%, 33%)` from `hsl(210, 25%, 13%)`)

ISSOtm:
Yeah, seems fair. Can you open a PR?

vizigr0u:
> Yeah, seems fair. Can you open a PR?

I quickly searched the repo for `--table-border-color` and couldn't find it so I figured it would be much faster for someone with the setup on their machine

ISSOtm:
Yeah, those are [provided by mdBook](https://github.com/rust-lang/mdBook/blob/090eba0db5026982c72f20352947cb1438c9f008/src/theme/css/variables.css); those can be overridden (e.g. `.ayu .bit-descrs`) in `custom/style.css`.

vizigr0u:
gotcha, I will try to get the repo setup on my machine and I'll make a PR

ISSOtm:
Fixed by #531.

PR #531, opened 2024-01-03
Improve contrast of bit-descrs borders

vizigr0u:
Proposed solution to improve border contrast for #529
image
image
image
image
image

### DIFF
--- a/custom/style.css
+++ b/custom/style.css
@@ -178,24 +178,8 @@ mfrac msup :not(:first-child) {
 }
 
 /* Improve contrast on bit-descrs borders */
-.ayu .bit-descrs {
-  --table-border-color: hsl(210, 25%, 30%);
-}
-
-.coal .bit-descrs {
-  --table-border-color: hsl(200, 7%, 30%);
-}
-
-.light .bit-descrs {
-  --table-border-color: hsl(0, 0%, 70%);
-}
-
-.navy .bit-descrs {
-  --table-border-color: hsl(226, 23%, 40%);
-}
-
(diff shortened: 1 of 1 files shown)


custom/style.css
```diff
@@ -177,6 +177,27 @@ mfrac msup :not(:first-child) {
...
+.light .bit-descrs {
+  --table-border-color: hsl(0, 0%, 70%);
+}
+
+.navy .bit-descrs {
+  --table-border-color: hsl(226, 23%, 40%);
+}
+
+.rust .bit-descrs {
+  --table-border-color: hsl(60, 9%, 60%);
```
ISSOtm, line 198: Do all of those match the `--table-header-bg`?

If so, perhaps all of these could be collapsed into a single
```css
.bit-descrs {
  --table-border-color: var(--table-header-bg);
}
```
rule, which would be simpler and also track any upstream changes.

ISSOtm (commented):
Thank you for taking the time! I agree the separations are more important in tables that may contain spanning cells.

custom/style.css
vizigr0u: haven't though about it, that might look even better, I'll try and update

vizigr0u:
@ISSOtm I tested and committed the change you suggested, it looks better than the colors I had arbitrarily chosen ;)

ISSOtm:
Awesome! Thanks for taking the time to see this through :)

(Skipping CI because this is just a CSS change.)

avivace (approved):

PR #532, opened 2024-01-04
Generate multiple svg graphs

glinesbdev:
This work is based on #530 and should be merged in after that PR.

Example Usage:

```python
import graph_render

graph1 = graph_render.GraphData(['Time (ms)'], 'graph1.csv', 'First Graph')
graph2 = graph_render.GraphData(['Demographic', 'Currency'], 'graph2.csv', 'Second Graph')

graph_render.gen_graphs([graph1, graph2])
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,4 @@
-pygal==3.0.0
+beautifulsoup4==4.12.2
+lxml==5.0.0
+matplotlib==3.8.2
+pandas==2.1.4
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-beautifulsoup4==4.12.2
-lxml==5.0.0
-matplotlib==3.8.2
-pandas==2.1.4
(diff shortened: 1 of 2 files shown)


avivace:
I'm on this, thanks @glinesbdev !

avivace (changes requested):
This is amazing, @glinesbdev . Putting some sense into what I wrote and making this modular will be super useful also when not generating graphs but when we'll want to make other svgs "theme-able". Thanks a lot!

I'm trying the command line usage but when I run:

```
python graph_render.py MBC5_Rumble_Mild.csv "MBC5_Rumble_Mild"
```

The svg is printed out in the stdout and then two svg are actually generate, a "test.svg" and an "output.svg"..

Is this expected?

avivace:
Done in https://github.com/gbdev/pandocs/blob/master/src/imgs/src/preproc.py (PR https://github.com/gbdev/pandocs/pull/530)

PR #534, opened 2024-01-13
The big cycle unit rework

SonoSooS:
I've read *all* pages, searching for instances where clock units in particular were either confusing, ambiguous, misleading, or simply just used wrong. This PR aids to fix all of these.

Notes:
- Four Player Adapter: even though serial runs locked to M-cycle boundary, I did not edit the bits-per-second formula, as I can't reword it due to not being able to test it myself to verify *why* the formula is that. Also 6 % 4 != 0, and that is definitely wrong.
- IR: Few hundred *what* cycles? With no other units of reference nearby, and not having a real CGB to test it myself, I can't fix this alone.
- OAM DMA Transfer: I edited all instances of cycle to M-cycle, but it looks bad tbh
- Power Up Sequence: what one cycle faster for SGB?
- Rendering: my wording may be a bit awkward (standardize the rest of this page too? dots vs. dot cycles)
- Timer and Divider Registers: I had to alter the table, I'm not happy how I had to do it, but considering how ridiculously ambiguous the "CPU Clock" phrase is, I'd rather write it this way. Having to clarify around the table instead of *in* the table I think fails approachability. As for the DIV register part, personally I think it's out of the scope of this PR to rewrite it completely, so I've put temporarily warnings there during the transitional period. Annoyingly, there is an annoying line break in the table I don't know how to fix. Must be fixed definitely for sure, I don't think it should stay like that... Oh also, I've updated the rest of the page to the current typographic standards while at it.

TODO (perhaps for other PRs if these are outside of the scope of this PR):
- Timer Obscure Behavior: DIV ticks at M-cycle intervals, those lower two bits don't exist. A fix to the diagrams is needed.
- Timer and Divider Registers: [$FF04] is only the visible region of DIV, real DIV increases every M-cycle. Too much detail? Not enough detail? At least make it so that the visible portion is increased every 64 M-cycle, instead of having to clarify paragraphs later that it runs twice as fast in Double Speed mode.


src/Timer_and_Divider_Registers.md
```diff
@@ -51,13 +51,13 @@ due to a timer overflow, the old value is transferred to TIMA.
...
     <thead>
-      <tr><th rowspan=2>Clock select</th><th rowspan=2>Base clock</th><th colspan=3>Frequency (Hz)</th></tr>
+      <tr><th rowspan=2>Clock select</th><th rowspan=2>Increment every</th><th colspan=3>Frequency (Hz)</th></tr>
       <tr><th>DMG, SGB2, CGB in single-speed mode</th><th>SGB1</th><th>CGB in double-speed mode</th></tr>
     </thead><tbody>
-      <tr><td>00</td><td>CPU Clock / 1024</td><td>  4096</td><td>  ~4194</td><td>  8192</td></tr>
-      <tr><td>01</td><td>CPU Clock / 16  </td><td>262144</td><td>~268400</td><td>524288</td></tr>
-      <tr><td>10</td><td>CPU Clock / 64  </td><td> 65536</td><td> ~67110</td><td>131072</td></tr>
-      <tr><td>11</td><td>CPU Clock / 256 </td><td> 16384</td><td> ~16780</td><td> 32768</td></tr>
+      <tr><td>00</td><td>256 M-cycle  </td><td>  4096</td><td>  ~4194</td><td>  8192</td></tr>
```
ISSOtm, line 57:
```suggestion
      <tr><td>00</td><td>256 M-cycles </td><td>  4096</td><td>  ~4194</td><td>  8192</td></tr>
```

src/Timer_and_Divider_Registers.md
```diff
@@ -51,13 +51,13 @@ due to a timer overflow, the old value is transferred to TIMA.
...
+      <tr><td>01</td><td>4 M-cycle </td><td>262144</td><td>~268400</td><td>524288</td></tr>
```
ISSOtm, line 58:
```suggestion
      <tr><td>01</td><td>4 M-cycles</td><td>262144</td><td>~268400</td><td>524288</td></tr>
```

src/Timer_and_Divider_Registers.md
```diff
@@ -51,13 +51,13 @@ due to a timer overflow, the old value is transferred to TIMA.
...
+      <tr><td>10</td><td>16 M-cycle </td><td> 65536</td><td> ~67110</td><td>131072</td></tr>
```
ISSOtm, line 59:
```suggestion
      <tr><td>10</td><td>16 M-cycles</td><td> 65536</td><td> ~67110</td><td>131072</td></tr>
```

src/Timer_and_Divider_Registers.md
```diff
@@ -51,13 +51,13 @@ due to a timer overflow, the old value is transferred to TIMA.
...
+      <tr><td>11</td><td>64 M-cycle </td><td> 16384</td><td> ~16780</td><td> 32768</td></tr>
```
ISSOtm, line 60:
```suggestion
      <tr><td>11</td><td>64 M-cycles</td><td> 16384</td><td> ~16780</td><td> 32768</td></tr>
```

src/Timer_Obscure_Behaviour.md
```diff
@@ -1,5 +1,29 @@
 # Timer obscure behaviour
 
+<!-- TODO: actually fix the diagrams -->
```
ISSOtm, line 3: This comment is to keep track of that TODO.

Hopefully the diagrams can be overhauled as part of this PR, but I'd rather have the band-aid and open an issue rather than hold this PR back, if that came to be.

ISSOtm (changes requested):
Thanks a lot, Sono! I appreciate that you took the time to survey the whole doc, and corrected some inconsistencies.

Arguably, what's left is explaining those two terms *somewhere*, but that would be part of a "overview"/"introduction" PR, so it's fine not to deal with it here.

src/CGB_Registers.md
```diff
@@ -168,7 +168,7 @@ would be recommended to use Single Speed whenever possible.
 
 In Double Speed Mode the following will operate twice as fast as normal:
 
-- The CPU (2.10 MHz, so 1 cycle = approx. 0.5 µs)
+- The CPU (2.10 MHz, so 1 T-cycle = approx. 0.5 µs)
```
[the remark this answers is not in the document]
SonoSooS: Oh, I can't math, my bad!

src/Timer_Obscure_Behaviour.md
avivace: Is this done now?

avivace (changes requested):
Are those handwritten?
As a general practice, to have some maintainability, those graphs should be drawn on something like draw.io and both the "project" file (more easily editable) and the svg should be uploaded and kept in the repository.

Also, there are lot of useless arrow "diversion" caused by elements not aligned, such as:
image

image

ISSOtm:
Those are, indeed, hand-written. No editing tool I'm aware of supports "themed" SVGs. I understand the theoretical maintenance loss, but hear me out:
- Embedded text is still easy to modify (including to localize),
- SVG is a well-known format,
- Worst case the "next person" can open such a SVG in Inkscape or other editing program, and lose the theming from the conversion but keep the information otherwise intact.[^fallbacks]
- Finally, those diagrams rarely get edited; and when they do, they can be redone from scratch: even hand-crafting those took me only a day each, tops, so someone using a GUI should have far less trouble. (And I've been thinking of ways to simplify them further, which would make them more maintainable *and* faster to create.)

All of these are why I personally consider the tradeoff acceptable when weighed against the gains, chiefly in accessibility and hackability.

[^fallbacks]: Existing SVGs did not have that until 5eaf6f82e6d3fb2d5b23fa3cad10d047708898b9 because I did not think of providing "fallback" values to the `var()` directives; a few diagrams in outstanding PRs do not yet have this, though; I'll have to get to that.

ISSOtm:
> Also, there are lot of useless arrow "diversion" caused by elements not aligned, such as:

I paid a lot of attention trying to avoid bending the arrows, which required a lot of tweaking overall. Both of these are actually intentional, so I will lay out my rationale for you to judge.

> image

This is actually so that the "Timer tick" bubble lies at the same place on both the DMG and CGB diagrams. This mattered more in a tentative design I made where the two diagrams were shown on separate (responsive) columns, side by side. Now that this design has been dropped (the diagrams are too horizontal, they were terribly squished), I guess the bubble can be moved down a little?

> image

This was done because the two elements the arrows originate from are too close, necessitating at least one bend. Then, I felt that having only a single bend looked awkward, and made the two actions look asymmetrical, and thus having different priority. (Am I reading too much into it?)

I considered another workaround in making one of the arrows arrive to the register from the top or bottom instead of the side; but looking at the diagrams, vertical arrows designate single bits of registers, whereas horizontal ones designate the register as a whole... which made such a workaround impractical.

*Another* design I considered was pointing either of the two arrows (probably the "Reset" one) at the register title itself; only problem is, `DIV` *specifically* resists such a design, since the low bits are really unnamed[^sys_counter], so there's nothing to point to.
That said, now that I thought more about it, it seems possible to perhaps convey that name within the diagram itself (making the right slanted edge of `DIV` dotted, and otherwise extending the grouping to all 14 bits and inserting the name). What do you think?

[^sys_counter]: The "main text" calls them *and `DIV`* "System counter", though I think "M-cycle counter" is more appropriate due to what it's counting; cc @SonoSooS for opinion.

ISSOtm:
And to answer a question that was asked on IM: yes, I'd say this PR should be considered content-complete; there is one last diagram not converted on the "Timer Obscure Behavior" page, but during the conversion process I realised that it is plain wrong, and fell into a rabbit hole of researching and correcting behaviour.

Which is to say, I'd rather keep *those* corrections to a separate PR, working on it after *this* one is deemed acceptable. And I'd rather not expend the effort to convert a diagram before modifying it into something correct, even if it means temporarily having a style discrepancy and keeping some wrong info. (We already have many differing styles throughout the document anyway; big whoop.)

src/Timer_Obscure_Behaviour.md
ISSOtm: Almost all by b7f4844777eb73c8774fe95fbfcfd7a056e47f45; see https://github.com/gbdev/pandocs/pull/534#issuecomment-1897102969 for why the "almost".

tl;dr yes

src/Timer_Obscure_Behaviour.md
```diff
@@ -1,41 +1,67 @@
...
 
+<!-- TODO: fix the wording even more -->
+
+:::warning Use of names referring to DIV
+
+When detailing relations in the paragraphs below, 
+ DIV is just the visible part of the system counter.
+
+The system counter is constantly incrementing every M-cycle
+ while the CPU is not in STOP mode.
```
ISSOtm, line 11:
```suggestion
:::warning System counter

DIV is just the visible part of the **system counter**.

The **system counter** is constantly incrementing every M-cycle, unless the CPU is in [STOP mode](<#Using the STOP Instruction>).
```

ISSOtm (changes requested):
Alright, as far as my opinion goes, there's only the single TODO left to take care of. I expect someone other than me will want to review the few changes I have introduced, though.

src/Timer_Obscure_Behaviour.md
ISSOtm: I don't know if it's easily understandable, but it seems reasonable to me. (At some point, we have to hazard guesses.) "System counter" isn't set in stone IMO, as I'd rather come up with new terms if they're clearer and feasible to change (unlike e.g. hardware reg names); but this name clears the clarity bar IMO.

And indeed, a notice/tip would be preferable to a warning; amended the suggestion accordingly.

src/Timer_Obscure_Behaviour.md
```diff
@@ -1,5 +1,29 @@
...
+As DIV is incremented every M-cycle in clockstep with the CPU, 
+ the lower 2 bits of DIV in the illustrations don't actually exist.
+
+:::
+
+<!-- TODO: fix the wording if necessary -->
+
+:::warning Use of names referring to DIV
+
+DIV is actually the entire internal counter which is incremented every M-cycle.
```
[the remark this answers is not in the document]
ISSOtm: Yeah, "system counter" is at least slightly more specific than "internal counter", so I prefer the former.

src/Timer_Obscure_Behaviour.md
avivace: Seems reasonable to me as well. I also think a standard "tip" type fits better than the warning.

ISSOtm (approved):
Alright, now this PR looks good to me!

Considering the last month of inactivity, and that the last change had some agreement from @avivace, I'll be merging this PR. Follow-ups can happen in separate ones.

PR #536, opened 2024-01-14
explain more about the sgb border's colour limits

sabra55:


ISSOtm (commented):
I see what you're trying to say, and I agree with the intent.

However, I feel like this may be confusing, as the GB screen is still displayed by the SNES (just via a framebuffer); most of the additional colour capabilities come from the additional bit depth. I'm not sure how to convey that well?

sabra55:
> I see what you're trying to say, and I agree with the intent.
> 
> However, I feel like this may be confusing, as the GB screen is still displayed by the SNES (just via a framebuffer); most of the additional colour capabilities come from the additional bit depth. I'm not sure how to convey that well?

"The game screen, however, is still displayed by the GB hardware and is still limited to 4 colo(u)rs per tile rather than 16, like the border." maybe?

ISSOtm:
I think that's better; but how about reversing the order?

> A border can be much more colorful than the game screen, as
it uses 4 bits per pixel (16 colors per tile) instead of 2bpp (4 colors/tile).

sabra55:
> I think that's better; but how about reversing the order?
> 
> > A border can be much more colorful than the game screen, as
> > it uses 4 bits per pixel (16 colors per tile) instead of 2bpp (4 colors/tile).

oh yeah, that's way more natural.

ISSOtm:
Alright, thus I've gone ahead and pushed that new wording. Merging.

PR #537, opened 2024-01-15
Touch up SGB packet documentation

ISSOtm:
Turn the diagram into a SVG, and reformat/improve the doc.

### DIFF
--- a/src/SGB_Command_Packet.md
+++ b/src/SGB_Command_Packet.md
@@ -1,62 +1,51 @@
 # Command Packet Transfers
 
-Command packets (aka Register Files) are transferred from the Game Boy to
-the SNES by using P14 and P15 output lines of [the JOYPAD register](<#FF00 — P1/JOYP: Joypad>) (FF00).  These same lines are also used to select the two rows in the
-Game Boy keyboard matrix (which still works).
+Command packets are transferred from the Game Boy to the SNES by using bits 6 and 7 of of [the `JOYP` register][JOYP].
 
 ## Transferring Bits
 
-A command packet transfer must be initiated by setting both P14 and P15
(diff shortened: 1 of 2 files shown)


avivace (changes requested):
It seems like the base text size is a bit too big compared to the rest, it should be 120-140% of the base.

image

ISSOtm:
Being that this is a SVG which automatically scales to the device width, there isn't really control over that. Compare Firefox's "Responsive" view simulating a phone screen:

Screenshot showing the diagram with a wildly different proportion

Is it a big deal?

avivace:
I think shouldn't, check for example how https://gbdev.io/pandocs/Rendering.html#ppu-modes behaves, there is a limit on how "big it can become" while being free to get as small as possible preserving the layout when seen from smaller viewports..

ISSOtm:
The same width limit applies to this diagram; though, perhaps what you're saying is that the proportions of the text versus the rest of the diagram are wrong?

Note that the PPU mode diagram becomes tiny on mobile screens, though; not that I believe anything can be done about that?

Screenshot of the aforementioned diagram rendered on a mobile screen

avivace:
> The same width limit applies to this diagram; though, perhaps what you're saying is that the proportions of the text versus the rest of the diagram are wrong? That I could fix.

Yep, I would say that should fix it

ISSOtm:
Then you can try playing with the proportions by adjusting `font-size` in the SVG's embedded CSS. I find the current design fine, so I'll leave that tweaking up to you; I'd just like a chance to review that before merging?

avivace (approved):
Alright, merging this in the current state as I don't have bandwidth for the suggested style changes, I'll come back at them later

PR #540, opened 2024-01-29
Fixed misleading typo

MyselfLeo:
In the instruction set page, the binary for `ld [imm16], a` is set to `0xEC` which is an invalid code. A one and a zero had been swapped. Fixed it so that `ld [imm16], a` is `0xEA` as it should be according to [https://gbdev.io/gb-opcodes/optables/](https://gbdev.io/gb-opcodes/optables/).


ISSOtm (approved):
Quite amusing that two such typoes were found in a short time span. But yep, that's the problem with this conversion having been manual... ah well.

Thank you for fixing this!

PR #542, opened 2024-02-09
Fix ld a, [imm16]

jenningsfan:
The ld a, [imm16] opcode is listed as 1111 1100(FC), whereas it should actually be 1111 1010(FA)


ISSOtm (approved):
Another swap bites the dust... thank you! ^^'

PR #543, opened 2024-02-15
Improve documentation of timer bugs

ISSOtm:
The bulk of this research hails from @Gekkio's currently-unpublished (but very promising!) schematics (DMG-CPU-Inside was originally being used, but several finer yet crucial details were unclear; gekkio cleared those up swiftly).

The newly described behaviour makes more sense, is simpler, and most importantly more accurate to what the GB actually contains.


src/Timer_Obscure_Behaviour.md
```diff
@@ -32,99 +32,63 @@ On **CGB**:
...
-![](imgs/timer_tima_tma_detailed.svg "imgs/timer_tima_tma_detailed.svg")
-
-TMA is a latch. As soon as it is written, the output shows that value.
-That explains that when TMA is written and TIMA is being incremented,
-the value written to TMA is also written to TIMA. It doesn't affect the
-IF flag though.
+Notice how the bits themselves are connected to the multiplexer and then to the falling-edge detector; this causes a few odd behaviors:
+
+- Resetting the entire system counter (by writing to `DIV`) can reset the bit currently selected by the multiplexer, thus sending a "Timer tick" and/or "[DIV-APU event](<#DIV-APU>)" pulse early.
+- Changing which bit of the system counter is selected (by changing the "Clock select" bits of [`TAC`]) from a bit currently set to another currently reset, will send a "Timer tick" pulse.
```
avivace, line 38: "from a bit currently set to another currently reset" sounds a bit strange to me.. what is a "bit currently reset"? Could we reword?

src/Timer_Obscure_Behaviour.md
```diff
@@ -32,99 +32,63 @@ On **CGB**:
...
+  (For example: if the system counter is equal to \$3FF0 and `TAC` to \$FC, writing \$05 or \$06 to `TAC` will instantly send a "Timer tick", but \$04 or \$07 won't.)
+- On monochrome consoles, disabling the timer if the currently selected bit is set, will send a "Timer tick" once.
+  This does not happen on Color models.
```
avivace, line 41: This sounds like a repetition, or aren't monochrome/color mutually exclusive here?

src/Timer_Obscure_Behaviour.md
```diff
@@ -32,99 +32,63 @@ On **CGB**:
...
+- On Color models, a write to `TAC` that fulfills the previous bullet's conditions *and* turns the timer on (it was disabled before) may or may not send a "Timer tick".
+  This depends on a race condition, so it is device-dependent.
```
avivace, line 43: I'm not sure what this line is adding, since it's a "race condition" for which no additional details are provided (device dependent in what way?)

avivace (changes requested):

src/Timer_Obscure_Behaviour.md
ISSOtm: "A bit which is currently set" and "a bit which is currently reset"? Or "a bit equal to 1" and "a bit equal to 0"?

What matters here is the values contained in the bits that the multiplexer is considering. (Two, because the muxer is switching from one to the other.)

src/Timer_Obscure_Behaviour.md
ISSOtm: It is indeed a repetition, I tried using it for emphasis. The second sentence could be removed, I suppose.

src/Timer_Obscure_Behaviour.md
ISSOtm: What I want to convey here is that from device to device, the behaviour may be consistent one way or another, *or be inconsistent entirely*. This is what the "race condition" statement is intended to convey.

src/Timer_Obscure_Behaviour.md
avivace: I don't feel like an emphasis is needed here, the context is pretty clear and monochrome/color is a straightforward division of the models, so I suggest keeping only one.

src/Timer_Obscure_Behaviour.md
avivace: I would say inconsistent better conveys the message, then. "Since it depends on a race condition it's device-dependent" is a very confusing phrasing to me, as I don't see why a race condition should imply that is device-dependent (race condition should be used as a synonim to "indeterministic behaviour due to timing issues or other _incontrollable_ variables)

avivace: If the fact that it's device-dependent really depends on the fact that it's a race condition, then I would expand and explain why, if relevant. Otherwise, back to inconsistency as a term.

src/Timer_Obscure_Behaviour.md
avivace: Then I would use `unset`, or `currently set to 0`. A bit `currently reset` is confusing to read (as you would normally say `to reset a bit`, but less often a `reset bit` instead of a `bit set to 0`) and seems also to imply an action the preceded this situation (someone reset it), which may be true, but it's simply not useful as an information. If a bit is 0, something re-et it at some point, yes.

src/Timer_Obscure_Behaviour.md
ISSOtm: It is, internally, a race condition between some signals; but I guess technical correctness/precision here is more confusing than papering over details, yeah.

src/Timer_Obscure_Behaviour.md
ISSOtm:
```suggestion
- Changing which bit of the system counter is selected (by changing the "Clock select" bits of [`TAC`]) from a bit currently set to another that is currently unset, will send a "Timer tick" pulse.
```

PR #544, opened 2024-02-18
Add a memory map of VRAM

### LINKED ISSUE #533: Vram map - suggestion
As the title suggests, I think a table similar to the memory map but for the vram would be very helpful (especially for the cgb) as the layout of the vram isnt obvious and the information about it is very spread out thought the graphics section so a full table with the address ranges and their use (maybe at the top of the main graphics page) would be very helpful.

idk if it would be necessary, but because of the second bank not being just a copy of the use of the first due to the whole bg attribute feature a second separate table could help clarity greatly

ISSOtm:
Fixes #533.

The hand-written SVGs return! (Sorry, Antonio... :P)

The SVG was originally (see 966b0d13810e110fcf290422a10e3efc58a2c5d9) generated automatically from a Python script. At that point, it was fully self-contained (no JS required for interactivity)... at the cost of being just shy of **one megabyte**.

Considering this was ridiculously large, and *visibly taxing* on the browser, the SVG was entirely overhauled to move the interactivity from a \*\*\*\*ton of elements and a bit of CSS, to only a bunch of elements and a bit of JS. (This was never fully realised, thus never committed.)

The result was still very large, so I started making use of the `<use>` element, which allows duplicating elements without copy-paste—perfect for these massive grids! (And chances are, the browser may be able to optimise that. Idk.) This was impossible with the old CSS-only design, due to it being based on some elements having the same position **within the parent** (which `<use>` inherently precludes.)

Unfortunately, `<use>` is also very finicky with regards to `mouseover`/`mouseenter` events (especially the former), so the final implementation (fdbf3226ec587059866810a933c70de0fa091d97) may raise some eyebrows among JS devs. I tried to be clean, but I'm only an amateur in that field... :D


src/js/vram_map.js
```diff
@@ -0,0 +1,64 @@
...
+	// Check all three properties to avoid relying on the order in which the events are triggered.
+	// (The bank is always set at the same time as the address.)
+	if (mapInfo.x === null || mapInfo.y === null || mapInfo.baseAddr === null) {
+		return;
+	}
+
+	svg.getElementById("map_what").textContent = mapInfo.bank == 0 ? "Tile ID for" : "Attribute for";
+	svg.getElementById("map_x").textContent = mapInfo.x;
+	svg.getElementById("map_y").textContent = mapInfo.y;
+	svg.getElementById("map_addr").textContent = `($${toHex(mapInfo.baseAddr + mapInfo.x + mapInfo.y * 32, 4)})`;
```
Rangi42, line 56: For shorter lines, and symmetry with `tileAddr` above.
```suggestion
	const mapAddr = mapInfo.baseAddr + mapInfo.x + mapInfo.y * 32;
	svg.getElementById("map_addr").textContent = `($${toHex(mapAddr, 4)})`;
```

src/js/vram_map.js
```diff
@@ -0,0 +1,64 @@
...
+
+
+// Since the hover events are dispatched to several separate handlers, we need something to
+// centralize all of their information.
+let mapInfo = { x: null, y: null, baseAddr: null, bank: null };
+const updateMapCoords = () => {
+	// Do nothing if all of the handlers haven't fired at least once yet.
+	// Check all three properties to avoid relying on the order in which the events are triggered.
+	// (The bank is always set at the same time as the address.)
+	if (mapInfo.x === null || mapInfo.y === null || mapInfo.baseAddr === null) {
```
Rangi42, line 49: Checking the bank as well is harmless and cheap, and safe in case later it were somehow set separately.
```suggestion
	// Check all the properties to avoid relying on the order in which the events are triggered.
	if (Object.values(mapInfo).includes(null)) {
```

src/js/vram_map.js
```diff
@@ -0,0 +1,64 @@
...
+const toHex = (num, nbDigits) => num.toString(16).toUpperCase().padStart(nbDigits, '0');
+const svg = document.getElementById("vram_map");
+
+// Since the hover events are dispatched to several separate handlers, we need something to
+// centralize all of their information.
+let tileInfo = { x: null, y: null, baseAddr: null };
+const updateTileID = () => {
+	// Do nothing if all of the handlers haven't fired at least once yet.
+	// Check all three properties to avoid relying on the order in which the events are triggered.
+	if (tileInfo.x === null || tileInfo.y === null || tileInfo.baseAddr === null) {
```
Rangi42, line 24:
```suggestion
	// Check all the properties to avoid relying on the order in which the events are triggered.
	if (Object.values(tileInfo).includes(null)) {
```

src/js/vram_map.js
ISSOtm: Nice, didn't know it could be done that way!

ISSOtm:
This last commit moves the JS file from stand-alone to being contained within the SVG itself; this makes the SVG fully interactive even when viewed stand-alone instead of within Pan Docs.

src/imgs/src/vram_map.svg
```diff
@@ -0,0 +1,230 @@
...
+	<g>
+		<text x="340" y="470" class="centered descr" id="map_what"></text>
+		<text x="338" y="490" class="right descr">X =</text><text x="342" y="490" class="descr" id="map_x"></text>
+		<text x="338" y="510" class="right descr">Y =</text><text x="342" y="510" class="descr" id="map_y"></text>
+		<text x="340" y="530" class="centered descr" id="map_addr"></text>
+	</g>
+	<script>// <![CDATA[
+"use strict";
+/*
+ * Use of `on<event>` properties is normally discouraged, however they *have to* be used here.
```
Rangi42, line 170:
```suggestion
 * Use of `on<event>` properties is normally discouraged in favor of `addEventHandler`s.
 * However, they *have* to be used here.
```

src/imgs/src/vram_map.svg
```diff
@@ -0,0 +1,230 @@
...
+/*
+ * Tile blocks
+ */
+// Since the hover events are dispatched to several separate handlers, we need something to
+// centralize all of their information.
+let tileInfo = { x: null, y: null, baseAddr: null };
+const updateTileID = () => {
+	// Do nothing if all of the handlers haven't fired at least once yet.
+	// Check all properties to avoid relying on the order in which the events are triggered.
+	if (Object.values(mapInfo).includes(null)) {
```
Rangi42, line 191:
```suggestion
	if (Object.values(tileInfo).includes(null)) {
```

src/imgs/src/vram_map.svg
```diff
@@ -0,0 +1,230 @@
...
+		return;
+	}
+	const tileID = tileInfo.y * 16 + tileInfo.x + (tileInfo.baseAddr >> 4 & 0x80);
+	svg.getElementById("tile_id").textContent = `$${toHex(tileID, 2)}`;
+	const tileAddr = tileInfo.baseAddr + (tileInfo.y * 16 + tileInfo.x) * 16;
+	svg.getElementById("tile_addr").textContent = `${toHex(tileAddr, 4)}–${toHex(tileAddr + 15, 4)}`;
+};
+// These are event handlers, called into by `onmouseenter` attributes within `vram_map.svg`.
+// Have each hover event attempt to recalc the tile ID; this will cause up to 2 unnecessary recalcs,
```
Rangi42, line 200: English language convention pedantry.
```suggestion
// Have each hover event attempt to recalc the tile ID; this will cause up to two unnecessary recalcs,
```

src/imgs/src/vram_map.svg
```diff
@@ -0,0 +1,230 @@
...
+		return;
+	}
+	svg.getElementById("map_what").textContent = mapInfo.bank == 0 ? "Tile ID for" : "Attribute for";
+	svg.getElementById("map_x").textContent = mapInfo.x;
+	svg.getElementById("map_y").textContent = mapInfo.y;
+	const mapAddr = mapInfo.baseAddr + mapInfo.x + mapInfo.y * 32;
+	svg.getElementById("map_addr").textContent = `($${toHex(mapAddr, 4)})`;
+};
+// These are event handlers, called into by `onmouseenter` attributes within `vram_map.svg`.
+// Have each hover event attempt to recalc the tile ID; this will cause up to 2 unnecessary recalcs,
```
Rangi42, line 224:
```suggestion
// Have each hover event attempt to recalc the tile ID; this will cause up to two unnecessary recalcs,
```

src/imgs/src/vram_map.svg
Rangi42, line 1: An alternative suggestion for how to update the tile+map properties more tersely, using the `...` spread operator:

```js
const updateTileID = newTileInfo => {
	tileInfo = { ...tileInfo, ...newTileInfo };
	// Do nothing if all of the handlers haven't fired at least once yet.
	// Check all properties to avoid relying on the order in which the events are triggered.
	if (Object.values(tileInfo).includes(null)) {
		return;
	}
```

```js
const tile  = x        => updateTileID({ x });
const row   = y        => updateTileID({ y });
const block = baseAddr => updateTileID({ baseAddr });
```

```js
const updateMapCoords = newMapInfo => {
	mapInfo = { ...mapInfo, ...newMapInfo };
	// Do nothing if all of the handlers haven't fired at least once yet.
	// Check all properties to avoid relying on the order in which the events are triggered.
	if (Object.values(mapInfo).includes(null)) {
		return;
	}
```

```js
const mapX = x                => updateMapCoords({ x });
const mapY = y                => updateMapCoords({ y });
const tmap = (baseAddr, bank) => updateMapCoords({ baseAddr, bank });
```

Rangi42 (changes requested):

src/imgs/src/vram_map.svg
```diff
@@ -0,0 +1,230 @@
...
+const updateTileID = () => {
+	// Do nothing if all of the handlers haven't fired at least once yet.
+	// Check all properties to avoid relying on the order in which the events are triggered.
+	if (Object.values(mapInfo).includes(null)) {
+		return;
+	}
+	const tileID = tileInfo.y * 16 + tileInfo.x + (tileInfo.baseAddr >> 4 & 0x80);
+	svg.getElementById("tile_id").textContent = `$${toHex(tileID, 2)}`;
+	const tileAddr = tileInfo.baseAddr + (tileInfo.y * 16 + tileInfo.x) * 16;
+	svg.getElementById("tile_addr").textContent = `${toHex(tileAddr, 4)}–${toHex(tileAddr + 15, 4)}`;
```
Rangi42, line 197: For consistency, these hex values should have dollar sign prefixes. This also needs a bit more space between the two columns.
```suggestion
	svg.getElementById("tile_addr").textContent = `$${toHex(tileAddr, 4)}–$${toHex(tileAddr + 15, 4)}`;
```

src/imgs/src/vram_map.svg
```diff
@@ -0,0 +1,230 @@
...
+		<text x="590" y="68" class="descr">Tile block 0</text>
+		<use x="388" y="116" href="#tiles_block" class="hover" onmouseenter="block(0x8800)"/>
+		<text x="590" y="164" class="descr">Tile block 1</text>
+		<use x="388" y="212" href="#tiles_block" class="hover" onmouseenter="block(0x9000)"/>
+		<text x="590" y="260" class="descr">Tile block 2</text>
+	</g>
+	<g>
+		<text x="340" y="134" class="centered descr">Tile ID:</text>
+		<text x="340" y="154" class="centered descr" id="tile_id"></text>
+		<text x="340" y="174" class="centered descr">Addr:</text>
```
Rangi42, line 148:
```suggestion
		<text x="340" y="174" class="centered descr">Address:</text>
```

Rangi42 (changes requested):

src/imgs/src/vram_map.svg
ISSOtm: I find the current logic to be a little more obvious, and this not *much* terser... so I think I prefer the current code.

src/imgs/src/vram_map.svg
ISSOtm: Throughout Pan Docs, the `$` prefix [isn't always present](https://github.com/gbdev/pandocs/wiki/Document-Style#4-units-and-prefixes):

> To prevent clutter, don't use a prefix for hex numbers when it's clear from the context that a number is hexadecimal. For example addresses and lists of opcodes.

If consistency with the map address is desired, I'd rather modify *that* one instead.

src/imgs/src/vram_map.svg
```diff
@@ -0,0 +1,230 @@
...
+	<text x="90" y="212" class="right">0x9000</text>
+	<text x="90" y="308" class="right">0x9800</text>
+	<text x="90" y="500" class="right">0x9C00</text>
+	<text x="90" y="692" class="right">0x9FFF</text>
+	<text x="590" y="20" >0x8000</text>
+	<text x="590" y="116">0x8800</text>
+	<text x="590" y="212">0x9000</text>
+	<text x="590" y="308">0x9800</text>
+	<text x="590" y="500">0x9C00</text>
+	<text x="590" y="692">0x9FFF</text>
```
ISSOtm, line 126:
```suggestion
	<text x="90" y="20"  class="right">$8000</text>
	<text x="90" y="116" class="right">$8800</text>
	<text x="90" y="212" class="right">$9000</text>
	<text x="90" y="308" class="right">$9800</text>
	<text x="90" y="500" class="right">$9C00</text>
	<text x="90" y="692" class="right">$9FFF</text>
	<text x="590" y="20" >$8000</text>
	<text x="590" y="116">$8800</text>
	<text x="590" y="212">$9000</text>
	<text x="590" y="308">$9800</text>
	<text x="590" y="500">$9C00</text>
	<text x="590" y="692">$9FFF</text>
```

src/imgs/src/vram_map.svg
Rangi42: Okay. It's fine as-is then.

src/imgs/src/vram_map.svg
Rangi42: Alright. I like the semantics of an `update` function that actually does the info updating, but I see how `...` and the implicit `{ x }` for `{ x: x }` may not be obvious.

Rangi42 (approved):

src/Memory_Map.md
```diff
@@ -5,18 +5,92 @@ The Game Boy has a 16-bit address bus, which is used to address ROM, RAM, and I/
...
+$FF50   |         |       DMG      | Set to non-zero to disable boot ROM
+$FF51   |  $FF55  |       CGB      | [VRAM DMA](<#LCD VRAM DMA Transfers>)
+$FF68   |  $FF6B  |       CGB      | [BG / OBJ Palettes](<#LCD Color Palettes (CGB only)>)
+$FF70   |         |       CGB      | [WRAM Bank Select](<#FF70 — SVBK (CGB Mode only): WRAM bank>)
+
+## VRAM memory map
+
+VRAM is, by itself, normal RAM, and may be used as such; however, the PPU interprets it in specific ways.
+
+Bank 1 does not exist except on CGB, where it can be switched to (only in CGB Mode) using [the `VBK` register](<#FF4F — VBK (CGB Mode only): VRAM bank>).
```
avivace, line 42: Remove repetition:

```suggestion
Bank 1 does not exist except on CGB, where it can be switched to using [the `VBK` register](<#FF4F — VBK (CGB Mode only): VRAM bank>).
```

src/Memory_Map.md
```diff
@@ -5,18 +5,92 @@ The Game Boy has a 16-bit address bus, which is used to address ROM, RAM, and I/
...
+
+Each bank first contains 384 tiles, of 16 bytes each.
+These tiles are commonly thought of as grouped in three "blocks" of 128 tiles each; see [this detailed explanation](<#VRAM Tile Data>) for the relevance of the blocks.
```
avivace, line 45:
```suggestion
These tiles are commonly thought of as grouped in three "blocks" of 128 tiles each; see [this detailed explanation](<#VRAM Tile Data>) for more details.
```

avivace (changes requested):
Great job, here's some first comments

src/Memory_Map.md
ISSOtm: I don't think that's repetition, as this wording is specifically designed to single out CGB in compatibility mode; where the bank exists in hardware, but the register is disabled so the bank can't be accessed.

avivace (approved):
Looks amazing to me. Just a note: it doesn't seem to work exactly as expected as a "standalone SVG" when loading the image directly in the browser.

PR #547, opened 2024-03-20
Fix SGB output lines documentation

### LINKED ISSUE #546: SGB command packet transfers are described using wrong bits of JOYP
On the page for SGB Command Packet Transfers, they are consistently described as using bits 6 and 7 of the JOYP register, when in reality it should be bits 4 and 5. This is described correctly [on the old version of the docs](https://gbdev.gg8.se/wiki/articles/SGB_Functions#SGB_Command_Packet_Transfers):

> Command packets (aka Register Files) are transferred from the gameboy to the SNES by using P14 and P15 output lines of the JOYPAD register (FF00h), these lines are normally used to select the two rows in the gameboy keyboard matrix (which still works).

Proof that this is indeed how the feature works can be found in the code for games such as [Pokemon Gold](https://github.com/pret/pokegold/blob/master/engine/gfx/color.asm#L712), where the values $10 (a 1 bit), $20 (a 0 bit), and $30 (no data) are used for the transfer.

Zarithya:
Fixes #546.

### DIFF
--- a/src/SGB_Command_Packet.md
+++ b/src/SGB_Command_Packet.md
@@ -1,16 +1,16 @@
 # Command Packet Transfers
 
-Command packets are transferred from the Game Boy to the SNES by using bits 6 and 7 of of [the `JOYP` register][JOYP].
+Command packets are transferred from the Game Boy to the SNES by using bits 4 and 5 of of [the `JOYP` register][JOYP]. These lines are normally used to select the two rows in the Game Boy keyboard matrix (which still works).
 
 ## Transferring Bits
 
-A command packet transfer must be initiated by setting [`JOYP`][JOYP] bits 6 and 7 both to 0; this will reset and start the ICD2 packet receiving circuit.
-Data is then transferred (LSB first), setting bit 6 to 0 will indicate a `0` bit, and setting bit 7 to 0 will indicate a `1` bit.
+A command packet transfer must be initiated by setting [`JOYP`][JOYP] bits 4 and 5 both to 0; this will reset and start the ICD2 packet receiving circuit.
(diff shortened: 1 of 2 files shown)


avivace (approved):

src/SGB_Command_Packet.md
```diff
@@ -1,16 +1,16 @@
 # Command Packet Transfers
 
-Command packets are transferred from the Game Boy to the SNES by using bits 6 and 7 of of [the `JOYP` register][JOYP].
+Command packets are transferred from the Game Boy to the SNES by using bits 4 and 5 of of [the `JOYP` register][JOYP]. These lines are normally used to select the two rows in the Game Boy keyboard matrix (which still works).
```
ISSOtm, line 3:
```suggestion
Command packets are transferred from the Game Boy to the SNES by using bits 4 and 5 of of [the `JOYP` register][JOYP].
These lines are normally used to select one of the two button groups (which [still works as usual][JOYP]).
```

ISSOtm (changes requested):
I think this is importing the wording from the historical doc? I suggest rewording it a little so it's a bit clearer.

Thank you for this PR! ^^

Zarithya:
@ISSOtm Wasn't sure what wording to use there so I erred on the side of using the same wording as the old doc. I've accepted your change into the PR.

ISSOtm:
(GitHub tip: you can [dismiss a review](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/dismissing-a-pull-request-review) and then re-request it!)

ISSOtm (approved):
Thanks!

PR #548, opened 2024-04-17
Correct description of SC clock speed switch

quinnyo:
This is a pretty straightforward error correction. It didn't seem worth making an issue for it first, hopefully that's OK.

The change is to the description of the serial clock speed selector (bit 1, register SC).
The erroneous description stated that the optional higher speed was a doubling in frequency, but the actual behaviour is switching between the standard 8 kHz and the ridiculous 256 kHz (*32ing it!*).
The new description is also now in accord with the table of clock frequencies a little bit further down, under the **Internal Clock** heading.

(The serial clock frequency is *also* doubled by the system double-speed mode, which may be where some confusion came from)


avivace (approved):

ISSOtm (approved):
Thank you! No problem with not opening an issue for changes this small :)

PR #549, opened 2024-05-06
Fix some typos & inconsistencies in Licensees

IsaacMarovitz:
Companies codes I cannot confirm:
- `San-X`, old code `0x25` new code "25"
- `.Entertainment i` old code `0x3C`
- `Yonezawa/s'pal`, old code `0x96` new code "96"

### DIFF
--- a/src/The_Cartridge_Header.md
+++ b/src/The_Cartridge_Header.md
@@ -79,10 +79,10 @@ Code | Publisher
 `08` | [Capcom](https://en.wikipedia.org/wiki/Capcom)
 `13` | [EA (Electronic Arts)](https://en.wikipedia.org/wiki/Electronic_Arts)
 `18` | [Hudson Soft](https://en.wikipedia.org/wiki/Hudson_Soft)
-`19` | B-AI
-`20` | KSS
-`22` | Pow
-`24` | PCM Complete
+`19` | [B-AI](https://www.giantbomb.com/b-ai/3010-8160)
+`20` | [KSS](https://en.wikipedia.org/wiki/KSS_(company))
+`22` | [Planning Office WADA](https://www.mobygames.com/company/15869/planning-office-wada)
+`24` | [PCM Complete](https://www.mobygames.com/company/9489/pcm-complete)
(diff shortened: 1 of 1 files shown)


src/The_Cartridge_Header.md
```diff
@@ -346,8 +346,8 @@ AA    | Broderbund
 AC    | Toei Animation    
 AD    | Toho
 AF    | Namco             
-B0    | acclaim           
-B1    | ASCII or Nexsoft
+B0    | Acclaim           
+B1    | ASCII Corportation or Nexsoft
```
ISSOtm, line 350:
```suggestion
B1    | ASCII Corporation or Nexsoft
```

ISSOtm (changes requested):
Thank you! That's a good quality improvement.

Do you have a source for those? It would be nice to be able to have some references, and (Wikipedia?) links would be a plus, even.

IsaacMarovitz:
> Do you have a source for those? It would be nice to be able to have some references, and (Wikipedia?) links would be a plus, even.

As far as I can tell, this information was initially sourced from SephirothBCH's contribution on the gbdev wiki all the way back in 2000. I'm not sure where they got the original list from, and that list shares many of the issues present here.

IsaacMarovitz:
For companies that underwent renames, should those be listed somewhere here?

ISSOtm:
> For companies that underwent renames, should those be listed somewhere here?

Preferably. I'm thinking of using footnotes for that:

```md
- `xy`: [FooBar Interactive](//en.wikipedia.org/wiki/FooBar_Interactive)[^foobar]

<!-- And at the bottom of the page... -->

[^foobar]: Known as [Frobble Limited](//en.wikipedia.org/wiki/Frobble_Limited) since 2013.

```

> I can also provide links to each company's Wikipedia page (if they exist) if that's easier.

IMO, having both would be ideal.

avivace (approved):

ISSOtm (approved):
Outstanding work, thank you very much!!

PR #554, opened 2024-06-02
Addressed lints

TylerBloom:
While compiling the renderer, there were some warnings about unused imports. After I fixed that, I ran clippy in both renderer and preproc and addressed one lint that came up there.


avivace (approved):

ISSOtm (approved):
Thanks!

PR #555, opened 2024-06-15
Describe serial interrupt in serial interrupt section

### LINKED ISSUE #528: Serial transfer is re-described in the interrupts page
There's a big `**XXXXXX**` at that point, too. :P

- [ ] Port any bits of that info not present in the Serial page
- [ ] Delete the now-redundant description
- [ ] Explain *when* the interrupt is requested (= when the transfer finishes), which is what should be documented here in the first place... 😅

quinnyo:
This removes the (out of place) description of the serial port from the 'Interrupt Sources' page and replaces it with a description of when the serial interrupt is requested (which is the main thing that should be there).

This fixes #528, but I haven't moved any of the deleted text to the Serial Data Transfer page. It was almost all redundant. I don't think we're losing anything.

### DIFF
--- a/src/Interrupt_Sources.md
+++ b/src/Interrupt_Sources.md
@@ -56,42 +56,8 @@ The timer interrupt [is requested] every time that the timer overflows (that is,
 
 ## INT $58 — Serial interrupt
 
-**XXXXXX\...**
-
-Transmitting and receiving serial data is done simultaneously. The
-received data is automatically stored in SB.
-
-The serial I/O port on the Game Boy is a very simple setup and is crude
-compared to standard RS-232 (IBM-PC) or RS-485 (Macintosh) serial ports.
-There are no start or stop bits.
-
-During a transfer, a byte is shifted in at the same time that a byte is
-shifted out. The rate of the shift is determined by whether the clock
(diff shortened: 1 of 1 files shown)


ISSOtm:
Here's what I think they are intended to mean:

> The state of the last bit shifted out determines the state of the output line until another transfer takes place.

This is saying that the last bit emitted on the bus "sticks" until the next clock cycle, instead of e.g. briefly pulsing one way or another.

> The Game Boy does not support wake-on-LAN. Completion of an externally clocked serial transfer does not exit STOP mode.

I'd guess that the serial circuitry still functions in STOP mode (since it's essentially async from the rest of the system?), but does not exit STOP mode?

pinobatch:
"[Wake-on-LAN](https://en.wikipedia.org/wiki/Wake-on-LAN)" refers to a feature by which an Ethernet adapter can bring a PC out of its counterpart to STOP mode.

quinnyo:
Thanks for the feedback!
I see no reason to port either statement over.

avivace (approved):

ISSOtm (approved):

Issue #398, opened 2022-02-08
Serial transfer control information edit


jbshelton:
More testing needs to be done, but at the moment with my previous experience, bit 7 of $FF02 (SB) needs to be set in order for the Gameboy to accept an external clock and start shifting in/out data- it is **mandatory**, _not optional_ as the page implies.

quinnyo:
This is my understanding of it too.
Bit 7 of the SC register can be thought of as enabling/disabling the whole serial port. The serial port has to be enabled to respond to clock pulses. A disabled serial port can't be turned on remotely.

PR #556, opened 2024-06-30
Clarify that setting SC bit 7 is mandatory on slave to accept a Serial Data Transfer

avivace:
See https://github.com/gbdev/pandocs/issues/398

@quinnyo @ISSOtm @jbshelton

### DIFF
--- a/src/Serial_Data_Transfer_(Link_Cable).md
+++ b/src/Serial_Data_Transfer_(Link_Cable).md
@@ -3,9 +3,8 @@
 Communication between two Game Boy systems happens one byte at a time. One
 Game Boy generates a clock signal internally and thus controls when the
 exchange happens. In SPI terms, the Game Boy generating the clock is
-called the "master."  The other one uses an external clock (receiving
-it from the other Game Boy) and has no control over when the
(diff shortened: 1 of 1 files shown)


quinnyo:
I think this looks good to merge, but there are some other instances of the same issue elsewhere on the page: in the intro and the very last paragraph (under 'Delays and Synchronization'). I don't know if you want to fix them as part of this PR or separately?

avivace:
> I think this looks good to merge, but there are some other instances of the same issue elsewhere on the page: in the intro and the very last paragraph (under 'Delays and Synchronization'). I don't know if you want to fix them as part of this PR or separately?

I'd say let's handle them all in this PR? I just pushed some more changes

ISSOtm (approved):
I'm now wondering what value is obtained from attempting a transfer when the other GB(C) has not set its bit 7. Could it depend on the last transferred bit?

avivace:
> I'm now wondering what value is obtained from attempting a transfer when the other GB(C) has not set its bit 7. Could it depend on the last transferred bit?

maybe you can open a new 'research' issue regarding this?

PR #568, opened 2024-09-18
MBC6: Fix description of flash commands

msinger:
I found some mistakes in the flash commands description:
* Erase sector size is actually 128 KiB, not 8 kB.
* The address of the last byte in the command sequence for "erase sector"
  must be an address within the erased sector.
* There is really only one "exit mode" command that works for all modes.
* The data being written to the final address in programming mode doesn't
  have to be 0. It can be anything but $F0.
* $F0 doesn't have to be written to the final address. It can be written
  to any address.

I experimented with the Nintendo Power GB Memory cartridges and with Net
de Get. I wrote a documentation here:
http://iceboy.a-singer.de/doc/np_gb_memory.html

There are things not documented in Pan Docs:
* The flash in Net de Get also has a hidden 256 byte region that can be read
  and programmed.
* It has commands for protecting and "unprotecting" sector 0.

I don't know, if you want to add them. Maybe those things are not used by
Net de Get. I can't tell, because all my Net de Get cartridges came with an
empty flash.

### DIFF
--- a/src/MBC6.md
+++ b/src/MBC6.md
@@ -98,26 +98,24 @@ either 4 or 6 and Y to 5 or 7, depending on the bank region:
 
 ```
 ------------- ------------- ------------- ------------- ------------- ------------- ------------------------------------------------
-2:Y555=$AA    1:XAAA=$55    2:Y555=$80    2:Y555=$AA    1:XAAA=$55    ?:X000=$30    Erase sector\* (set 8 kB region to $FFs)
-2:Y555=$AA    1:XAAA=$55    2:Y555=$80    2:Y555=$AA    1:XAAA=$55    ?:Y555=$10    Erase chip\* (set entire flash to $FFs)
-2:Y555=$AA    1:XAAA=$55    2:Y555=$90                                                 ID mode (reads out JEDEC ID (C2,81) at $X000)
-2:Y555=$AA    1:XAAA=$55    2:Y555=$A0                                                 Program mode\*
-2:Y555=$AA    1:XAAA=$55    2:Y555=$F0                                                 Exit ID/erase chip mode
-2:Y555=$AA    1:XAAA=$55    ?:X000=$F0                                                 Exit erase sector mode
-?:????=$F0                                                                               Exit program mode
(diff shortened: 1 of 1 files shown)


src/MBC6.md
```diff
@@ -126,6 +124,16 @@ properly and re-write (without erasing) the 128 byte block if some bits
 didn't get set to 0 properly. After writing all blocks in a sector
 Flash Write Enable should be set to 0.
 
+The last byte of the erase sector command needs to be written to an
+address that lies within the sector that you want to erase. There are
+eight sectors, 128 KiB each. For example, to erase sector 2, the last
```
avivace, line 129:
```suggestion
eight sectors, 128 KiB each. E.g., to erase sector 2, the last
```

avivace (changes requested):

avivace:
Thanks @msinger ! I'd say we may also want to add http://iceboy.a-singer.de/doc/np_gb_memory.html to the references, somewhere?

ISSOtm:
> I found some mistakes in the flash commands description:

Thank you for taking the time to correct them!

> I don't know, if you want to add them. Maybe those things are not used by Net de Get. I can't tell, because all my Net de Get cartridges came with an empty flash.

This is fine to add regardless, as we aim to document as much of the console as possible so that homebrew devs can make use of them. Thank you!

> Should I add the link to the _External links_ section at the bottom?

Yes please!

> I also noticed that there are kB units in the first paragraph. CONTRIBUTING.md says that binary prefixes should be used. I wasn't sure if it is okay to change them to KiB with this PR, even if I don't touch anything else in that region. If you like I make it unified.

If you feel like improving the document's format, then yes please. *Ideally* make it a separate commit, but it's no big deal if we end up with a few “drive-by” fixes.

msinger:
I added the flash commands and the reference. My spellchecker tells me that "unprotect" is not a real word. I'm not a native speaker, so I don't know if this is true. The flash datasheets use this word for this operation. I googled it and it is also used for "unprotecting" Excel sheets and stuff. So it seems to be commonly used. Let me know if you want to use another word instead.

ISSOtm:
Neither Antonio nor I are native speakers either, and regardless, we accept neologisms here. This is not a Merriam-Webster office ;)

ISSOtm (approved):
Thank you for taking the time to add this!

avivace (approved):

ISSOtm:
There we to! Thank you again ^^

PR #569, opened 2024-09-25
Add a note to the HDMA timings about how many tiles you can maximal transfer during VBlank.

daid:
Add a note to the HDMA timings about how many tiles you can maximal transfer during VBlank. This is one of those numbers you can figure out by going from M-Cycles to dots and then checking the amount of dots in VBlank, but it's a common enough thing you would want to know that I think it's good to have it here.


ISSOtm (approved):
Yes, this is good information to point out explicitly. “Saved you a ~~click~~ calculation.”

Thank you, Daid!

daid:
I figured it's a easy leap from this to "oh, it's easy to transfer 128 tiles in a single vblank".

Note that in theory you could transfer a bit more, if you use the hblank of the last line and the OAM scan of the first line as well you can transfer about 10 tiles more. But getting that timing right would be quite complicated. And with this, I figured most people would go "let's just transfer 128 tiles per vblank"

PR #577, opened 2024-11-01
Add behavior of the HDMA HBlank mode in halt mode

alloncm:


ISSOtm (approved):
I have seen this confirmed several times, so I'm approving this despite the lack of a reference. (Better have the information *somewhere*!)

Thank you!

alloncm:
I can add a reference to a test ROM I wrote. 
The issue with it is that it doesn't verify the transfer continues once the CPU is unhalted (that's why I didn't add it in this PR). 
I verified this behavior manually with SameBoy though.

https://github.com/alloncm/MagenTests?tab=readme-ov-file#vram-dma-hblank-mode

avivace (approved):
OK for me in this state but I'd add the link to the test rom

src/CGB_Registers.md
```diff
@@ -79,7 +79,10 @@ the program execution continues during the "spaces" between each data
 block. Note that the program should not change the Destination VRAM bank
 (FF4F), or the Source ROM/RAM bank (in case data is transferred from
 bankable memory) until the transfer has completed! (The transfer should
-be paused as described below while the banks are switched)
+be paused as described below while the banks are switched).
+
+Upon halting the CPU (using the [halt instruction](<#Using the HALT Instruction>)),
+the transfer will also be halted and will be resumed only when the CPU resumes execution.
```
avivace, line 85:
```suggestion
the transfer will also be halted and will be resumed only when the CPU resumes execution ([test rom](https://github.com/alloncm/MagenTests?tab=readme-ov-file#vram-dma-hblank-mode) exhibiting this behaviour)
```

avivace (changes requested):
e.g. like this

avivace (approved):

PR #582, opened 2025-01-15
Update Wording of SC Check in Serial Data Transfer Page

iMackshun:
-Updated the wording on the Serial Transfer Page to not mention checking for $01, as this is incorrect. Removed that line and added a sentence to describe remind the reader to only check Bit 7. This may be implied after removing the text in parentheses, but may not hurt to stress the importance of only checking Bit 7.
-While testing in Emulicious, I was checking comparing SC to $01 and my code wasn't working because SC was actually equal to $7F. Bit 7 is indeed 0, but the other bits changed, thus making checking for $01 incorrect, and maybe situational. I was under the assumption that the other bits wouldn't change after I set SC to $81.


ISSOtm (approved):
Thank you!

PR #583, opened 2025-01-19
Clarify the STAT register PPU mode when the PPU is disabled

alloncm:
I verified this on SameBoy with some manual tests.

Let me know if you think this requires a proper test rom.

### DIFF
--- a/src/STAT.md
+++ b/src/STAT.md
@@ -29,7 +29,7 @@ is set, and (if enabled) a STAT interrupt is requested.
 - **Mode 1 int select** (*Read/Write*): If set, selects the Mode 1 condition for [the STAT interrupt](<#INT $48 — STAT interrupt>).
 - **Mode 0 int select** (*Read/Write*): If set, selects the Mode 0 condition for [the STAT interrupt](<#INT $48 — STAT interrupt>).
 - **LYC == LY** (*Read-only*): Set when [LY](<#FF44 — LY: LCD Y coordinate \[read-only\]>) contains the same value as [LYC](<#FF45 — LYC: LY compare>); it is constantly updated.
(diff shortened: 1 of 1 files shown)


avivace (approved):

src/STAT.md
```diff
@@ -29,7 +29,7 @@ is set, and (if enabled) a STAT interrupt is requested.
 - **Mode 1 int select** (*Read/Write*): If set, selects the Mode 1 condition for [the STAT interrupt](<#INT $48 — STAT interrupt>).
 - **Mode 0 int select** (*Read/Write*): If set, selects the Mode 0 condition for [the STAT interrupt](<#INT $48 — STAT interrupt>).
 - **LYC == LY** (*Read-only*): Set when [LY](<#FF44 — LY: LCD Y coordinate \[read-only\]>) contains the same value as [LYC](<#FF45 — LYC: LY compare>); it is constantly updated.
-- **PPU mode** (*Read-only*): Indicates [the PPU's current status](<#PPU modes>).
+- **PPU mode** (*Read-only*): Indicates [the PPU's current status](<#PPU modes>), when the [PPU is disabled](<#LCDC.7 — LCD enable>) read as 0 (HBlank).
```
ISSOtm, line 32: **bikeshed**: I think this wording is a little clearer?
```suggestion
- **PPU mode** (*Read-only*): Indicates [the PPU's current status](<#PPU modes>). Reports 0 instead when the [PPU is disabled](<#LCDC.7 — LCD enable>).
```

ISSOtm (changes requested):
Thank you for adding this! This is indeed a good remark, as I see it tripping people up. I believe at least one test ROM matches this, though @LIJI32 would have a better idea than I of which one.

avivace (approved):

PR #584, opened 2025-02-24
scrolling: Add note about disabling/re-enabling the window mid-frame

bpeel:
Hi,

In a game I’m making I wanted to have the window be displayed at the top and bottom of the screen while showing the background layer for the middle part so that I could scroll it freely. I wasted quite a bit of time thinking it wasn’t working because I made the wrong assumption that it would use the tile map data from the bottom of the screen after re-enabling the window. I was wondering if it might be worth explicitly mentioning the actual behaviour in the Pandocs in case someone else runs into the same problem.


src/Scrolling.md
```diff
@@ -62,3 +62,5 @@ For the window to be displayed on a scanline, the following conditions must be m
 
 If the WY condition has already been triggered and at the start of a row the window enable bit was set,
 then resetting that bit before the WX condition gets triggered on that row yields a nice window glitch pixel where the window would have been activated.
+
+If the window is disabled for part of the rendering and then re-enabled later on in the same frame then the tile map data used to render the second part will be directly following the rows in memory for the first part with no gaps. This happens regardless of whether the window is disabled by setting its position offscreen or by clearing the window enable bit. For example, if you want to draw one row of tiles using the window at the top of screen and then another row of tiles at the bottom of the screen by disabling the window for the middle portion, then the tile map data used for rendering will be the first and second rows, not the first and 18th.
```
avivace, line 66: "disabled for part of the rendering" could be clarified a bit better?

alloncm:
HI, I'm not sure if I understand correctly but I think this behavior is already explained [here](https://github.com/gbdev/pandocs/blob/master/src/Tile_Maps.md?plain=1#L107), by the window internal line counter.

Perhaps this information should be more visible / moved to other page.

bpeel:
> Hi, I'm not sure if I understand correctly but I think this behavior is already explained [here](https://github.com/gbdev/pandocs/blob/master/src/Tile_Maps.md?plain=1#L107), by the window internal line counter.

Oh! I must have read that at some point but didn’t remember it and then didn’t think to look at that page when trying to figure out why disabling the window didn’t seem to be working. I think it explains the workings better and more succinctly than my attempt.

> Perhaps this information should be more visible / moved to other page.

Yes, I think it would be nice to move that information to the part about “mid-frame behaviour” because it is more directly related to that.

src/Scrolling.md
ISSOtm: Here is a more drastic change, then:

```suggestion
The way the Window selects which line of its tilemap to render may be surprising: the Y position is selected by an internal counter, which is reset to 0 during VBlank and **only** incremented when the Window starts being rendered on a given scanline.
In particular, this means that hiding the Window mid-frame in any way (via either `WX` or `LCDC`, usually to display a status bar at the top *and* bottom of the screen) will also inhibit incrementing that Y-position counter.
```

This can probably be accompanied by an image showing some data in the tilemap, and how the window references it. (The image would have two rectangular parts, one showing the tilemap and the other the screen; arrows showing where WX is set to e.g. $40 and $A7 mid-frame; and arrows showing which parts of the tilemap are displayed where on-screen.)

I don't have the energy to make such an image from scratch, though. Would someone else be kind enough to at least draft it? I will gladly provide editing help.

ISSOtm (changes requested):
Thank you for this addition!

src/Scrolling.md
nummacway: Avivace's original suggestion focusses on the row of _tiles_, not the scanline of _pixels_. ISSO's suggestion focusses only on pixels. The two are not equivalent.
Say I draw 4 rows of the Window at the top, then disable Window and Enable it again for aline whose modulo is not 4, which tiles will be fetched and which of their lines of pixels are drawn?

Another interesting question would be the change of the Window tilemap mid-frame.

ISSOtm: The behaviour is per-pixel, and the fetcher internally discards the low 3 bits when fetching tiles.

The tilemap bit is not cached, so it takes effect on the next tilemap fetch.

nummacway: Just to make sure I understand it correctly:

Say my Window looks like this if drawn completely (this is 160x144; every bar is 2px tall):
image

If I draw it on a white background for 4 lines at the top and 4 at the bottom, will it fetch the normal tiles at the bottom?
image

bpeel: I can confirm that that is the resulting behaviour on BGB with [this little example](https://github.com/bpeel/window-example/blob/main/window-example.asm).

nummacway: Since I cannot `make` it on Windows, can you upload the `.gb` so I can run this on real hardware? I do trust BGB, but this seems to be some very obscure topic we're discussing here.

avivace: I like @ISSOtm rewording!

src/Scrolling.md
```diff
@@ -62,3 +62,6 @@ For the window to be displayed on a scanline, the following conditions must be m
 
 If the WY condition has already been triggered and at the start of a row the window enable bit was set,
 then resetting that bit before the WX condition gets triggered on that row yields a nice window glitch pixel where the window would have been activated.
+
+The way the Window selects which line of its tilemap to render may be surprising: the Y position is selected by an internal counter, which is reset to 0 during VBlank and **only** incremented when the Window starts being rendered on a given scanline.
+In particular, this means that hiding the Window mid-frame in any way (via either `WX` or `LCDC`, usually to display a status bar at the top *and* bottom of the screen) will also inhibit incrementing that Y-position counter.
```
alloncm, line 67: I think this should also reference [this tip in the Tile Maps page](https://github.com/gbdev/pandocs/blob/master/src/Tile_Maps.md?plain=1#L107) or even move it / merge them.

nummacway:
Shouldn't this explain the Window's mid-_scanline_ behaviour (changing LCDC.5, WX and WY mid-scanline) as well? Mid-scanline behaviour _is_ explained for _scrolling_.

ISSOtm:
This would indeed be a good idea, but to get *some* improvements in, I'd rather do it in a separate PR.

src/Scrolling.md
avivace: Agree, this is now a "repetition", we could actually merge it there and link it from here or just keep it here and link it from there.. maybe let's open an issue about this so we don't block this one?

avivace (approved):

avivace:
> Shouldn't this explain the Window's mid-_scanline_ behaviour (changing LCDC.5, WX and WY mid-scanline) as well? Mid-scanline behaviour _is_ explained for _scrolling_.

Yep, agree! Let's merge this in the meantime. I opened https://github.com/gbdev/pandocs/issues/591

src/Scrolling.md
avivace: I opened https://github.com/gbdev/pandocs/issues/592

PR #585, opened 2025-02-25
Fix Typo NR14->44

elp2:
This must mean the noise register's length, not Pulse 1's.


ISSOtm (approved):
Thank you for taking the time to correct this!

PR #587, opened 2025-02-27
Tablify and fix MBC3 inconsistencies.

elp2:
Does it make sense to list A000-BFFF as a RW Register and Memory range? I think it makes sense to just remove A000-BFFF from the registers section since the description just copies what is above.

### DIFF
--- a/src/MBC3.md
+++ b/src/MBC3.md
@@ -18,7 +18,7 @@ Contains the first 16 KiB of the ROM.
 Same as for MBC1, except that accessing banks $20, $40, and $60 is
 supported now.
 
-### A000-BFFF - RAM Bank 00-03, if any (Read/Write)
+### A000-BFFF - RAM Bank 00-07 or RTC register (Read/Write)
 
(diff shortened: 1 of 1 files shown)


ISSOtm:
Paging @Rangi42, I think the Pokémon hackers can confirm whether MBC3 (not MBC30!) can have 8 SRAM banks?

ISSOtm (changes requested):
The maximum number of RAM banks needs to be explained somehow... but that also involves talking about MBC3 versus MBC30. Is that variant already mentioned somewhere?

elp2:
How would you like me to mention? I would assume that since we can address the 8 banks of RAM as mentioned 00-07 this is already covered that that max is 8 for the MBC3x series? Can add a line to that extent if you want.

avivace (approved):

avivace:
Alright, I'm going to start merging what we have. Feel free to bring this up again in an issue @ISSOtm .

@elp2 thanks for the contribution!

PR #594, opened 2025-03-22
Add description for out of bounds MBC ram access

alloncm:
This is based on:
- Source code of the SameBoy emulator
- Disassembling Pokemon Pinball which depends on this behavior
- Mbc1 schematics - https://www.devrs.com/gb/files/mbc1.gif

Since there are many types of MBC's I added that this is for most of the MBC's

### DIFF
--- a/src/MBCs.md
+++ b/src/MBCs.md
@@ -23,7 +23,7 @@ Double Speed Mode.
 
 ## MBC Unmapped RAM Bank Access
 
-On most of the MBC's in case an unmapped RAM bank is selected (which would be translate to an out of bounds RAM address by the MBC controller), 
+In most MBCs, if an unmapped RAM bank is selected (which would be translate to an out of bounds RAM address by the MBC controller), 
 the MBC will simply wrap around the internal ram address and would access a valid RAM address.
 
 The MBC internal address being accessed can be calculated using this formula: `((address - external_ram_start_address) + (active_ram_bank * ram_bank_size)) % max_external_ram_size`.
--- a/src/MBCs.md
+++ b/src/MBCs.md
@@ -25,4 +25,4 @@ Double Speed Mode.
 
 On most of the MBC's in case an out of bounds address is accessed (selecting unmapped RAM bank),
 only the mapped bits from the address will be used.
-The actual address being accessed can be calculated using this formula: `address & (max_ram_size - 1)`.
+The actual address being accessed can be calculated using this formula: `address % max_ram_size`.
--- a/src/MBCs.md
+++ b/src/MBCs.md
@@ -20,3 +20,9 @@ so, it might be nevertheless possible to use Double Speed during periods
 which use only code and data which is located in internal RAM. Despite the 
 above, a self-made MBC1-EPROM card appears to work stable and fine even in 
 Double Speed Mode.
+
+## MBC Out Of Bounds RAM Access
+
+On most of the MBC's in case an out of bounds address is accessed (selecting unmapped RAM bank),
+only the mapped bits from the address will be used.
+The actual address being accessed can be calculated using this formula: `address & (max_ram_size - 1)`.
--- a/src/MBCs.md
+++ b/src/MBCs.md
@@ -21,8 +21,9 @@ which use only code and data which is located in internal RAM. Despite the
 above, a self-made MBC1-EPROM card appears to work stable and fine even in 
 Double Speed Mode.
 
-## MBC Out Of Bounds RAM Access
+## MBC Unmapped RAM Bank Access
 
-On most of the MBC's in case an out of bounds address is accessed (selecting unmapped RAM bank),
-only the mapped bits from the address will be used.
-The actual address being accessed can be calculated using this formula: `address % max_ram_size`.
+On most of the MBC's in case an unmapped RAM bank is selected (which would be translate to an out of bounds RAM address by the MBC controller), 
+the MBC controller will simply wrap around the internal ram address and would access valid RAM address.
+
+The MBC internal address being accessed can be calculated using this formula: `((address - external_ram_start_address) * ram_bank_size) % max_external_ram_size`.
--- a/src/MBCs.md
+++ b/src/MBCs.md
@@ -24,6 +24,6 @@ Double Speed Mode.
 ## MBC Unmapped RAM Bank Access
 
 On most of the MBC's in case an unmapped RAM bank is selected (which would be translate to an out of bounds RAM address by the MBC controller), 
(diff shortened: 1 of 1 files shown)


nummacway:
Use "modulo max_ram_size" instead of "and (max_ram_size - 1)"?
Why does Pinball depend on it? It doesn't even have banked RAM.

alloncm:
Perhaps "depends" is not the best way to phrase it, what I actually saw is that Pokemon Pinball sets the ram bank to a non zero value and since it does not have banked ram, all accesses are clipped (according to the formula).
Failing to emulate this exact behavior causes the game to freeze at some point.

nummacway:
What point? I want to try with a 32 KiB cart.

alloncm:
It depends on the way you are handling this case, originally I just ignored those out of bounds writes, (and returned 0 on reads) with this way it froze after trying to reload a save from sram (boot up -> save -> restart -> load save -> freeze).
I dont remember exactly if the load save is also done with the same non zero bank (I think it was bank 4) if it is it supposed to work with a 32kb cart, if not I think it will freeze.

src/MBCs.md
```diff
@@ -20,3 +20,9 @@ so, it might be nevertheless possible to use Double Speed during periods
 which use only code and data which is located in internal RAM. Despite the 
 above, a self-made MBC1-EPROM card appears to work stable and fine even in 
 Double Speed Mode.
+
+## MBC Out Of Bounds RAM Access
+
+On most of the MBC's in case an out of bounds address is accessed (selecting unmapped RAM bank),
```
ISSOtm, line 26: Is there any MBC on which this doesn't happen?

ISSOtm (changes requested):

src/MBCs.md
pinobatch: In MBC3, banks $08-$0C select the RTC, and in MBC7, bank $40 selects the accelerometer and EEPROM.

nummacway: What happens if you request banks $04-$07 or $0D-$0F?
I actually have everything you need to check this, I'm just lazy. :sloth:

CasualPokePlayer: $04-$07 and $0D-$0F just map to "nothing", read returning open bus and writes doing nothing. (also this is assuming MBC3, with MBC30 $04-$07 are just more RAM banks as expected)

nummacway: I decided to not be lazy and tested this. $04-$07 and $0D-$0F return $FF on every address and are read-only. The high RAM bank nibble is ignored. So $4000 is effectively a 4-bit register.

CasualPokePlayer: The 0xFF is actually just an open bus value: 0xFF would just be a usual open bus value if you're reading from somewhere else other than the cartridge bus (e.g. WRAM within the GBC, or just HRAM in all units). It can be understood as just being the same as reading SRAM while SRAM is disabled via the 0000-1FFF register.

ISSOtm: $FF is a common SRAM open-bus value especially with some flashcarts, so this makes sense.

alloncm: I believe others know better than me (as appears in this thread), since there are many types and shapes of MBC's I thought it could be a good idea to mention this is probably valid for most of them and not all of them.

src/MBCs.md
```diff
@@ -20,3 +20,9 @@ so, it might be nevertheless possible to use Double Speed during periods
...
+only the mapped bits from the address will be used.
```
ISSOtm, line 27: Wait--is this talking about address, or bank? I know they are combined to ultimately make a single address passed to the chip; but the entire document is written from the perspective of the GB CPU, since that's the most consistent and generally-useful one, and from that perspective, the bank and address are separate.

ISSOtm (changes requested):

src/MBCs.md
alloncm: Agree! Just rephrased it to be more CPU oriented, let me know what you think

src/MBCs.md
```diff
@@ -20,3 +20,10 @@ so, it might be nevertheless possible to use Double Speed during periods
 which use only code and data which is located in internal RAM. Despite the 
 above, a self-made MBC1-EPROM card appears to work stable and fine even in 
 Double Speed Mode.
+
+## MBC Unmapped RAM Bank Access
+
+On most of the MBC's in case an unmapped RAM bank is selected (which would be translate to an out of bounds RAM address by the MBC controller), 
+the MBC controller will simply wrap around the internal ram address and would access valid RAM address.
+
+The MBC internal address being accessed can be calculated using this formula: `((address - external_ram_start_address) * ram_bank_size) % max_external_ram_size`.
```
ISSOtm, line 29: I'd have expected `(active_ram_bank % nb_ram_banks) * 0x1000 + (address - 0xA000)`. Am I missing something?

ISSOtm (changes requested):

src/MBCs.md
alloncm: You are correct, I forgot to add the active ram bank to the formula.

I wrote the formula differently but they are supposed to be equivalent (unless I made a mistake again), let me know which way you prefer it.

avivace (approved):

src/MBCs.md
```diff
@@ -20,3 +20,10 @@ so, it might be nevertheless possible to use Double Speed during periods
 which use only code and data which is located in internal RAM. Despite the 
 above, a self-made MBC1-EPROM card appears to work stable and fine even in 
 Double Speed Mode.
+
+## MBC Unmapped RAM Bank Access
+
+On most of the MBC's in case an unmapped RAM bank is selected (which would be translate to an out of bounds RAM address by the MBC controller),
```
avivace, line 26:
```suggestion
In most MBCs, if an unmapped RAM bank is selected (which would be translate to an out of bounds RAM address by the MBC controller), 
```

src/MBCs.md
```diff
@@ -20,3 +20,10 @@ so, it might be nevertheless possible to use Double Speed during periods
 which use only code and data which is located in internal RAM. Despite the 
 above, a self-made MBC1-EPROM card appears to work stable and fine even in 
 Double Speed Mode.
+
+## MBC Unmapped RAM Bank Access
+
+On most of the MBC's in case an unmapped RAM bank is selected (which would be translate to an out of bounds RAM address by the MBC controller), 
+the MBC controller will simply wrap around the internal ram address and would access valid RAM address.
```
avivace, line 27: the repetition of "MBC controller" doesn't sound good here

avivace (changes requested):

avivace (approved):

PR #595, opened 2025-03-27
Make the priority flag description clearer by explicitly referring to "color IDs" instead of "colors"

RaoulLuque:
When reading the description of byte 3 in the OAM section and more precisely, the priority flag, I was confused whether the sentence
"Priority: 0 = No, 1 = BG and Window colors 1–3 are drawn over this OBJ"
was referring to colors, as in actual colors values (0: White, ...) or whether it was referring to color IDs (which might be mapped by the palette to other color values then their ID might suggest). I suppose it refers to color IDs, which is why I would make that more explicit.

If that is wrong I would suggest changing it to make it clear that color IDs are not meant and instead color values. I was not able to find a conclusive source on this.

Note that the link that is used for the [color IDs](https://gbdev.io/pandocs/Tile_Data.html) is the same as used in the [palettes section](https://gbdev.io/pandocs/Palettes.html#ff47--bgp-non-cgb-mode-only-bg-palette-data).

### DIFF
--- a/src/OAM.md
+++ b/src/OAM.md
@@ -53,7 +53,7 @@ tile is "NN & \$FE", and the bottom 8×8 tile is "NN | \$01".
   "Attributes"  7:"Priority" 6:"Y flip" 5:"X flip" 4:"DMG palette" 3:"Bank" 2-0:"CGB palette";
 }}
 
-- **Priority**: `0` = No, `1` = BG and Window [color IDs](./Tile_Data.md) 1–3 are drawn over this OBJ
+- **Priority**: `0` = No, `1` = BG and Window [color indices](<#Data format>) 1–3 are drawn over this OBJ
 - **Y flip**: `0` = Normal, `1` = Entire OBJ is vertically mirrored
 - **X flip**: `0` = Normal, `1` = Entire OBJ is horizontally mirrored
 - **DMG palette** *\[Non CGB Mode only\]*: `0` = OBP0, `1` = OBP1
--- a/src/OAM.md
+++ b/src/OAM.md
@@ -53,7 +53,7 @@ tile is "NN & \$FE", and the bottom 8×8 tile is "NN | \$01".
   "Attributes"  7:"Priority" 6:"Y flip" 5:"X flip" 4:"DMG palette" 3:"Bank" 2-0:"CGB palette";
 }}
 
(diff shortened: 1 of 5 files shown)


src/OAM.md
```diff
@@ -53,7 +53,7 @@ tile is "NN & \$FE", and the bottom 8×8 tile is "NN | \$01".
   "Attributes"  7:"Priority" 6:"Y flip" 5:"X flip" 4:"DMG palette" 3:"Bank" 2-0:"CGB palette";
 }}
 
-- **Priority**: `0` = No, `1` = BG and Window colors 1–3 are drawn over this OBJ
+- **Priority**: `0` = No, `1` = BG and Window [color IDs](./Tile_Data.md) 1–3 are drawn over this OBJ
```
ISSOtm, line 56:
```suggestion
- **Priority**: `0` = No, `1` = BG and Window [color indices](./Tile_Data.md) 1–3 are drawn over this OBJ
```
[The priority description](https://gbdev.io/pandocs/Tile_Maps.html#bg-to-obj-priority-in-cgb-mode) calls them “color index”es, which I think is a little clearer. (Is the term “color ID” used anywhere else? We should probably change it there too, if so.)

ISSOtm, line 56:
```suggestion
- **Priority**: `0` = No, `1` = BG and Window [color IDs](<#Data format>) 1–3 are drawn over this OBJ
```
We use some custom syntax to more easily link to the appropriate section within other pages.

ISSOtm (changes requested):
Thank you for the contribution! It does indeed refer to colour IDs, your guess was correct!

Would you mind checking the corresponding flag in the BG attributes description (Tile_Maps.md)? I believe it's more or less a copy-paste.

src/OAM.md
RaoulLuque: Ah, I see, good to know :) I actually copied the link syntax from the one used in the [BG palette data description](https://github.com/gbdev/pandocs/blob/215cae000901f987576400bdbca77ea35528150c/src/Palettes.md?plain=1#L8), so should we also update the link there ?

ISSOtm: Yes please!

RaoulLuque:
Good catch, I'll search for more occurrences and change all of them accordingly 👍

avivace (approved):

PR #602, opened 2025-06-09
Fix MBC1M addressing diagram

TheThief:


TheThief:
The current diagram directly contradicts the information in the text above it - the text says that in mode 1 banks $00, $10, $20 and $30 can be selected, changing the "zero bank" for games, but the diagram shows it selecting any arbitrary bank, which is something MBC1 can't do even in mode 1. Additionally, MBC1M is just MBC1 but with a single bit disconnected, so the table should be almost the same as the one further up the same file for $0000-$3FFF in a regular MBC1, but with a single bit removed. The table further up has zero in that spot too.

It seems to be an error introduced in the conversion from the old ascii art to the new tables, because the old diagram also said "always 0" for the middle bits: https://github.com/gbdev/pandocs/commit/a3f9c0896d52c8246f13077ca65fcd8a40ba0f79#diff-7e4b4fae3f0cddadcd95ba06293c452188c2dbf8ac964ce2eef35c081998eacdL217-L223

Mooneye's MBC1M test can be cited as confirmation that only the 2 bit bank register is used for the $0000-$3FFF region: https://github.com/Gekkio/mooneye-test-suite/blob/main/emulator-only/mbc1/multicart_rom_8Mb.s#L60-L68

ISSOtm (approved):
Sounds good to m,e then. Thank you for taking the time to fix this!

PR #622, opened 2025-09-10
Attempt to fix broken WRAM bank hyperlink

### LINKED ISSUE #621: Non-hyperlinked register name
WRAM bank non-hyperlinked on https://gbdev.io/pandocs/Hardware_Reg_List.html

Info about that register here:

https://gbdev.io/pandocs/CGB_Registers.html?highlight=svbk#ff70--svbkwbk-cgb-mode-only-wram-bank

ReiquelApplegate:
Fixes #621 

Remember, I'm still not setup to generate the HTML and check if this fix succeeded myself.


ISSOtm:
Hmm, there's also a commit that switches from “single-speed” to “normal-speed”, which seems unintended. I can remove that commit myself if you'd like me to?

ISSOtm:
Confirmed, but strange that it still showed as a change in the UI. Ah well.

ISSOtm (approved):
Checked, and this does fix the issue. Thank you!

avivace (approved):

PR #624, opened 2025-09-14
Specified the status of the VRAM DMA register once a transfer is done.

Phidias618:
I specified what the content of the VRAM DMA registers are once a transfer is done.

"The source and destination address regisers are both incremented by $10 bytes for each block of $10 bytes transfered after the transfer is done.
FF55 is always equal to $FF once the transfer ends."

It is useful information when multiple VRAM DMA transfer need to be performed in a row.
For example if one wants to fill VRAM with only 0 using one array of 16 null bytes as the source, as the destination address registers don't need to be updated after each transfer, but the source address registers always need to be reset back to the address of the begining of the source array after each transfer
Here is some assembly code that takes advantage of this in order to clear one bank of VRAM
_zero_x16 is an array of 16 null bytes, aligned on a 16 bytes boundary
[clear_vram.s](https://github.com/user-attachments/files/22323433/clear_vram.txt)

### DIFF
--- a/src/CGB_Registers.md
+++ b/src/CGB_Registers.md
@@ -127,7 +127,7 @@ This allows for a transfer of 2280 bytes during VBlank, which is up to 142.5 til
 #### Status of the registers after the end of the transfer
 
 The source and destination address regisers are both incremented by $10 bytes for each block of $10 bytes transfered after the transfer is done.
-FF55 is set to $FF once the transfer ends.
+FF55 is always equal to $FF once the transfer ends
 
 ### VRAM Banks
 
--- a/src/CGB_Registers.md
+++ b/src/CGB_Registers.md
@@ -124,6 +124,11 @@ Speed Mode).
 
 This allows for a transfer of 2280 bytes during VBlank, which is up to 142.5 tiles.
 
+#### Status of the registers after the end of the transfer
+
+The source and destination address are both incremented by $10 bytes for each block of $10 bytes transfered after the transfer is done.
+FF55 is set to $FF once the transfer ends.
+
 ### VRAM Banks
 
 The CGB has twice the VRAM of the DMG, but it is banked and either bank
--- a/src/CGB_Registers.md
+++ b/src/CGB_Registers.md
@@ -39,14 +39,33 @@ tested on Echo RAM, OAM, FEXX, IO and HRAM\]. Trying to specify a source
 address in VRAM will cause garbage to be copied.
 
 The four lower bits of this address will be ignored and treated as 0.
-The address specified by those registers is cached internally and incremented by $10 for each block of $10 bytes successfully transferred. The cached address persists until the registers are written again.
+After a transfer the address contained in this register pair is 
+by $10 for each block of $10 bytes transfered, these 
+registers being write-only, this can only be observed by doing another 
+transfer without updating these registers.
+
 
 #### FF53–FF54 — HDMA3, HDMA4 (CGB Mode only): VRAM DMA destination (high, low) \[write-only\]
 
 These two registers specify the address within 8000-9FF0 to which the
 data will be copied. Only bits 12-4 are respected; others are ignored.
 The four lower bits of this address will be ignored and treated as 0.
-The address specified by those registers is cached internally and incremented by $10 for each block of $10 bytes successfully transferred. The cached address persists until the registers are written again.
+
+#### State of the VRAM DMA source/destination registers after a transfer
+
+After a transfer, the source/destination registers are incremented by $10 
+for each block of $10 bytes transfered.
+Despite both the VRAM DMA source/destination registers being write-only,
+knowing their state after a transfer can turn useful when performing
+multiple transfer in a row.
+For instance, a transfer of one large block is mostly equivalent to
+multiple transfers of smaller blocks, without needing to update the source
+nor the destination registers between each of the smaller transfers.
+Another use case would be to fill VRAM with the same $10 bytes block repeated 
+all over, as only the source address register would need be updated after each
+transfer, the destination register being automatically incremented by
+the block size after each transfer.
+
 
 #### FF55 — HDMA5 (CGB Mode only): VRAM DMA length/mode/start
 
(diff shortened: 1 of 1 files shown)


alloncm:
Hi, the state of this register after a transfer complete is already described [here](https://github.com/gbdev/pandocs/blob/a9a3c0bdbd0fa7e27335a312ff42d1aac7a48db3/src/CGB_Registers.md?plain=1#L72) and [here](https://github.com/gbdev/pandocs/blob/a9a3c0bdbd0fa7e27335a312ff42d1aac7a48db3/src/CGB_Registers.md?plain=1#L90) (for each transfer mode).

I'm not sure if I prefer to have this info in a different section than the one describing the register behavior itself (like this PR suggesting) but I do believe we should have this info described in a single place (DRY 😄).

What do you think?

avivace (changes requested):
Yep, this information is already present (and twice) - we should have it only in 1 place and maybe link/quote that paragraph if helpful

alloncm:
Personally I think this is the better option:
> I can move the state of the register after a transfer to the sections describing FF51-FF52 and FF53-FF54

I completely missed the description of the FF51-FF54 registers and I think moving it to the section describing how the registers behaves will help people find it easily (the state of the register after transfer is done is tightly coupled to its behavior).

alloncm (commented):
While reviewing this, another question came to mind - what happens if only one register is written to after a transfer, does it updates only the relevant bits (high or low part) or it reloads both registers from the last values written.

My intuition says only the relevant bits but maybe is worth checking?

src/CGB_Registers.md
```diff
@@ -39,12 +39,14 @@ tested on Echo RAM, OAM, FEXX, IO and HRAM\]. Trying to specify a source
 address in VRAM will cause garbage to be copied.
 
 The four lower bits of this address will be ignored and treated as 0.
+The address specified by those registers is cached internally and incremented by $10 for each block of $10 bytes successfully transferred. The cached address persists until the registers are written again.
```
ISSOtm: I have a suggestion for different wording to hopefully convey the same info.

```suggestion
Writing to either of these registers updates part of an internal persistent address, which also gets incremented by 16 for each tile transferred.
```

ISSOtm (changes requested):

Phidias618:
To avoid any further confusion i have added a few examples.
To group all of the example into one single section i ended up creating a new section : "State of the source/destination registers after a transfer".
I also changed the wording a little bit.

src/CGB_Registers.md
```diff
@@ -39,13 +39,34 @@ tested on Echo RAM, OAM, FEXX, IO and HRAM\]. Trying to specify a source
 address in VRAM will cause garbage to be copied.
 
 The four lower bits of this address will be ignored and treated as 0.
+After a transfer the address contained in this register pair is 
+by $10 for each block of $10 bytes transfered, these
```
avivace, line 43: Maybe this can be moved to a footnote (the fact that they cannot be observed directly etc)

avivace (changes requested):

avivace:
Looks OK to me now, just a minor comment.

src/CGB_Registers.md
```diff
@@ -39,13 +39,34 @@ tested on Echo RAM, OAM, FEXX, IO and HRAM\]. Trying to specify a source
...
+multiple transfer in a row.
+For instance, a transfer of one large block is mostly equivalent to
+multiple transfers of smaller blocks, without needing to update the source
+nor the destination registers between each of the smaller transfers.
+Another use case would be to fill VRAM with the same $10 bytes block repeated 
+all over, as only the source address register would need be updated after each
+transfer, the destination register being automatically incremented by
+the block size after each transfer.
+
+
```
alloncm, line 69: IMO the "State of the VRAM DMA source/destination registers after a transfer" is redundant and just add noise to the page.

I think the best way representing this info is by adding the lines you added to `FF51-FF52` to `FF53-FF54` too by just copy pasting it.

Something like this:

```suggestion
After a transfer the address pointed by those two registers equals `initial_address + ($10 * num_of_blocks_transferred)`.
Since those are write-only registers, this can only be observed by doing another transfer without writing them again.

#### FF53–FF54 — HDMA3, HDMA4 (CGB Mode only): VRAM DMA destination (high, low) \[write-only\]

These two registers specify the address within 8000-9FF0 to which the
data will be copied. Only bits 12-4 are respected; others are ignored.
The four lower bits of this address will be ignored and treated as 0.

After a transfer the address pointed by those two registers equals `initial_address + ($10 * num_of_blocks_transferred)`.
Since those are write-only registers, this can only be observed by doing another transfer without writing them again.

```

alloncm (changes requested):
I believe the extra section adds unnecessary noise (and the examples don't add much that the inline description doesn't already cover). 
IMO keeping this info inline under each register's section makes it easier to find and keeps the page cleaner.

src/CGB_Registers.md
ISSOtm: I've found that it makes sense to describe the *overall* functioning of the system in a single place, so that the reader can start to form a vague mental picture before getting into the individual components; otherwise, IME, the scattered information is very hard to assemble into the big picture. This line of reasoning is what has led us to creating the [Graphics](https://gbdev.io/pandocs/Graphics.html) or [Audio](https://gbdev.io/pandocs/Audio.html) introduction pages.

ISSOtm (changes requested):
Thanks for the contribution! I think we may improve the flow of the page by making that section be a more general description of the VRAM DMA process, perhaps something like “VRAM DMA operation”? I think there's editorial work that can be done in reorganising this info into the neighbouring sections.

But, if @Phidias618 doesn't feel like going through that back-and-forth, please let me know and I can push some commits suggesting my changes; or, if you'd rather opt out of the entire review process, that's also fine, but please let us know so we don't end up pinging or waiting on you. Thanks!

src/CGB_Registers.md
alloncm: I pretty much agree with you about the overview sections for complex systems, I just don't think that this specific case (post-transfer register state) rises to that level.
It's niche enough that IMO most developers will never really need it, and I worry that a dedicated section gives it more focus than it deserves and risks burying the more critical parts about the VRAM DMA system.

PR #629, opened 2025-10-02
Add MASK_EN note

### LINKED ISSUE #625: SGB freeze-frame on LCD off should be mentioned
Probably in sgb-command-17

dmitry-shechtman:
Close #625

### DIFF
--- a/src/SGB_Command_System.md
+++ b/src/SGB_Command_System.md
@@ -24,6 +24,11 @@ after having displayed the picture). The Cancel Mask function may be
 also invoked (optionally) by completion of PAL_SET and ATTR_SET
 commands.
 
+:::tip
+
+SGB freezes/unfreezes the Game Boy screen automatically whenever the LCD is
+turned off/on.
+
 ## SGB Command $0C — ATRC_EN
 
 Used to enable/disable Attraction mode, which is enabled by default.


src/SGB_Command_System.md
```diff
@@ -24,6 +24,11 @@ after having displayed the picture). The Cancel Mask function may be
 also invoked (optionally) by completion of PAL_SET and ATTR_SET
 commands.
 
+:::tip
+
+SGB freezes/unfreezes the Game Boy screen automatically whenever the LCD is
+turned off/on.
```
avivace, line 30: I think we're missing the closing of the box here. See https://github.com/gbdev/pandocs/blob/master/CONTRIBUTING.md#custom-containers

dmitry-shechtman: Of course. I'm learning the syntax as I go.

avivace (approved):

ISSOtm:
Wait, really? I thought only another `MASK_EN` command could unfreeze the screen.

dmitry-shechtman:
Maybe it needs clarification. If the screen was frozen by clearing LCDC, it will be unfrozen when its bit 7 is set.

ISSOtm:
Ah, I see: it's another freezing mechanism, separate from `MASK_EN`, right?

dmitry-shechtman:
Yes. Enlightened by beware.

src/SGB_Command_System.md
```diff
@@ -24,6 +24,13 @@ after having displayed the picture). The Cancel Mask function may be
...
```
ISSOtm, line 30:
```suggestion
The screen stops updating when the "LCD" is [turned off](<#LCDC.7 — LCD enable>), though it gets blanked out after a while.
```

ISSOtm (changes requested):
Then I'd like to suggest the following wording:

src/SGB_Command_System.md
ISSOtm: AFAIK, leaving the screen turned off for a significant period of time causes the SGB firmware to blank out the framebuffer.

dmitry-shechtman: @nitro2k01?

ISSOtm: Addition of that claim is tracked by #625, btw.

dmitry-shechtman: Can we safely remove the "gets blanked out after a while" bit?

avivace: In abscence of proof regarding the 'blank out' I think the original wording is better, I'd keep that one.

PR #633, opened 2025-10-16
Document NR43 values freezing the channel

ISSOtm:
Turns out `EVER` and `ETOV` are not connected to `ETYR` (https://github.com/furrtek/DMG-CPU-Inside/blob/master/Schematics/20_CHANNEL4.png) which means that 14 and 15 don't select any clock source at all, freezing the channel.

### DIFF
--- a/src/Audio_Registers.md
+++ b/src/Audio_Registers.md
@@ -402,7 +402,7 @@ This register allows controlling the way the amplitude is randomly switched.
 - **Clock divider**: See the frequency formula below.
   Note that <var>divider</var> = 0 is treated as <var>divider</var> = 0.5 instead.
 
-The frequency at which the LFSR is clocked is <math><mfrac><mn>262144</mn><mrow><mi>divider</mi><mo>×</mo><msup><mn>2</mn><mi>shift</mi></msup></mrow></mfrac></math> Hz.
+The frequency at which the LFSR is clocked is <math><mfrac><mn>262144</mn><mrow><mi>divider</mi><mo>×</mo><msup><mn>2</mn><mi>shift</mi></msup></mrow></mfrac></math> Hz, except that <var>shift</var> being equal to 14 or 15 stops the channel from being clocked entirely.
 
 If the bit shifted out is a 0, the channel emits a 0; otherwise, it emits the volume selected in `NR42`.
 


nitro2k01:
Almost. Going by the schematic, It's `EMOF` (0) and `ELAR` (1) that don't select a clock source through ETYR. 

But something else is weird here. A higher value for clock shift (bits 4-7) should result in a higher power of 2 in the denominator and a lower resulting frequency by the formula we use to calculate the clock speed for the LFSR. Looking at the schematic, a higher value for clock shift selects a less divided tap of the divider, which should produce a higher frequency. I'd suspect that maybe the schematic is wrong in this instance, for example by confusing the normal and inverted lines for the FF22 bits. 

I'd first of all want some real life tests of how the hardware behaves.

nitro2k01:
I looked into it more. Here are my findings.

- I tested basic writing various values to NR43. It's definitely Ex an Fx that stop the clock, whereas for example Dx still output a slow sequence of geiger counter like clicks, meaning the clock is definitely going in that case. Ex and Fx should've output an even slower sequence of clicks if it followed the same pattern. This was tested on DMG, CGB, AGS units.
- The address decoding part of the schematic (EMOF-ETOV) is logically correct as far as I can tell, however it is mislabeled. All the bits indicated there should be inverted.
- One concern I have is that when no line is selected, nothing is driving the clock line, so the clock signal is undefined. It probably decays to either 0 or 1 in practice, but the time this takes might not be deterministic. This behavior (both whether it decays to 0 or 1, and the times this takes) may vary between hardware models. Meaning, you may or may not have 1 extra clock. If you just want to make some weird noises, this is largely irrelevant, but if you're hoping to control the DC level for sample playback, it's pretty important. So a big fat research needed on that one.

Another tangentially related thing that could need research is "NET03" which seems to indicate an APU test mode. From what I can tell, and this almost seems strange that no one has stumbled upon this by mistake, you should be able to activate by writing 1 to bit 4 of NR52. Although D4 _is_ gated by a myusterious "FROM_CPU" signal, so I don't know, maybe not that simple. (The AND gate EFOP in [9_APU_CONTROL](https://github.com/furrtek/DMG-CPU-Inside/blob/master/Schematics/9_APU_CONTROL.png) for reference.) 

Anyway, that would allow the LFSR clock signal to be output directly, which would simplify research in this case.

For the time being I think the wording in the PR is fine though.

ISSOtm:
Naw, the clock isn't unconnected, it's zero, since the muxer cell is a combination of AND and OR.

PR #648, opened 2026-01-25
Made `MBC3` article a little more clear

evanwporter:


avivace (changes requested):
I'm not in favor of this change. I believe the original wording is better. `As for the MBC1 `is more precise here, since the sentence refers specifically to MBC1’s bank-selection behavior rather than a general similarity (which seems to be suggested/implied by the 'with')

ISSOtm (approved):
Yes, that's the entire point: on this MBC, *like on MBC1*, a write of $00 maps to Bank 1. I prefer this new wording for that exact reason.

avivace:
> Yes, that's the entire point: on this MBC, _like on MBC1_, a write of $00 maps to Bank 1. I prefer this new wording for that exact reason.

What's the reason to prefer a wording that implies a weaker, more generic, similarity instead of a stronger, more concise one that trasmits better the concept (which is, MBC1’s bank-selection behavior is _exactly_ the same) ?

pinobatch:
"A stronger, more concise one" would still need communicate both of the following:
1. Like MBC1, MBC3 treats a written bank number $00 as if it were $01.
2. Unlike MBC1, MBC3 doesn't share the two most significant bits of bank number between SRAM A14-A13 and ROM A20-A19, nor does it have the mode setting at $6000 to select whether to zero those two bits at $0000-$3FFF and $A000-$BFFF.

Rangi42:
"As for X, Y" is a standard English idiom saying Y *about X*. So "As for the MBC1, writing ..." reads as saying something about how writing a value works *for the MBC1*. The current sentence is incorrect and the PRed sentence is correct.

That said, the first sentence of this paragraph could also be better. "Same as for MBC1" is an unnecessarily terse sentence fragment -- a whole sentence like "This register works the same as with the MBC1" would be more clear[^1] (and slightly less concise, which is a fine tradeoff). However, the "Same as for X" pattern is used for many registers in multiple MBC pages, so that should be a separate PR.

[^1]: The sentence could be "This register works the same as for/with/on/in/etc the MBC1", depending on which preposition Pan Docs generally uses to connect MBCs with their registers. I'm not commenting on what is or should be the correct/conventional one. IMO any of them would be fine.

avivace (approved):

avivace:
I got it now @Rangi42 thanks !!

PR #652, opened 2026-03-26
Upgrade crates to support mdBook 0.5 breaking changes

### LINKED ISSUE #641: Upgrade to mdbook 0.5.x
The custom code (and the base mdbook settings) need to be upgraded to mdbook 0.5.x as at the moment we have to lock everything to 0.4.x to build.

See migration guide here: https://github.com/rust-lang/mdBook/blob/master/CHANGELOG.md#05-migration-guide

mtthgn:
This PR specifically does the following:

1. It removes the `multilingual` attrbitue from `book.toml`. This was never really used and was removed as part of 0.5:
    - https://github.com/rust-lang/mdBook/pull/2775
 2. It makes the `preproc` crate now depend on `mdbook-preprocessor`. As part of 0.5 the mdbook crate was broken down into components.
    - The `PreProcessor` trait slightly changed. The `supports_render` function now returns a `Result<bool>` instead of merely a `bool`: https://github.com/rust-lang/mdBook/pull/2797
    - `get_preprocessor` was removed, so I had to use `get` instead: https://github.com/rust-lang/mdBook/pull/2773
 3. It makes the `renderer` crate now depend on `mdbook-renderer` and `mdbook-html`. This one is mostly just updating `use` names.

I tested this with both docker and locally installing everything and was able to get the docs rendering on localhost with both development approaches. Things seemed to look right to me, but an extra set of eyes would be helpful.

I'd also appreciate confirmation I didn't break expectations with the `out_of_repo` functionality. I wasn't really following what that flag is for, but I believe I'm still correctly reading it from the config.

Closes #641


avivace (changes requested):

avivace (approved):
Great work @mtthgn - Export looks ok to me!  Only small things remaining is the one page export which is still carrying the 'title' of the single.md page we replace at the bottom.

ISSOtm:
@avivace Please do not merge non-trivial PRs unilaterally.

PR #656, opened 2026-04-01
Fixes unintentional removal of the search feature

### LINKED ISSUE #655: Latest update removed Search funcionality
Previous version had an icon that opened the search bar.
It was very useful to search for arbitrary RAM registers, like "LYC" or "FF45".

Image

New version doesn't have it.

Image

mtthgn:
When Pandocs was upgraded to 0.5.2, the custom renderer was no longer rendering search because of a change that gated search functionality behind a cargo feature. This commit ensures that feature is set on the renderer crate.

Closes #655

### DIFF
--- a/renderer/Cargo.toml
+++ b/renderer/Cargo.toml
@@ -15,3 +15,7 @@ mdbook-html = "0.5.2"
 regex = "1.10"
 termcolor = "1.4.1"
 url = "2.5.2"
+
+[features]
+default = ["search"]
+search = ["mdbook-html/search"]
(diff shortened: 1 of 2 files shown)


avivace (approved):

renderer/Cargo.toml
```diff
@@ -15,3 +15,7 @@ mdbook-html = "0.5.2"
 regex = "1.10"
 termcolor = "1.4.1"
 url = "2.5.2"
+
+[features]
+default = ["search"]
+search = ["mdbook-html/search"]
```
ISSOtm, line 21: IMO we might as well enable it always, it's not like this renderer is ever used with the search *disabled*.

ISSOtm (commented):
Thanks for looking into this!

PR #657, opened 2026-04-15
Note OBJ_TRN and TEST_EN do not work on any SGB retail unit

CasualPokePlayer:
This is easily confirmed with disassembly (the OBJ_TRN and TEST_EN handlers are an immediate rts on all retail SGB revisions). Some (but not all) beta SGB BIOS revisions appear to not have this rts and thus appear to have these commands functional.

### DIFF
--- a/src/SGB_Command_Border.md
+++ b/src/SGB_Command_Border.md
@@ -101,6 +101,8 @@ Thus a border can use three 15-color palettes.
 
 ## SGB Command $18 — OBJ_TRN
 
+This command does nothing on all retail SGB revisions. The following description only applies to some prototype units.
+
 Used to start transferring object attributes to SNES object attribute memory (OAM). Unlike all other
 functions with names ending in "\_TRN", this function does not use the usual
 one-time 4 KiB VRAM transfer method. Instead, when enabled (below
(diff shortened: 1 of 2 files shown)


avivace (approved):

ISSOtm:
Well, I [was disassembling](https://codeberg.org/ISSOtm/sgb-bios) three revisions of the SGB1 firmware (whose hashes are not labelled “beta” on Dat-O-Matic, [v0](https://datomatic.no-intro.org/index.php?page=show_record&s=49&n=2683) — [v1](https://datomatic.no-intro.org/index.php?page=show_record&s=49&n=2684) — [v2](https://datomatic.no-intro.org/index.php?page=show_record&s=49&n=2685)), and from what I remember, v1 added the `RTS`, but the `OBJ_TRN` functionality *is* present in v0. So I'm a little skeptical of this claim.

CasualPokePlayer:
https://github.com/leinacc/sgb0-disasm/blob/c36b3c3e68dbcf95c5da1c94a54413427c84faf8/code/bank_000.s#L12941

The disassembly assembles to v0. This is also easy to check in a hex editor.

ISSOtm (changes requested):
Confirmed the above. Thanks for your research!

Now, I do have another suggestion: `OBJ_TRN` is positioned weirdly, and at this point I'm thinking it may be appropriate to move those two commands to a new section—perhaps “SGB prototype commands”?

avivace:
@ISSOtm can you open an Issue / another PR directly for the creation of such new section? I'll merge this in the meantime to keep the scope and not block it.

avivace (approved):

PR #637, opened 2025-11-16
Update MBC3.md

nummacway:
Results from CasualPokePlayer and me, as discussed in the Research channel on Discord on Tuesday. Also make RAM bank count and RAM size consistent and add some more details how it handles wirting invalid values.

PR #661, opened 2026-04-18
Add MBC30 mapper

### LINKED ISSUE #512: Mention MBC30 in MBC3 documentation
Pokemon Crystal JP was the only game to use this MBC, which was basically MBC3 but with 8 SRAM banks instead of 4 (and technically allowing 256 ROM banks instead of 128, although Crystal didn't use that). It's mentioned in a footnote on the cartridge header page, but nowhere in the MBC section. More flashcarts are supporting MBC30 (i.e. allowing 8 SRAM banks and 256 ROM banks) as games like Pokemon Coral and Crystal Clear make use of it, so I think this should be documented more clearly.

quinnyo:
This adds a section to the MBC3 page describing the MBC30 mapper, thus resolves #512.

On balance I think MBC30 is different enough to warrant some isolation -- avoiding complicating the MBC3 documentation -- and similar enough to piggyback on the MBC3 page.

// I started working on this when #637 got closed (which is now open again). I'm making this PR now just because that's what I was heading towards doing, but I'm happy to adapt this or whatever really!

### DIFF
--- a/src/MBC3.md
+++ b/src/MBC3.md
@@ -92,3 +92,12 @@ Year-10000-Proof, provided that the cartridge gets used at least every
 
 When accessing the RTC Registers, it is recommended to wait 4 µs
 (4 M-cycles in Normal Speed Mode) between any separate accesses.
+
+
+# MBC30
+
+(4 MiB ROM, 64 KiB RAM, timer)
+
+The MBC30 is practically identical to MBC3 in operation, but is capable of addressing twice as much memory for both ROM and RAM.
+The only title to be shipped with the MBC30 mapper was _Pocket Monsters: Crystal Version_ in Japan, with the various worldwide versions using MBC3.
+


src/MBC3.md
```diff
@@ -92,3 +92,12 @@ Year-10000-Proof, provided that the cartridge gets used at least every
...
 When accessing the RTC Registers, it is recommended to wait 4 µs
 (4 M-cycles in Normal Speed Mode) between any separate accesses.
+
+
+# MBC30
+
+(4 MiB ROM, 64 KiB RAM, timer)
+
+The MBC30 is practically identical to MBC3 in operation, but is capable of addressing twice as much memory for both ROM and RAM.
+The only title to be shipped with the MBC30 mapper was _Pocket Monsters: Crystal Version_ in Japan, with the various worldwide versions using MBC3.
```
ISSOtm, line 102: I've heard from @Rangi42 that it may actually be a MBC3O, possibly “oversized” or something like that. What do you think?

We may nonetheless want to mention the “MBC30” name as a common alias in online documentation.

ISSOtm (changes requested):
Thanks! I'd rather review this specific PR for the MBC3O-specific changes, and then go back to #637 for more general layout changes.

src/MBC3.md
CasualPokePlayer: https://gbhwdb.gekkio.fi/cartridges/CGB-BXTJ-0/gekkio-1.html
https://gbhwdb.gekkio.fi/cartridges/CGB-BXTJ-0/gekkio-2.html

That is a 0, not an O (compare it with the other 0 on the chip, also look at the board's MBC30's 0 and 0's elsewhere on the board).

Rangi42: The "MBC3O" naming was first noticed by @pinobatch: https://github.com/gbdev/pandocs/issues/512#issuecomment-1864811717

> Is it really `MBC30` with a zero or `MBC3O` with a letter, meaning "MBC3 Oversize"?

> If a letter, that would match a pattern on NES where O designates a variant of another cartridge board designed to take a larger memory. This includes UOROM (UNROM + 2 Mbit ROM instead of 1 Mbit) and SOROM (SNROM + two 64 kbit SRAM instead of one).

I see an "O" in "KOREA", but don't notice any clear difference between it and whichever glyph is in the "MBC3<i>_</i>" name.

Rangi42:
This PR fixes #512.

src/MBC3.md
ISSOtm: It's plausible that the font renders these two glyphs identically. I (we?) am making a judgement call here, because the “O” terminology seems to make sense, like we ended up taking a stab in the dark with the SM83 naming.

Unless anyone has ties with some Nintendo-working uncle, I think we'll have to stick to a guess.

quinnyo: It looks like a zero to me. The only known reference to this part number is what you can see on the PCB and mapper IC. I don't see any evidence for it being a letter 'O'.

It being a zero isn't a guess:
In the photos above, there are three instances of the mapper part number, `MBC3~` with `~` being the mystery character. The two boards (gekkio-1, gekkio-2) each feature a mapper IC from a different manufacturer. The PCB silkscreen also includes the part number.
- Each of these three locations has a `~` character unique to it.
- The mapper ICs both feature a three digit week of manufacture code, both with a leading zero.
  - The leading zero on the week code matches the `~` in the part number on the same IC.
  - again, the `~` are clearly different between the two chips, but they match the zero etched on the same IC.
- The PCB silkscreen features multiple instances of unambiguous zeroes.
  - Of particular interest is `CR2025`, which is a close match for the size of the `MBC3~` text. `CR2025` is the onboard button cell designation. This is definitely a zero: `20` is the diameter in mm, `25` is the height in tenths of mm.

Markings on ICs other than the mapper should not be considered -- there's no reason to expect the typeface to match across manufacturers, plants, etc.

The NES 'O' thing:
- This chip first appeared in 2000. I assume the NES 'O' carts are at least ten years older than that?
- The MBC naming scheme generally doesn't appear to match the NES/Famicom one in any significant way, with either 0/O.

avivace (approved):

avivace (approved):
I believe only the (imo, unresolvable) 0 vs O question remains. Feel free to open an Issue re: that

PR #662, opened 2026-04-19
Rewrite BCPD description of the colour data format

### LINKED ISSUE #643: Rewrite Palettes/BCPD/BGPD to be more intuitive
Currently, the bit layout for BGPD is difficult to read properly. The bits are arranged left to right from 0 to 15 as "little-endian RGB555", but the CGB uses these bits in reverse order as BGR555 and they are written to BGPD in this format, too.

quinnyo:
This is a rewrite of the BCPD section to hopefully reduce possible confusion and make the colour data format less ambiguous.
Particularly:
- avoid relying on reader interpretation of 'RGB555'
- make explicit that "little-endian" is about byte order of the 16 bit colour word
- reverse the bit table so MSB is first (this table is just about the only one that used the LSB order)
- explicitly state that you access one byte at a time (need to read/write twice per colour)

Fixes #643 (I think)


nummacway:
I suggest we add an example:
Farbberechnung

ISSOtm:
That's actually a pretty good idea! A visual representation of how the encoding is transformed is very helpful.

What I would like, actually, is to add a third layer, which shows how the colour is converted into RGB555 format, and *then* shows the two bytes being swapped in memory. Is the above image available in SVG form? I'd like to tweak it to propose some further improvements. (Perhaps we can add that in a follow-up PR, but still.)

src/Palettes.md
```diff
@@ -59,15 +59,19 @@ Unlike BCPD, this register can be accessed outside VBlank and HBlank.
 
 ### FF69 — BCPD/BGPD (CGB Mode only): Background color palette data / Background palette data
 
-This register allows to read/write data to the CGBs background palette memory, addressed through [BCPS/BGPI](<#FF68 — BCPS/BGPI (CGB Mode only): Background color palette specification / Background palette index>).
-Each color is stored as little-endian RGB555:
+This register provides read/write access to the CGB background palette memory.
+The byte accessed is specified in [BCPS/BGPI](<#FF68 — BCPS/BGPI (CGB Mode only): Background color palette specification / Background palette index>).
```
ISSOtm:
```suggestion
This register provides read/write access to the byte located at the address within the CGB's background palette memory specified in [BCPS/BGPI](<#FF68 — BCPS/BGPI (CGB Mode only): Background color palette specification / Background palette index>).
```

src/Palettes.md
```diff
@@ -59,15 +59,19 @@ Unlike BCPD, this register can be accessed outside VBlank and HBlank.
...
+The color data format is RGB555, meaning a single color is composed of three 5 bit components, one for each of red, green, and blue.
```
ISSOtm:
```suggestion

The color data format is RGB555, meaning a single color is composed of three 5-bit components, one for each of red, green, and blue.
```

src/Palettes.md
```diff
@@ -59,15 +59,19 @@ Unlike BCPD, this register can be accessed outside VBlank and HBlank.
...
+Each 15 bit color occupies the lower part of a 16 bit word, with the 16th bit unused:
```
ISSOtm:
```suggestion
Each 15-bit color occupies the lower part of a 16-bit word, with the 16th bit unused:
```

src/Palettes.md
```diff
@@ -59,15 +59,19 @@ Unlike BCPD, this register can be accessed outside VBlank and HBlank.
...
-{{#bits 16 <
-  "One color"  0-4:"Red intensity" 5-9:"Green intensity" 10-14:"Blue intensity";
+{{#bits 16 >
+  "Channel"  15:"N/A" 14-10:"Blue" 9-5:"Green" 4-0:"Red";
 }}
 
-Much like VRAM, data in palette memory cannot be read or written during the time
-when the PPU is reading from it, that is, [Mode 3](<#PPU modes>).
+As the colors are two bytes in size, you must read/write this register *twice* to access a whole color.
+The low byte (bits 0-7) is stored first (lower) in memory -- with *litle-endian* byte order.
```
ISSOtm:
```suggestion
The low byte (bits 0-7) is stored first (lower) in memory (*little-endian* byte order).
```

src/Palettes.md
```diff
@@ -59,15 +59,19 @@ Unlike BCPD, this register can be accessed outside VBlank and HBlank.
...
 
-{{#bits 16 <
-  "One color"  0-4:"Red intensity" 5-9:"Green intensity" 10-14:"Blue intensity";
+{{#bits 16 >
+  "Channel"  15:"N/A" 14-10:"Blue" 9-5:"Green" 4-0:"Red";
 }}
 
-Much like VRAM, data in palette memory cannot be read or written during the time
-when the PPU is reading from it, that is, [Mode 3](<#PPU modes>).
+As the colors are two bytes in size, you must read/write this register *twice* to access a whole color.
```
ISSOtm:
```suggestion
As each color is two bytes in size, you must read/write this register *twice* to access a whole color.
```

ISSOtm (changes requested):
Thank you for the wording improvements!

avivace:
[rgb555_encoding(1)(1).drawio](https://github.com/user-attachments/files/26901478/rgb555_encoding.1.1.drawio)
I've started trying to redraw it in draw.io @ISSOtm

nummacway:
Are these two distinct proposals? Or are they two distinct pictures that are both to be added?

For my design, I actually like my "ATA/ATAPI cable" style because it is immediately clear which and _how many_ bits are involved.

avivace:
I like the cable design as well @nummacway but I couldn't find a way to reproduce it in draw.io (which  keeps the diagrams in a 'editable' project XML files so it's a preferred format here) :/ 

How was your diagram drawn?

nummacway:
Didn't know draw.io was preferred.
I'll change your diagram to snap to grid, to make future updates easier, and will try to have it import my shapes. I found no way to draw this with draw.io's tools.
I drew them in Affinity, because that's what I use. Because the way it exports text, it was never meant as the final result, and I would have recreated it with a text editor.

nummacway:
This is only the top part of the original file.
[Farbberechnung.drawio](https://github.com/user-attachments/files/26909490/Farbberechnung.drawio)
I found no way of inserting an editable SVG, so I had to recreate my four custom shapes with drawio's language. For reference, these are their source codes (concatenated in a single file):
[ataatapi-shapes.xml](https://github.com/user-attachments/files/26909493/ataatapi-shapes.xml)
I do find my two separate colors for the green "cables" nice, but I have no idea which colors to use to align with Pan Docs. I leave that to you.

For those who don't want to open drawio, here's a screenshot:
image

ISSOtm:
I've started transcribing the design as a themable SVG (since no matter how much I try, I can't edit SVGs in graphical editors ^^') I'll host a rendered version online when I've gotten parity, though I also have design changes to suggest. In particular, this uses Pan Docs' font, which has different glyph widths, thus we may consider tweaking the font size? (Though I'm also thinking of swapping the values of the red and green components, simply to balance the amount of characters? xD)

image

(Disregard the misaligned digits, I haven't bothered fixing those yet.) One change I've made is that the “ATA ribbons” are drawn *behind* the borders, which I'm not fully sure about. One possible change is to squish the ribbons vertically by 1 pixel at each edge, so that the black edges still render there. Feel free to throw feedback!

ISSOtm:
> I do find my two separate colors for the green "cables" nice, but I have no idea which colors to use to align with Pan Docs. I leave that to you.

One layout idea I had was to separate the format conversion and the RAM encoding into two separate layers, so there'd be three total: the RGB8 colour, the RGB555 version (still with the “ribbon cables”, and then two arrows pointing out how the bytes are swapped in memory order, and also dotted lines going left and right to suggest that there are more memory cells around.

quinnyo:
I like the diagram, but it took me a little while to understand where you're coming from.
I see now that it's about the relationship or converting between 24-bit RGB and the CGB. I did not think about 24-bit colour or really any other format or conversion at all while I was working on this. [This probably says more about me than anything else.]

So I don't think I'm really the target audience for the diagram but I can see the value in it because a lot of people would be coming into this more familiar with this flavour of 24-bit colour than any other. It's a good way to illustrate *"what RGB555 means"* on a technical level.

<hr>

I've got one issue with the diagram itself: the order of the ribbon layers. Could the blue ribbon be on a higher layer than the green one? (blue should occlude green) RGB --> R, G_low, G_high, B

<hr>

@ISSOtm @avivace thanks for the review and patches!
[every time I write 'N(-)bit(s)' I have to check what I'm supposed to do, and this time I didn't! 🐧]

nummacway:
The layer idea was: Small cables over wide cables.

ISSOtm:
[My current iteration](https://codeberg.org/ISSOtm/pandocs/src/branch/bcpd-colour-format) of the diagrams can be viewed online at https://eldred.fr/pandocs+662/Palettes.html#lcd-color-palettes-cgb-only; criticism welcome.

[EDIT]: Updated with the state of the PR as of https://github.com/gbdev/pandocs/pull/662#issuecomment-4391229106.

nummacway:
"5-bit red" is 16 not 17.

ISSOtm:
Since there have been no replies to my last comment in the last two weeks, I'm going to push my changes to this PR, and we'll keep working from there. Anyone opposed to this, please voice your opinion (now or after) and those changes will be reverted.

ISSOtm:
I'm now personally happy with the state of this PR, but I'd like opinions from @quinnyo and @nummacway before merging.

ISSOtm:
Also @DelayRGC, since she opened the original issue.

src/Palettes.md
```diff
@@ -30,70 +29,64 @@ They work exactly like [`BGP`](<#FF47 — BGP (Non-CGB Mode only): BG palette da
...
-can be addressed.
+Unlike VRAM, OAM, or wave RAM, CRAM cannot be accessed directly, but instead through pairs of registers: one “data” register provides a one-byte view into CRAM, while the corresponding “index” register controls which CRAM address the data register is bound to.
+Much like VRAM, CRAM is inaccessible when the PPU is reading from it, that is, during [Mode 3](<#PPU modes>): writes are ignored, and reads return $FF.
 
-First comes BGP0 color number 0, then BGP0 color number 1, BGP0 color number 2, BGP0 color number 3,
-BGP1 color number 0, and so on. Thus, address $03 allows accessing the second (upper)
-byte of BGP0 color #1 via BCPD, which contains the color's blue and upper green bits.
+[^bit15]:
+The 16th bit, bit 15, is **ignored** by the hardware.
+In discussion, that bit is generally clear (for example, the canonical pure white is `7FFF` and not `FFFF`), but the hardware treats both identically: it's fine to fill color RAM with $FF bytes to set it to all-white!
```
avivace:
```suggestion
Conventionally, that bit is generally clear (for example, the canonical pure white is `7FFF` and not `FFFF`), but the hardware treats both identically: it's fine to fill color RAM with $FF bytes to set it to all-white.
```

'In discussion' sounds a bit confusing here, imo

avivace: also the ! is a bit off tone

src/Palettes.md
```diff
@@ -30,70 +29,64 @@ They work exactly like [`BGP`](<#FF47 — BGP (Non-CGB Mode only): BG palette da
...
-Each color is stored as little-endian RGB555:
-
-{{#bits 16 <
-  "One color"  0-4:"Red intensity" 5-9:"Green intensity" 10-14:"Blue intensity";
-}}
+As each color is two bytes in size, you must read/write this register *twice* to access a whole color.
 
-Much like VRAM, data in palette memory cannot be read or written during the time
-when the PPU is reading from it, that is, [Mode 3](<#PPU modes>).
+This is made much easier through the use of the address auto-increment: `BGPI`'s “address” field is automatically incremented (possibly wrapping around) after each write to this register, even if the write fails due to CRAM being inaccessible.
```
avivace:
```suggestion
This is made much easier through the use of the address auto-increment: `BGPI`'s "address" field is automatically incremented (possibly wrapping around) after each write to this register, even if the write fails due to CRAM being inaccessible.
```

src/Palettes.md
```diff
@@ -30,70 +29,64 @@ They work exactly like [`BGP`](<#FF47 — BGP (Non-CGB Mode only): BG palette da
...
+Reads, however, never trigger auto-increment.
 
 :::tip NOTE
 
-All background colors are initialized as white by the boot ROM, however it is a
-good idea to initialize all colors yourself, e.g. if implementing
-a soft-reset mechanic.
+All background colors are initialized as white by the boot ROM, however it is a good idea to initialize all colors yourself, e.g. if implementing a soft-reset mechanic.
```
avivace:
```suggestion
All background colors are initialized as white by the boot ROM, however it is a good idea to initialize all colors yourself, e.g. by implementing a soft-reset mechanic.
```

_e.g.  if implementing ..._ doesn't sound good

src/Palettes.md
```diff
@@ -30,70 +29,64 @@ They work exactly like [`BGP`](<#FF47 — BGP (Non-CGB Mode only): BG palette da
...
 
-These registers function exactly like BCPS and BCPD respectively; the 64 bytes
-of OBJ palette memory are entirely separate from Background palette memory, but
-function the same.
+These registers function exactly like BGPI and BGPD respectively; the 64 bytes of OBJ palette memory are entirely separate from Background palette memory, but function the same.
 
-Note that while 4 colors are stored per OBJ palette, color #0 is never used, as
-it's always transparent. It's thus fine to write garbage values, or even leave
-color #0 uninitialized.
+Note that while 4 colors are stored per OBJ palette, color #0 is never used, as it's always transparent. It's thus fine to write garbage values, or even leave color #0 uninitialized.
```
avivace:
```suggestion
Note that while 4 colors are stored per OBJ palette, color #0 is never used, as it's always transparent. It's thus fine to write garbage values, or even leave it uninitialized.
```

avivace (changes requested):
Some minor comments

src/Palettes.md
DelayRGC: "by" changes the meaning of the sentence. Maybe "for" would work better?

src/Palettes.md
ISSOtm: I wanted to avoid implying that it would be convention outside of the programs themselves. Perhaps "in writing"?

src/Palettes.md
ISSOtm: That would also change the meaning of the sentence. The point is that it's a good idea to set all palettes yourself if there's any chance that the init code may run more than once, such as if the game has a soft-reset mechanism.

DelayRGC: What do you think of "for a soft-reset mechanic" rather than "for/by/if implementing a soft-reset mechanic"?

src/Palettes.md
nummacway: Proposal (that is definitely not meant as a final idea, but maybe some words can be taken up):

> When referring to a certain color, the 16-bit value is usually written with the highest bit set to 0, e.g. white will be referred to as $7FFF, not $FFFF.

(Also note the dollars.)

src/Palettes.md
nummacway: I like Isso's original proposal.

src/Palettes.md
ISSOtm:
```suggestion
Colors are conventionally referred to with that bit set to 0 (for example, the canonical pure white is `7FFF` and not `FFFF`), but the hardware treats both identically: thus, it's fine to fill color RAM with $FF bytes to set it to all-white.
```

I think it's better not to use the dollars here, as they'd imply numbers slightly more than colours. But that's fairly nitpicky.

nummacway: Can we get an "e.g." after the colon?

ISSOtm: Is this better?

nummacway: I'm undecided on that one.

I would like to add one thing about the dollar sign: The notation without it could become ambiguous if no A-F digits are present. I think we need some way to mark that the color values (I'm not calling them colors!) are hexadecimal. Since we don't have the power (or need) for a unique notation like CSS's `#`, $ will do.

avivace: "referred to with that bit set to 0" sounds even more quirky, IMO

what about "The canonical form of a color value has that bit set to 0." ?

ISSOtm: Again, no, since that implies that this is also the form that should be stuck to on the hardware itself. I'm trying to keep it clear that this is only the case in literature. (I switched to “that bit set to 0” because “that bit clear” is easy to misread, I've found.)

quinnyo:
Hi, sorry I haven't been around, I'll be able to have a look at this in a day or so!

ISSOtm:
Ping?

avivace (approved):

quinnyo:
Reviewing now, sorry!

src/Palettes.md
```diff
@@ -30,70 +29,64 @@ They work exactly like [`BGP`](<#FF47 — BGP (Non-CGB Mode only): BG palette da
...
-of the hardware interface, palette RAM (or *CRAM* for *Color RAM*) is not
-accessed directly, but instead through the following registers:
+Colors on the Game Boy Color are stored as RGB555, meaning a single color is composed of three 5-bit components, one for each of red, green, and blue.
+Each 15-bit color occupies the lower part of a 16-bit “word”[^bit15]:
+
+{{#include imgs/src/rgb555.svg}}
+
+The color palettes are stored in two dedicated banks of palette RAM (or <abbr title="Color RAM">CRAM</abbr> for *color RAM*), 64 bytes each[^cram_size]: one for background/window palettes and the other for OBJ palettes.
+
+The two bytes of each color are stored **little-endian**, meaning that the low byte comes first.
```
quinnyo: explicit byte order -- avoid confusion about most/least significant bit

```suggestion
The two bytes of each color are stored in **little-endian** byte order, meaning that the low byte comes first.
```

src/Palettes.md
```diff
@@ -30,70 +29,64 @@ They work exactly like [`BGP`](<#FF47 — BGP (Non-CGB Mode only): BG palette da
...
-Since there are 8 palettes, 8 palettes × 4 colors/palette × 2 bytes/color = 64 bytes
-can be addressed.
+Unlike VRAM, OAM, or wave RAM, CRAM cannot be accessed directly, but instead through pairs of registers: one “data” register provides a one-byte view into CRAM, while the corresponding “index” register controls which CRAM address the data register is bound to.
+Much like VRAM, CRAM is inaccessible when the PPU is reading from it, that is, during [Mode 3](<#PPU modes>): writes are ignored, and reads return $FF.
 
-First comes BGP0 color number 0, then BGP0 color number 1, BGP0 color number 2, BGP0 color number 3,
-BGP1 color number 0, and so on. Thus, address $03 allows accessing the second (upper)
-byte of BGP0 color #1 via BCPD, which contains the color's blue and upper green bits.
+[^bit15]:
+The 16th bit, bit 15, is **ignored** by the hardware.
```
quinnyo: It's real though, right? I mean it's stored somewhere. The PPU "ignores" it, I suppose:

```suggestion
The 16th bit, bit 15, has no effect on the colour and is not used for any other purpose by the system.
```

src/Palettes.md
```diff
@@ -30,70 +29,64 @@ They work exactly like [`BGP`](<#FF47 — BGP (Non-CGB Mode only): BG palette da
 
 ## LCD Color Palettes (CGB only)
 
-The CGB has a small amount of RAM used to store its color palettes. Unlike most
-of the hardware interface, palette RAM (or *CRAM* for *Color RAM*) is not
-accessed directly, but instead through the following registers:
+Colors on the Game Boy Color are stored as RGB555, meaning a single color is composed of three 5-bit components, one for each of red, green, and blue.
+Each 15-bit color occupies the lower part of a 16-bit “word”[^bit15]:
```
quinnyo: You don't need to put quotes around 'word', it's a well established term in computing and (remarkably) this usage matches its lettery ancestor quite well.

```suggestion
Each 15-bit color occupies the lower part of a 16-bit word[^bit15]:
```

src/Palettes.md
```diff
@@ -30,70 +29,64 @@ They work exactly like [`BGP`](<#FF47 — BGP (Non-CGB Mode only): BG palette da
 
 ## LCD Color Palettes (CGB only)
```
quinnyo, line 33: Introduction? CGB has colours, palettes.

The old bit summing up the CRAM size was essentially doing this before. (Not that that was a good solution)

src/Palettes.md
```diff
@@ -30,70 +29,64 @@ They work exactly like [`BGP`](<#FF47 — BGP (Non-CGB Mode only): BG palette da
...
 
-The CGB has a small amount of RAM used to store its color palettes. Unlike most
-of the hardware interface, palette RAM (or *CRAM* for *Color RAM*) is not
-accessed directly, but instead through the following registers:
+Colors on the Game Boy Color are stored as RGB555, meaning a single color is composed of three 5-bit components, one for each of red, green, and blue.
+Each 15-bit color occupies the lower part of a 16-bit “word”[^bit15]:
+
+{{#include imgs/src/rgb555.svg}}
+
+The color palettes are stored in two dedicated banks of palette RAM (or <abbr title="Color RAM">CRAM</abbr> for *color RAM*), 64 bytes each[^cram_size]: one for background/window palettes and the other for OBJ palettes.
```
quinnyo, line 47: Palettes haven't really been introduced at this point, but this reads like they have been.

src/Palettes.md
quinnyo: I'm not sure if the point here is truly about this "convention" or if this note is intended as a way to illustrate that bit 15 is not used by the system.

- If it's about convention, I'd recommend removing the statement entirely. This is because (I think that) Pan Docs can/should avoid commenting on, relying on, or defining convention whereever possible/reasonable.
- If it's an illustrative example, remove the reference/s to convention/canon. Strictly speaking, the reader already has the information, so rephrase this as a caution/tip.
  - Possibly focusing more on (not) making assumptions about bit 15, one way or the other, rather than just about `$FFFF` and `$7FFF` both being white.

<hr>

dollar prefix:
I'd prefer to see the dollar sign here. Maybe this isn't as common as I'd expect, but to me these prefixes are just a way to tell the parser (human or otherwise) that the value is rendered in a particular way. So `0x` or `$` just signifies "hexadecimal nybbles follow", not really anything about the meaning of the data.

A (two's complement) signed integer must have the same issue -- in other words, the signed integer `$FFFF` is not rendered as a number. You have to expend mental effort to interpret it.

<hr>

> I'm not calling them colors!

I agree. I began using the term 'colour word' in my initial changes. I chose to not use 'colour value', though, because of the existing meaning/s of 'value' in colour theory / visual art / HSV.

src/Palettes.md
```diff
@@ -30,70 +29,64 @@ They work exactly like [`BGP`](<#FF47 — BGP (Non-CGB Mode only): BG palette da
...
+The two bytes of each color are stored **little-endian**, meaning that the low byte comes first.
+For example, the two palettes shown in the previous diagram would be stored like this:
 
-### FF68 — BCPS/BGPI (CGB Mode only): Background color palette specification / Background palette index
+{{#include imgs/src/color_ram.svg}}
 
-This register is used to address a byte in the CGB's background palette RAM.
-Since there are 8 palettes, 8 palettes × 4 colors/palette × 2 bytes/color = 64 bytes
-can be addressed.
+Unlike VRAM, OAM, or wave RAM, CRAM cannot be accessed directly, but instead through pairs of registers: one “data” register provides a one-byte view into CRAM, while the corresponding “index” register controls which CRAM address the data register is bound to.
```
quinnyo: split long sentence
index before access

```suggestion
Unlike VRAM, OAM, or wave RAM, CRAM is not exposed in the memory map and cannot be accessed directly.
Instead, each bank of CRAM is accessed through a pair of registers: one register is used to select a CRAM address, and the other provides read/write access to the byte at that address.
```

src/Palettes.md
```diff
@@ -30,70 +29,64 @@ They work exactly like [`BGP`](<#FF47 — BGP (Non-CGB Mode only): BG palette da
...
+Much like VRAM, CRAM is inaccessible when the PPU is reading from it, that is, during [Mode 3](<#PPU modes>): writes are ignored, and reads return $FF.
```
quinnyo: With the prior point starting "Unlike VRAM, ..." this one starting "Much like VRAM" makes a slightly off balance contrast.
You could put a "However, ..." at the start, but that reinforces it as an intentional contrast, which doesn't really make sense.

src/Palettes.md
quinnyo: Yes, changing it to "by" would make it have a different, also undesirable/confusing meaning. But the "if implementing" thing *is* odd, and the whole point being made is a bit strange, actually:

When *wouldn't* it be a good idea to initialise ~~all~~ the colours yourself? If they are all white you can't see anything.
Regardless of the behaviour of the boot ROM, it's a "good idea" to initialise ~~the~~some palettes, because you want your palette/s to be the ones being used. At least I think so.

[Note: I don't actually know how true the boot ROM statement is, I assume someone checked.]

I know this behaviour is specific to the BG colours, but this "tip" (and the similar one for OB) could be moved up to the general explanation above -- it's really not specific to the register/s.

src/Palettes.md
```diff
@@ -30,70 +29,64 @@ They work exactly like [`BGP`](<#FF47 — BGP (Non-CGB Mode only): BG palette da
...
-Each color is stored as little-endian RGB555:
-
-{{#bits 16 <
-  "One color"  0-4:"Red intensity" 5-9:"Green intensity" 10-14:"Blue intensity";
-}}
+As each color is two bytes in size, you must read/write this register *twice* to access a whole color.
 
-Much like VRAM, data in palette memory cannot be read or written during the time
-when the PPU is reading from it, that is, [Mode 3](<#PPU modes>).
+This is made much easier through the use of the address auto-increment: `BGPI`'s "address" field is automatically incremented (possibly wrapping around) after each write to this register, even if the write fails due to CRAM being inaccessible.
```
quinnyo: "possible wrapping" makes it sound unreliable

You could say "mod 16" or (maybe more likely to ensure comprehension in this context) "bitwise `$1F`

```suggestion
This is made much easier through the use of the address auto-increment: `BGPI`'s "address" field is automatically incremented (wrapping around to zero after index 15) after each write to this register, even if the write fails due to CRAM being inaccessible.
```

quinnyo (commented):
It's pretty much there.

The structural change -- moving the explanation out from under the register headings -- is a very welcome one. Good move.

Also good to settle on one set of register names (and on the more sensible *I*ndex and *D*ata ones).

The biggest issue I see is not really introduced by this change, but is made more apparent by it -- the lack of basic introductory information about the CGB colours/palettes.

[possible further work] The bit 15 discussion seems to suggest there might be value in adding something about reading/writing hexadecimal colour words. Specifically about the nybbles not being aligned with the R,G,B components and the difficulty in interpreting that.

ISSOtm:
FWIW, my focus is currently on doing some drawing in order to take a break from programming, so I'll review quinn's contributions to the PR after some days. I'll keep it under two weeks though.

src/Palettes.md
```diff
@@ -67,7 +67,7 @@ Unlike `BGPD`, this register can be freely accessed outside VBlank and HBlank.
 
 As each color is two bytes in size, you must read/write this register *twice* to access a whole color.
 
-This is made much easier through the use of the address auto-increment: `BGPI`'s "address" field is automatically incremented (possibly wrapping around) after each write to this register, even if the write fails due to CRAM being inaccessible.
+This is made much easier through the use of the address auto-increment: `BGPI`'s "address" field is automatically incremented (wrapping around to zero after index 15) after each write to this register, even if the write fails due to CRAM being inaccessible.
```
nummacway: This is plain wrong. There is nothing related to index 15.
- 8 palettes with 4 colors each
- 32 colors
- 64 bytes

ISSOtm:
```suggestion
This is made much easier through the use of the address auto-increment: `BGPI`'s "address" field is automatically incremented (wrapping around from 63 back to 0) after each write to this register, even if the write fails due to CRAM being inaccessible.
```

ISSOtm:
Whoops, never mind the pushes, I fat-finger'd a push to the wrong remote ^^'

src/Palettes.md
ISSOtm: I ended up making it clear that bit 15 is ignored *for rendering*, which then makes Antonio's original proposal unambiguous&mdash;so I went with that.

src/Palettes.md
ISSOtm: I ended up going with “ignored by the rendering process”, which is more specific and doesn't imply that it isn't stored in CRAM.

src/Palettes.md
ISSOtm:
```suggestion
Much like VRAM, the CRAM data registers are inaccessible when the PPU is reading from it, that is, during [Mode 3](<#PPU modes>): writes are ignored, and reads return $FF.
```

How about this?

nummacway: This implies that writing to BCPD does not increment the index if autoinc is enabled in BCPI. But does it? It should be stated explicitly, though that's an edge case.

src/Palettes.md
ISSOtm: Simpler games that only use e.g. a single colour palette generally don't bother initialising more than that, and I don't want to imply that doing so is wrong.

I have moved the tip to the introduction.

src/Palettes.md
ISSOtm: That edge case is explicitly stated in the description of BGPD; I think it's fine to keep it only there, as “the universe will always make a better idiot”.

src/Palettes.md
quinnyo: Oh, that makes sense, for sure. I kept saying "all the colours" in my previous comment, I suppose just because the text does the same. It doesn't need to say "all".

I'm not sure what the intent/point of the tip is -- outside of the description of the system startup behaviour. Possibly:
- The system only initialises the palettes when it first starts up.
- The colours will all be white by default, so the software must set up at least one palette for any background tile(maps) to be visible.
- At least during the development process, you might want to initialise palettes that you're not using to something discernable, if only to assist in spotting errors/bugs.
- If you're relying on the palettes being all-white, don't forget to reset that as part of a soft-reset.

Just the fact:
```suggestion
All background colors are initialized as white by the boot ROM.
```

The software's responsibility:
```suggestion
All background colors are initialized as white by the boot ROM, so the software must set up any palettes it uses in order to have background tilemaps appear as anything other than solid white.
```

src/Palettes.md
quinnyo: That does make more sense. The issue I had was pretty vague, so I'll consider that a fix.

```suggestion
Much like VRAM, the CRAM data registers are inaccessible when the PPU is reading from CRAM, that is, during [Mode 3](<#PPU modes>): writes are ignored, and reads return $FF.
```

src/Palettes.md
nummacway: "as anything other than solid white." misses the point.

quinnyo: OK. Please explain it to me.

> I'm not sure what the intent/point of the tip is -- outside of the description of the system startup behaviour.

ISSOtm: My intent was indeed point 4, but I suppose it doesn't belong in Pan Docs much. Stating the hardware fact only should be sufficient here; we may even move it outside of a tip, though I'll check out the render to make that judgement.

ISSOtm:
OK, this should be all of it. Anything else?

avivace (approved):

quinnyo (commented):
LGTM

ISSOtm:
Big thanks to everyone who's pushed through this! I think the result is an excellent improvement ^^

PR #670, opened 2026-09-13
Give the Super Game Boy its own ToC chapter

ISSOtm:
Those pages hardly involve the MMIO, so it seems more appropriate to move those elsewhere; and, frankly, considering how many pages are in there, a new chapter seems warranted.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -37,20 +37,23 @@
   - [HALT](./halt.md)
 - [CGB Registers](./CGB_Registers.md)
 - [Infrared Communication](./IR.md)
-- [SGB Functions](./SGB_Functions.md)
-  - [Unlocking and Detecting SGB Functions](./SGB_Unlocking.md)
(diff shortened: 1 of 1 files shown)


avivace (approved):
OK for me, we could also put all those pages in a 'sgb' subfolder to give prettier URLs and improve the organisation of the source folder

ISSOtm:
I agree! The downside is that then we are breaking backwards compat, and we can't leave 301s from the old pages (we can try bodging it with client-side redirects, but... ick.)

Ironically, Codeberg Pages [has support for those](https://docs.codeberg.org/codeberg-pages/advanced-usage/#redirects), and @Rangi42 checked and found out that GitHub Pages appears to be the odd one out here, against Netlify, Cloudflare Pages, and now Grebedoc/Codeberg Pages.

Anyway, let's not get sidetracked, I'll open an issue for discussing the `src/` structure cleanup outside of this PR.

